PR #1, opened 2026-01-23
pair programming with Anders; configuring homebase-api

toddmitchell:


homebase-chat/src/commonMain/kotlin/id/homebase/chat/login/LoginViewModel.kt
```diff
@@ -0,0 +1,216 @@
...
+class LoginViewModel(
+    private val youAuthFlowManager: YouAuthFlowManager,
+    private val usernameStorage: UsernameStorage,
+    private val httpClient: HttpClient
+) : ViewModel() {
+
+    private val _uiState = MutableStateFlow(LoginUiState())
+    val uiState: StateFlow<LoginUiState> = _uiState
+
+    private val _uiEvent = Channel<LoginUiEvent>(Channel.BUFFERED)
```
acarlsen, line 35: I would recommend one time UI events be a property on the state that is then marked as consumed from composable, this ensure no events get lost or shown twice in case composable screen is not actively listening when event is fired etc. And is also recommend by official architecture.
https://developer.android.com/topic/architecture/ui-layer/events#consuming-trigger-updates

So something like this to call an event:
```kotlin
 _uiState.update {
    it.copy(uiEvent = LoginUiState.EventName)
}
```

Then call viewmodel function when event is consumed:
```kotlin
fun eventConsumed() {
  _uiState.update {
     it.copy(uiEvent = null)
  }
}
```

acarlsen (approved):
One suggestion.
Other than that everything looks good, I have only looked at changes to pre-existing files in the repo and the login UI, not all new new backend files that was already in the poc.

PR #2, opened 2026-01-25
Homebase api cut 2 ms

toddmitchell:


homebase-api/src/androidMain/kotlin/id/homebase/api/storage/SharedPreferences.android.kt
```diff
@@ -0,0 +1,100 @@
+package id.homebase.api.storage
+
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.util.Properties
+
+/** Desktop (JVM) implementation of SharedPreferences using Java Properties file. */
```
acarlsen, line 8: Should this be JVM impl in androidMain src?

homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ChatMessageReaderService.kt
```diff
@@ -0,0 +1,181 @@
...
+            fetchMessages(
+                conversationId = conversationId
+            )
+
+        _messages.value = result.records
+    }
+
+    // ----- existing logic, unchanged -----
+
+    suspend fun fetchMessages(
```
acarlsen, line 80: I see ChatMessageReaderService.kt is injected as a singleton - maybe to make it reusable in multiple context, lets say you for some reasons want to display two different message lists (screen and dialog with other conversation) maybe it would be better to subscribe to a specific flow by id, something like this:
```kotlin
/**
 * Returns a Flow that emits messages only for the specified conversation ID.
 * The Flow automatically refreshes when relevant drive events occur.
 */
fun observeMessages(conversationId: Uuid): StateFlow<List<MessageUiModel>> {
    val messagesFlow = MutableStateFlow<List<MessageUiModel>>(emptyList())

    scope.launch {
        // Initial load
        refreshMessages(conversationId, messagesFlow)

        // Listen for updates
        eventBus.events.collect { event ->
            if (event is BackendEvent.DriveEvent.Completed &&
                event.driveId == chatDrive
            ) {
                refreshMessages(conversationId, messagesFlow)
            }
        }
    }

    return messagesFlow.asStateFlow()
}

private suspend fun refreshMessages(
    conversationId: Uuid,
    flow: MutableStateFlow<List<MessageUiModel>>
) {
    val result = fetchMessages(conversationId = conversationId)
    flow.value = result.records
}

```

acarlsen (changes requested):

homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ChatMessageReaderService.kt
toddmitchell: Ok, yes this makes sense.  I will merge this into main then come back to this immediately.

PR #23, opened 2026-02-05
feat: Implement message reply functionality with new UI components, state management, and updated actions.

2002Bishwajeet:


acarlsen (commented):
Looks good.
Suggest we use same code style formatting, I am just using default Kotlin one with 4 space indents, it seems you have 8 spaces sometimes, which makes for a lot of whitespace changes.
https://kotlinlang.org/docs/coding-conventions.html#indentation

PR #80, opened 2026-03-04
feat: sql cipher

2002Bishwajeet:


Seifert69:
Merging - but wondering if we delete the DB key from secure storage when you logout? We should.

PR #122, opened 2026-03-20
Dbm mini fix

Seifert69:


Seifert69:
Two minor issues, had no impact on how it worked, but was essentially not really right.

PR #130, opened 2026-03-21
Fast conversation overview on cold start

Seifert69:
Changes that might make the convo list, on cold start, go from seconds to ms

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/auth/OwnerSessionRepository.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/auth/OwnerSessionRepository.kt
@@ -1,6 +1,7 @@
 package id.homebase.api.client.auth
 
 import co.touchlab.kermit.Logger
+import com.russhwolf.settings.Settings
 import id.homebase.api.common.OdinId
 import io.ktor.client.HttpClient
 import io.ktor.client.request.get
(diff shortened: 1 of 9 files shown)


homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
```diff
@@ -132,8 +132,6 @@ class ConversationListViewModel(
                 ownerSessionRepository.user
             ) { convos, contacts, ownerSession ->
 
-                if (ownerSession == null) return@combine emptyList()
```
Seifert69, line 135: This is blocking according to Claud

homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/contact/ConnectionService.kt
```diff
@@ -27,12 +27,6 @@ class ConnectionService(
     val connections: StateFlow<ConnectionState> =
         _connections.asStateFlow()
 
-    init {
-        scope.launch {
```
Seifert69, line 31: Already done in refresh

Seifert69:
Todd please review & try to run it locally and see if your still see conversations, or if it broke completely.

Context                                                                                                                       
                                                                                                                               
 On cold start with ~83 conversations, the conversation overview screen shows "No unread conversations" for several seconds    
 before displaying the list. The user confirmed that individual DB reads (conversations + unread counts) are fast. This        
 analysis traces the actual cause of the delay.                                                                                
                                                                                                                               
 ---                                                                                                                           
 Root Cause: Network-Gated Combine                                                                                             
                                                                                                                               
 The core issue is in ConversationListViewModel.kt:129-154:                                                                    
                                                                                                                               
 combine(                                                                                                                      
     conversationStream.conversations,                                                                                         
     contactService.contacts,                                                                                                  
     ownerSessionRepository.user     // ← STARTS AS NULL                                                                       
 ) { convos, contacts, ownerSession ->                                                                                         
     if (ownerSession == null) return@combine emptyList()  // ← HARD BLOCK                                                     
     convos.map { enricher.enrich(it, contactMap, ownerSession) }                                                              
 }.collect { enriched ->                                                                                                       
     updateListContent()  // only called with real data after all 3 flows are non-empty                                        
 }                                                                                                                             
                                                                                                                               
 ownerSessionRepository.user starts as null (line 22 in OwnerSessionRepository.kt):                                            
 private val _user = MutableStateFlow<OwnerSession?>(null)                                                                     
                                                                                                                               
 It only becomes non-null after a network HTTP call completes:                                                                 
 // OwnerSessionRepository.kt:30-35                                                                                            
 private suspend fun fetch(odinId: OdinId): OwnerSession {                                                                     
     val url = "https://$odinId/cdn/sitedata.json"                                                                             
     val response = httpClient.get(url)  // ← NETWORK CALL                                                                     
     ...                                                                                                                       
 }                                                                                                                             
                                                                                                                               
 Call chain from app start to network call:                                                                                    
 1. AuthConnectionCoordinator.init {} collects youAuthFlowManager.authState                                                    
 2. Auth state becomes Authenticated → calls connect() + loadProfile() (async)                                                 
 3. loadProfile() → ownerSessionRepository.load(odinId) → HTTP GET to sitedata.json                                            
 4. Only after network response → _user.value becomes non-null                                                                 
 5. Only then does the ViewModel's combine emit non-empty result                                                               
 6. Only then does updateListContent() call ConversationListContentState.Items(...)                                            
                                                                                                                               
 This means: conversations cannot render until the owner profile is fetched from the network, even though all conversation data is already in the local DB.

homebase-api/src/commonMain/kotlin/id/homebase/api/client/auth/OwnerSessionRepository.kt
```diff
@@ -122,5 +173,6 @@ class OwnerSessionRepository(
 
     fun clear() {
         _user.value = null
+        settings.remove(KEY_ODIN_ID)
```
acarlsen, line 176: Should we also remove other cached values on clear (logout)?

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
acarlsen: It will return early if no ownersession, but I dont see whats blocking about (as in blocking a thread).

homebase-api/src/commonMain/kotlin/id/homebase/api/client/auth/OwnerSessionRepository.kt
Seifert69: That's probably a good idea. They're not sensitive but it's a good practice. And we don't risk two identities getting the wrong datat from te last one

Seifert69: Oddly I cannot reply to your comment below so I'll write the reply here (I'm on my cell):

> It will return early if no ownersession, but I dont see whats blocking about (as in blocking a thread).

I believe it's the combine which forces it to wait for getting the http call completed. The fix stores the last public data in a cache and returns that so we don't have to wait for the network call to complete. 

Feel free to change the code - I have no feelings (it's Claude's code). 

It'd be great if it was more clear and perhaps even the http call should be in a scope.launch ?

Seifert69: Both grok and Claude believe the http call is blocking the UI from rendering the conversation list. And I'm definitely seeing a delay consistent with network having to complete

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
Seifert69: See my comment in the other comment above (wasnt able to post it via mobile app)

Seifert69:
Too confusing... starting over and braking it down

PR #160, opened 2026-03-26
Clause proposed changes

acarlsen:
Investigating an obfuscated Android crash:
     java.util.ConcurrentModificationException
         at java.util.LinkedHashMap$LinkedHashIterator.nextNode
         at java.util.LinkedHashMap$LinkedKeyIterator.next
         at ci0.b (line 68)
         at ai0.invokeSuspend (line 12)   ← suspend/coroutine context
         Suppressed: [fa6{Cancelling}@950ec6b, Dispatchers.Default]

     Key signals:
     - LinkedKeyIterator — something iterates keys of a LinkedHashMap (or a LinkedHashSet, which is backed by LinkedHashMap and uses LinkedKeyIterator internally)
     - invokeSuspend — crash is inside a coroutine
     - Cancelling — the coroutine was being cancelled at the time
     - Dispatchers.Default — running on a thread pool (multi-threaded)

     Candidate 1: DriveSyncManager.kt — most suspicious

     File: homebase-api/src/commonMain/kotlin/id/homebase/api/sync/DriveSyncManager.kt

     The driveSyncs var is protected by driveSyncsMutex for suspend callers (start, syncAll, syncAllFailed) but three non-suspend methods bypass the mutex entirely:

     ┌────────────────┬─────────┬─────────────────────────────────────────────────┐
     │     Method     │  Line   │                      Issue                      │
     ├────────────────┼─────────┼─────────────────────────────────────────────────┤
     │ pause()        │ 154–156 │ Reads + iterates driveSyncs.values with no lock │
     ├────────────────┼─────────┼─────────────────────────────────────────────────┤
     │ stop()         │ 158–163 │ Reads + writes driveSyncs with no lock          │
     ├────────────────┼─────────┼─────────────────────────────────────────────────┤
     │ clearStorage() │ 165–176 │ Reads driveSyncs.values with no lock            │
     ├────────────────┼─────────┼─────────────────────────────────────────────────┤
     │ syncDrive()    │ 149–152 │ Reads driveSyncs[driveId] with no lock          │
     └────────────────┴─────────┴─────────────────────────────────────────────────┘

     pause() is called from AuthConnectionCoordinator.onDisconnected callback (line 118), and stop() from disconnect() (line 130). These can race with suspend functions holding driveSyncsMutex.

     driveSyncs is also not @Volatile, so writes from stop() may not be visible to other threads reading it in start().

     CME mechanism: The driveSyncs + (driveId to sync) operation (line 120, inside mutex) copies the current map into a new LinkedHashMap. If stop() (without mutex) races and replaces driveSyncs while another coroutine has a reference to the
     partially-constructed new map being iterated... this is possible though tricky to prove.

     More concretely: syncAllFailed() does:
     val failedIds = _driveStatuses.value.filter { ... }.keys  // LinkedKeyIterator!
     val jobs = driveSyncsMutex.withLock { failedIds.mapNotNull { driveSyncs[it] } }
     The failedIds.mapNotNull { ... } iterates a LinkedHashMap's key set using LinkedKeyIterator. However syncAllFailed appears to be dead code (never called in production). The same pattern could appear elsewhere.

     Candidate 2: DriveFileProviderCached.kt — general thread-safety concern

     File: homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/cache/DriveFileProviderCached.kt

     private val notFoundCache = mutableSetOf<String>()   // line 46

     mutableSetOf() returns a LinkedHashSet (backed by LinkedHashMap, uses LinkedKeyIterator). This field is accessed from multiple concurrent coroutines:

     - Line 80 and 192: if (cacheKey in notFoundCache) — OUTSIDE any mutex
     - Line 110 and 230: notFoundCache.add(cacheKey) — inside per-key mutex
     - Line 360: notFoundCache.clear() — clearCaches() suspend fn, no mutex

     contains and add individually don't use iterators, but concurrent add() calls from different per-key mutexes can trigger a rehash. Java's LinkedHashMap resize internally iterates using LinkedKeyIterator, and if another add() happens
     concurrently it will throw CME at exactly the shown stacktrace location.

     Most Likely Crash Path

     Given invokeSuspend at line 12 (very short coroutine body) and the Cancelling suppressed exception, this points toward the retry coroutine in DriveSyncManager.init (lines 80–83):

     scope.launch {
         delay(1000L)
         driveSyncsMutex.withLock { driveSyncs[event.driveId] }?.sync()
     }

     Or the syncAll() call chain launched in AuthConnectionCoordinator.onConnected (line 102–114), racing with a stop()/pause() triggered by a disconnection event.

     Proposed Fixes

     Fix 1: Make driveSyncs properly thread-safe

     Add @Volatile to driveSyncs and have stop(), pause(), clearStorage(), syncDrive() acquire the mutex (convert to suspend or use a different mechanism):

     @Volatile private var driveSyncs: Map<Uuid, DriveSync> = emptyMap()

     fun pause() {
         // Take a snapshot; map is immutable so iteration on snapshot is safe
         val snapshot = driveSyncs  // @Volatile read guarantees fresh value
         snapshot.values.forEach { it.cancel() }
     }

     fun stop() {
         val old: Map<Uuid, DriveSync>
         synchronized(this) {  // or use atomic reference swap
             old = driveSyncs
             driveSyncs = emptyMap()
         }
         old.values.forEach { it.cancel() }
         _driveStatuses.update { emptyMap() }
     }

     Fix 2: Protect notFoundCache with a mutex

     private val notFoundCache = mutableSetOf<String>()
     private val notFoundCacheMutex = Mutex()

     // Replace `cacheKey in notFoundCache` (line 80) with:
     if (notFoundCacheMutex.withLock { cacheKey in notFoundCache }) { ... }
     // etc.

     Or replace mutableSetOf() with a ConcurrentHashMap-backed set:
     private val notFoundCache = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())

     Critical Files

     - homebase-api/src/commonMain/kotlin/id/homebase/api/sync/DriveSyncManager.kt (lines 34, 149–176)
     - homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/cache/DriveFileProviderCached.kt (lines 41–46, 80, 110, 192, 230, 360)
     - homebase-common/src/commonMain/kotlin/id/homebase/core/auth/AuthConnectionCoordinator.kt (lines 118, 130)

⏺ Both fixes are in place. Here's a summary of what was done:

  DriveSyncManager.kt — Added @Volatile to driveSyncs (line 36). The map is always replaced never mutated, so iterating any snapshot is safe from CME; @Volatile guarantees non-suspend callers (pause, stop, clearStorage, syncDrive) always read
  the latest reference rather than a CPU-cached stale one.

  DriveFileProviderCached.kt — Added notFoundCacheMutex = Mutex() and wrapped all 6 access sites. The root issue: mutableSetOf() returns a LinkedHashSet backed by LinkedHashMap. Multiple coroutines on Dispatchers.Default each held their own
  per-key mutex (different keys → truly concurrent), so concurrent add() calls could trigger a rehash. During rehash, Java iterates the old table using LinkedKeyIterator — exactly matching LinkedHashMap$LinkedKeyIterator.next in your
  stacktrace — and if another add() changes modCount mid-iteration, CME is thrown.


Seifert69:
I ran through this with Claude on the latest Main - and we decided to make a more optimistic version that locks on writes, not on reads. Should perform marginally better.

PR #170, opened 2026-03-27
Fix 1-2s delay before optimistic message insert in airplane mode

Seifert69:
The optimistic DB write (which shows the sent message in the UI) was happening after outboxSync.tryEnqueue(), which immediately fires a background network upload attempt on Dispatchers.IO. In airplane mode, those IO threads hang on TCP connection timeouts. SQLDelight's DB writes also run on withContext(Dispatchers.IO), so the optimistic write had to wait for a free thread — causing a 1-2 second delay before the message appeared.

Fix: write the optimistic record *before* enqueueing to the outbox, so the message appears instantly regardless of network state.

Two-level change:
- ChatMessageSenderService.sendMessageInternal: move writeNewFile() to run before tryEnqueue() instead of after
- ConversationListViewModel.addMessage/replyToMessage: call writePlaceholderMessage() immediately after generating the message ID, before any encryption or network work (mirrors the existing addMessageWithFiles pattern)
- ChatMessageSenderService.writePlaceholderMessage: add optional replyPreview param so reply placeholders show the quoted context immediately


toddmitchell:
approved and merged yet perhap you should remove the optimistic write if tryEnqueue fails?

PR #214, opened 2026-04-04
Fix horizontal swipe gesture interfering with vertical scroll

Seifert69:
Replace detectHorizontalDragGestures with a custom awaitEachGesture handler that requires explicit horizontal intent before claiming the gesture. The finger must move at least 1.5x more horizontally than vertically (within ~33 degrees of horizontal) before the swipe activates. Otherwise the touch event passes through to the parent LazyColumn for normal vertical scrolling.

  - Replaced detectHorizontalDragGestures with a custom awaitEachGesture handler that decides drag direction explicitly
  - Direction check: after the finger moves past touch slop, the gesture is only claimed as horizontal if |horizontalDelta| > |verticalDelta| * 1.5 (~33
  degrees from horizontal). Otherwise the gesture is ignored, letting the parent LazyColumn scroll normally
  - isDragging deferred: only set to true after confirming horizontal intent, so vertical scrolling is never interrupted
  - All existing behavior (72dp threshold, 100dp max offset, haptic feedback, spring-back animation, reply/info actions) is preserved

  To test: scroll up/down through a conversation — diagonal finger movement should no longer accidentally trigger the swipe. Intentional horizontal swipes
  should still work as before. The 1.5x ratio is easy to tune if you find it needs adjustment.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/SwipeableMessageWrapper.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/SwipeableMessageWrapper.kt
@@ -3,7 +3,8 @@ package id.homebase.chat.widget
 import androidx.compose.animation.core.Animatable
 import androidx.compose.animation.core.spring
 import androidx.compose.foundation.background
-import androidx.compose.foundation.gestures.detectHorizontalDragGestures
+import androidx.compose.foundation.gestures.awaitEachGesture
+import androidx.compose.foundation.gestures.awaitFirstDown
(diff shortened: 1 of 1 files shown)


Seifert69:
@acarlsen please review. Trying to make the horizontal swipe on messages less sensitive so it doesn't interfere so much with daily scrolling in messages. I've tested it in the emulator, but it's impossible to know until I use it day to day on my phone. I don't have enough experience with swiping code to know if this PR makes sense. Merge if OK.

acarlsen (approved):
Seems to work fine here

PR #244, opened 2026-04-10
Dont loose message

Seifert69:
# Message Send Pipeline — Change Summary

## Problem

Users reported messages to Todd (and potentially others) disappearing — the message was never visible in the chat stream after tapping send. Investigation of `homebase.log` revealed no errors from the send pipeline, meaning failures were being silently swallowed.

## Root Causes Found

1. **`OptimisticWriter.writeNewFile()` swallowed exceptions** — caught exceptions, logged only `e.message` (no tag, no stack trace), and never re-threw. The caller had no idea the DB insert failed.

2. **Wrong operation ordering** — the optimistic DB insert ran *before* the outbox enqueue. If the insert failed, the outbox enqueue never ran, so the message was silently lost. The outbox is the source of truth; the optimistic insert is just a UI convenience.

3. **Input cleared before send completed** — the message text was cleared synchronously in the action handler, before the async send coroutine even started. If the send failed, the user's text was gone with no way to recover it.

4. **No double-send protection** — with deferred clearing, the user could tap send twice before the first send completed, producing duplicate messages.

### 1. `OptimisticWriter.kt` — proper error logging + re-throw

**Before:** All catch blocks used `Logger.e("message string")` — no tag, no throwable, exception swallowed.

**After:** All catch blocks use `Logger.e(throwable = e, tag = TAG) { "context" }` — full stack trace in logs. `writeNewFile()` and `writeUpdate()` now re-throw so callers can handle failures. Cleanup methods (`removeOptimisticFile`, `rollbackWrite`, etc.) still swallow since they're best-effort.

### 2. `ChatMessageSenderService.kt` — outbox-first ordering, drop placeholder

**Before:**
```
writePlaceholderMessage()  →  encrypt  →  optimisticWriter.writeNewFile()  →  outbox enqueue
```
If optimistic write failed, outbox never ran. Message lost.

**After:**
```
encrypt  →  outbox enqueue (success gate)  →  optimisticWriter.writeNewFile() (best-effort)
```
Once the outbox accepts the message, the operation is a success. The optimistic write makes the message appear in the chat stream immediately, but if it fails, the outbox delivery + sync will bring the message back.

`writePlaceholderMessage()` deleted entirely — it did the same thing as the optimistic write (DB insert + `BatchReceived` event emission → message appears in UI), but ran before encryption to shave off a few milliseconds of display latency. Since encryption is near-instant for text, the timing difference is imperceptible, and the optimistic write after outbox enqueue now serves the same purpose with the real encrypted data.

Trace logging added at each step: `sendMessageInternal: starting`, `encrypting`, `outbox enqueued`, `optimistic write complete/failed`.

### 3. `ConversationListViewModel.kt` — deferred input clearing + send button guard

**Input clearing deferred to success path:**
- `messageInputTextState.clear()` moved from the synchronous `SendMessage`/`SendFile` action handlers into `addMessage()`, `replyToMessage()`, and `addMessageWithFiles()` — only after `sendNewMessage()` returns successfully.
- Reply-to preview (`replyToMessage = null`) also deferred.
- Attachment overlay (`fullScreenOverlay = null`) also deferred.
- On failure: text, reply preview, and attachment overlay remain intact so the user can retry.

**Send button disabled during send:**
- `isSendingMessage` flag set `true` synchronously in the action handler (same composition frame — no race window).
- Set `false` in a `finally` block — guaranteed to run on success, failure, cancellation, or even `Error`.
- `addMessageWithFiles()` try/catch widened to cover the entire body (attachment processing, HEIC conversion, etc.) so an early failure can't leave the flag stuck.

**Trace logging added** to `addMessage()`, `replyToMessage()`, `addMessageWithFiles()` — entry, completion, and full throwable in catch blocks.

### 4. `MessageInputBar.kt` — send guard + deferred clearing

- `sendMessage()` checks `!isSendingMessage` before dispatching — blocks both button taps and keyboard Enter during send.
- Removed `textFieldState.clear()` and `linkPreviewData = null` — the ViewModel now owns clearing. `linkPreviewData` auto-clears via existing `LaunchedEffect` when the text is emptied.
- `isSendingMessage` parameter threaded to `MessageTextFieldCompact` and `BlueBackgroundIconButton` for visual disabled state on the compact layout's send button.

### 5. `ConversationContent.kt` — pass `isSendingMessage` to `MessageInputBar`

One line: `isSendingMessage = uiState.isSendingMessage`.

### 6. `ConversationListUiState.kt` — new field

`val isSendingMessage: Boolean = false` added to `MessageListUiState`.

## Architecture: Send Pipeline

```
User taps Send
       |
       v
  ViewModel action handler (synchronous)
    - set isSendingMessage = true
    - read text from messageInputTextState
    - launch coroutine
       |
       v
  sendNewMessage() / replyToMessage()
    - encrypt payload
    - build upload request
       |
       v
  outbox.tryEnqueue()  ← SUCCESS/FAIL DECISION POINT
    |            |
    | fail       | success
    v            v
  throw     optimisticWriter.writeNewFile() (best-effort)
    |            |  → DB insert + BatchReceived event
    |            |  → message appears in chat stream (no network needed)
    v            v
  catch:     messageInputTextState.clear()
  show toast  dismiss overlays
  text stays
       |
       v
  finally: isSendingMessage = false (always runs)
```

## Testing

- Send a text message — should clear input after send, message appears in stream
- Send a reply — input and reply preview clear after send
- Send with attachments — input clears, attachment overlay dismisses after send
- Force a failure (leave a group, then send) — text stays in edit box, error toast, button re-enables
- Rapid double-tap send — only one message sent (button disabled during send)
- Kill network, send message — outbox queues it, optimistic write makes message appear immediately in chat stream (no network needed); delivery to recipient happens when connectivity returns

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
@@ -58,6 +58,7 @@ data class MessageListUiState(
     val downloadingFiles: Set<String> = emptySet(),
     val recordingData: RecordingData? = null,
     val uiSheet: MessageListUiSheet? = null,
+    val isSendingMessage: Boolean = false,
 )
 
 sealed interface MessageListUiSheet {
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
@@ -126,6 +126,10 @@ class ConversationListViewModel(
     private val localVideoContextStore: LocalVideoContextStore,
 ) : ViewModel() {
 
+    companion object {
+        private const val TAG = "ConversationListViewModel"
+    }
+
     private val enricher = ConversationEnricher()
     val ownerSession = ownerSessionRepository.user
 
@@ -1839,6 +1843,7 @@ class ConversationListViewModel(
 
                 val newMessageId = Uuid.random()
                 pendingMessageId = newMessageId
+                Logger.d(tag = TAG) { "addMessage: message=$newMessageId conversation=$conversationId" }
 
                 // Write placeholder first — message appears instantly before encryption/network
                 chatMessageSenderService.writePlaceholderMessage(
@@ -1855,7 +1860,9 @@ class ConversationListViewModel(
                     previousMessageUniqueId = null,
                     payloadBundle = payloadBundle,
                 )
+                Logger.d(tag = TAG) { "addMessage: complete message=$newMessageId" }
             } catch (e: Exception) {
+                Logger.e(throwable = e, tag = TAG) { "addMessage failed for conversation=$conversationId" }
                 sendEvent(ShowErrorMessage("Failed to send message: ${e.message}"))
             }
         }
@@ -1881,6 +1888,7 @@ class ConversationListViewModel(
                 )
                 val newMessageId = Uuid.random()
                 pendingMessageId = newMessageId
+                Logger.d(tag = TAG) { "replyToMessage: message=$newMessageId conversation=$conversationId replyTo=${replyTo.id}" }
 
                 // Write placeholder first — reply appears instantly with quote context
                 chatMessageSenderService.writePlaceholderMessage(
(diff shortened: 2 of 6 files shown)


acarlsen (approved):
Looks ok to me

toddmitchell (approved):
yarp, makes sense this could work

PR #245, opened 2026-04-10
Skip warm start message reload

Seifert69:
Problem

When the app warm-starts (e.g. brought from background) with a conversation already open, the user sees a loading spinner while 1000 messages are re-queried from the local SQLite database — even though those exact messages are already cached in memory inside the singleton ChatMessageStream.

This causes a noticeable delay and UI flash every time the user returns to the app.

Root Cause

  ConversationListViewModel.loadMessagesForConversation() unconditionally:
  1. Sets isLoadingMessages = true (shows a full-screen spinner, hiding all messages)
  2. Calls ChatMessageStream.loadConversation() which queries SQLite for 1000 messages
  3. Only then re-establishes the observeMessages flow and hides the spinner

It never checks whether the messages are already available in the in-memory ActiveConversationState cache — which they are on warm start, because ChatMessageStream is a Koin singleton that survives ViewModel recreation.

Fix

ChatMessageStream.kt — Added hasCachedMessages(conversationId) method that checks whether ActiveConversationState already holds messages for a given conversation.

ConversationListViewModel.kt — Before showing the spinner or querying the DB, loadMessagesForConversation now checks two conditions:
  - The conversation being loaded is the same one already selected (conversationId == selectedConversationId)
  - The singleton cache has messages for it (chatMessageStream.hasCachedMessages(conversationId))

If both are true (the warm-start case), the spinner and DB query are skipped entirely. The observeMessages flow is still set up so incremental updates continue to arrive via EventBus.

If either condition is false (cold start, or switching to a different conversation), the full spinner + DB load runs exactly as before.

Files Changed
  - homebase-chat/.../services/ChatMessageStream.kt — Added hasCachedMessages()
  - homebase-chat/.../conversationlist/ConversationListViewModel.kt — Conditional skip in loadMessagesForConversation()

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
@@ -1669,8 +1669,13 @@ class ConversationListViewModel(
         messageIdForScroll: Uuid?,
         scrollToBottom: Boolean = false
     ) {
-        _messagesUiState.update { it.copy(scrollPosition = null, isLoadingMessages = true) }
+        val hasCachedMessages =
+            conversationId == _uiState.value.selectedConversationId &&
+                    chatMessageStream.hasCachedMessages(conversationId)
 
+        _messagesUiState.update {
+            it.copy(scrollPosition = null, isLoadingMessages = !hasCachedMessages)
+        }
 
         // When loading message for newly selected conversation, cancel any previous job to
         // avoid observing multiple messageStreams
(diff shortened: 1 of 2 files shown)


acarlsen:
When does this warm-start with new viewmodel load happen?
Simply having app in background and foregrounding it does not trigger loadMessagesForConversation, since viewModel is kept. Is it only if app is is tombstoned/killed if its a long time in the background?

Seifert69:
> When does this warm-start with new viewmodel load happen? Simply having app in background and foregrounding it does not trigger loadMessagesForConversation, since viewModel is kept. Is it only if app is is tombstoned/killed if its a long time in the background?

Answer: It happens when you swipe away the app on your mobile phone, and then reopen it, for example via clicking on a notification. Problem  is this. I am in a conversation with Todd. Swipe the app away. You send me a message. I click your notification. The app opens in the conversation with Todd (the messages are actually already in memory). The spinner spins ~1-2 seconds while loading 1000 messages (something we also need to fix), then it renavigates over to you, spins and loads 1000 messages.

Seifert69:
I'll try to merge it, and if my spinner stops on my Android then it is sold, otherwise we'll untangle it again.

PR #288, opened 2026-04-15
[DEV] feat: webview

2002Bishwajeet:
DO NOT MERGE


homebase-common/src/commonMain/kotlin/id/homebase/core/config/AppConfig.kt
```diff
@@ -95,11 +99,25 @@ val targetDriveAccessRequest: List<TargetDriveAccessRequest> =
...
+                DrivePermission.Write,
+                DrivePermission.React,
+                DrivePermission.Comment
+            )
+
+        ),
     )
 
 // Drives we listen to for sockets and synchronization
-val syncLabeledDrives: List<LabeledDrive> = listOf(chatLabeledDrive, contactLabeledDrive, feedLabeledDrive)
```
2002Bishwajeet, line 102: we should have this. why did we remove contact sync in websockets?

PR #306, opened 2026-04-16
Ignore duplicate auth callback deliveries

Seifert69:
Android re-delivered the same deep-link Intent on activity resume, causing YouAuthFlowManager to emit AuthState.Error mid-login even though the first callback had succeeded. Clear intent.data after dispatch (matching the conversation branch) and demote a missing registry entry to a debug log so late/duplicate callbacks on any platform cannot stomp the in-flight or completed auth state.


Seifert69:
Verified on Desktop

PR #314, opened 2026-04-18
fix: deliver notification-tap navigation events on cold start

Seifert69:
Tapping an Android push notification while the app was cold-starting landed on the conversation overview instead of the target conversation. The log showed the intent arriving (`MainActivity: Notification intent detected`) but no subsequent `Navigating to detail for <id>` entry.

Root cause: both `NotificationService._navigationEvents` and `AppViewModel._navigationEvents` were `MutableSharedFlow(replay = 0)`. `MainActivity.handleNotificationIntent` runs synchronously in onCreate and emits the event before `AppViewModel.init` subscribes to the service flow, and before `AppNavHost`'s `LaunchedEffect(Unit)` subscribes to the ViewModel flow. With replay=0, emissions that predate the first subscriber are silently dropped — so the event was lost at one of the two layers every cold start.

Replace both flows with `Channel<NotificationNavigationEvent>(BUFFERED) .receiveAsFlow()`. A Channel queues emissions until a collector attaches and delivers each event exactly once, which is the right semantic for a fire-once navigation intent. `receiveAsFlow` is single-consumer, matching the one-collector-per-flow topology we already have. Switches `tryEmit` → `trySend` at the three call sites.

### DIFF
--- a/homebase-common/src/commonMain/kotlin/id/homebase/core/notifications/NotificationService.kt
+++ b/homebase-common/src/commonMain/kotlin/id/homebase/core/notifications/NotificationService.kt
@@ -18,10 +18,13 @@ import id.homebase.core.navigation.ActiveConversation
 import id.homebase.core.settings.UserPreferences
 import id.homebase.core.util.Platform
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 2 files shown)


Seifert69:
There's a problem with this - stays in previous conversation and doesn't move to target conversation.

PR #333, opened 2026-04-20
feat(notifications): defer notification-tap navigation until drive sync lands the target conversation

Seifert69:
When a push notification arrives for a brand-new conversation (or a new message in an already-known conversation) and the user taps it, the local drive may not yet have synced the target. The previous "fire once" flow would still route to a Detail screen that couldn't render properly — no header, no participants, no scroll-to-message — and silently stay that way because the tap event was already consumed.

This change introduces a small `PendingNotificationTap` singleton that holds `(conversationId, messageId, createdAt)` across the sync gap. ConversationListViewModel observes it alongside
`conversationStream.conversations`; when the conversation appears in the list (which happens reactively once sync lands a `BatchReceived` event), the VM fires `selectConversation(convoId, msgId, scrollToBottom=true)` exactly once and clears the tap. No explicit hook into `DriveSyncManager.syncState` is needed — the conversation list already updates reactively on sync, which is the signal we care about.

Strict gating: the pending tap is only set when the payload carries **both** a parseable conversationId AND a parseable messageId. A messageId-less payload is treated as ambient "new activity" — the user lands on ChatList without a speculative auto-jump, avoiding misleading navigation when backend coalesces notifications. The Channel-based `OpenConversation` event still fires unconditionally so AppNavHost still pops the back stack to ChatList.

Additionally fixes a latent bug in `loadMessagesForConversation`: `messageIdForScrollNullable` was cleared on every
`ChatMessagesData.Messages` emission, so "scroll to message X" could never survive the sync gap even when X arrived on a later emit. It now clears only when the message is actually found, so the next emission retries the lookup.

The AppNavHost `OpenConversation` collector no longer writes `pendingConversationId` to ChatList's `savedStateHandle`; resolution moved to the VM via the singleton. The savedStateHandle path is kept intact for other callers (post-create-conversation routing, etc.).

Manual-click and onCleared both clear the pending tap so a stale notification can't yank the user to a different conversation after they've picked one explicitly.

## Files

- NEW homebase-common/.../notifications/PendingNotificationTap.kt
- NEW homebase-common/src/jvmTest/.../PendingNotificationTapTest.kt      (7 cases)
- NEW homebase-common/src/jvmTest/.../ChatTapGatingTest.kt               (6 cases)
- NEW homebase-chat/src/jvmTest/.../ConversationListPendingTapResolutionTest.kt (5 cases)
- MOD homebase-common/.../NotificationService.kt — inject holder; set tap when both ids parse; extracted pure `extractChatTapIds(typeId, tagId): Pair<Uuid, Uuid>?`
- MOD homebase-core/.../di/AppModule.kt — register `PendingNotificationTap` singleton
- MOD homebase-core/.../ui/navigation/AppNavHost.kt — drop savedStateHandle write on OpenConversation; keep pop-back-to-ChatList
- MOD homebase-chat/.../ConversationListViewModel.kt — inject holder; `combine(pendingTap, conversations)` resolution collector; clear on `ConversationClicked`; `onCleared`; fix message-scroll retry; pure `resolveNotificationTap(tap, dataReady, conversationIds)`

## Verification

- `homebase-common:jvmTest homebase-chat:jvmTest homebase-core:jvmTest` green for all touched tests. Pre-existing `NotificationSettingsUiTest.showsReRegisteringState` failure (broken since d7d0ce95 "Fix hardcoded strings" 2026-04-16 — hardcoded English literal replaced with a stringResource the JVM Compose test harness doesn't resolve) is unrelated to this change.

## Manual test plan

1. Device A sends a message from a brand-new conversation to device B (killed / backgrounded).
2. Device B receives the notification; tap it **before** drive sync completes.
3. Expect: ChatList opens → Detail screen populates with correct header/participants/avatar once sync lands, scrolled to the new message.
4. Log `grep -i pendingNotificationTap` should show one `set` then one `resolved` line.


Seifert69:
TESTING ON DEV BEFORE MERGE

PR #353, opened 2026-04-22
feat(sync): drive registry, mount/unmount, and 403 permission handling

Seifert69:
Introduces a mandatory/optional drive split and runtime drive management:

**DriveRegistry** (`homebase-common`)
- New `DriveRegistry` class persists optional drives as a JSON blob in the encrypted DB key-value store (UUID key `00000000-0000-0000-0000-000000d71700`).
- First-launch migration: when no entry exists, seeds with `feedLabeledDrive` so existing users retain feed sync without any explicit action.
- `addDrive` is idempotent; `removeDrive` writes an explicit empty array so the seed path is not re-triggered on subsequent startups.
- `hasDrive` convenience helper for callers that need a membership check.

**Mandatory vs optional drives** (`AppConfig`)
- `LabeledDrive` is now `@Serializable` (required for JSON persistence).
- Added `profileLabeledDrive` (was previously absent from sync).
- Renamed `syncLabeledDrives` → `mandatorySyncDrives` containing chat, contacts, and profile. Feed moves to the optional registry.
- `AuthConnectionCoordinator` merges mandatory + registry drives at connect time; `DriveSyncManager` factory in Koin does the same at startup.

**mountDrive / unmountDrive** (`DriveSyncManager`)
- `mountDrive(driveId, label)`: creates and starts a `DriveSync` mid-session (e.g. when a user activates an add-on). HTTP polling starts immediately; WebSocket push arrives after next reconnect.
- `unmountDrive(driveId)`: cancels the sync, removes the drive from `driveStatuses`, and lets `computeSyncState` return `Idle` naturally when the map is empty. Does NOT modify `DriveRegistry` — session-only removal.

**403 Forbidden → PermissionDenied** (`DriveSync` / `DriveSyncManager`)
- New `BackendEvent.DriveResult.PermissionDenied` sealed variant.
- `DriveSync` catches `ForbiddenException` before the generic handler, sets `killroy = false` (no retry loop), and emits `PermissionDenied`.
- `DriveSyncManager` responds by calling `unmountDrive` for the session, clearing the sync indicator instead of leaving it permanently orange. Other drives are unaffected; a single denied drive does not poison the overall `SyncState`.

**Tests**
- `DriveRegistryTest` (7 cases): seed, add, idempotency, targeted remove, remove-all-no-reseed, hasDrive true/false.
- `DriveSyncTest` (2 cases): 403 emits `PermissionDenied` not `Failure`; 403 does not trigger a retry after 2 s.
- `DriveSyncManagerTest` (4 new cases): `PermissionDenied` unmounts drive and clears indicator; denied drive doesn't mark other drives failed; `mountDrive` adds drive; `unmountDrive` removes drive and clears status.

**Docs**: `ADDING_ADDON_APPS.md` updated with mandatory vs optional drive table, activation pattern, 403 session-only unmount explanation, and two new known gotchas (WebSocket subscription timing, session-only unmount).


Seifert69:
Probably ignore this branch and work from the one based on MAIN

PR #359, opened 2026-04-23
Dev screen and app update check

acarlsen:


Seifert69:
Don't merge until we touch base on the two approaches.

PR #384, opened 2026-04-27
Group heal tooling and outbox-chain serialization for chat sends

Seifert69:
Description

  ## Summary

  Two related strands shipped together on this branch:

  1. **Diagnostic + recovery tooling** for the "recipient didn't receive my group" bug class:
     per-recipient delivery indicators on the Group Settings screen, a "Heal group" button
     for the original author, and richer logging so a stuck case can be traced through both
     peers' `homebase.log` files.
  2. **Outbox dependency-chain serialization** for chat sends so that the conversation file,
     the admin file, status messages, and user-typed messages release through the local
     outbox in strict order rather than fanning out the moment an ancestor is acknowledged.
     Closes the local-outbox half of the orphan-recovery race that produced Shelly's stuck
     placeholder on conversation `0e684619` (Apr 19 log timeline analyzed in commit
     `8dedb8f3`).

  ## Background — the bug we were chasing

  A recipient ("Shelly") in a 4-person group conversation showed only `[creator, self]` as
  participants on her phone, despite the desktop app's transfer-history reporting all three
  recipients as `Delivered=OK`. Pre-`3d09be30` (Apr 24) orphan-recovery wrote a 2-person
  placeholder to her local DB when the first message arrived 11 ms ahead of the conversation
  file in her drive sync; the placeholder shadowed the real conversation file and persisted
  through every subsequent sync. The Apr 24 patches removed the server-write side-effect and
  fixed the upsert-guard for new orphans, but the **upstream race** — admin file and message
  arriving in batch 1, conversation file in batch 2 — was still possible. This branch
  serializes the local outbox so the race can no longer happen on the sender side.

  ## Changes

  ### Diagnostic and recovery tooling (commits `07c0031e`, `ec0a9eb0`)

  - **Per-recipient delivery indicators on Group Settings.** Up to two small grey check
    icons appear inline next to each participant's name — one per group file the current
    user authored (main conversation file + admin file). Red `ErrorOutline` if the server
    reports anything other than `Delivered`. Hidden entirely if the current user didn't
    author either file. Reuses the existing `MessageInfoViewModel` transfer-history loader
    pattern (`driveFileProvider.getTransferHistory(chatTargetDrive.alias, fileId)` with the
    same status mapping).
  - **"Heal group" button** at the bottom of Group Settings. Visible only if the current
    user authored at least one of the two group files. Tap → service redistributes the
    file(s) the user authored to all current participants via `updateConversationInternal` /
    `updateAdminFile`. Result snackbar; `isHealing` spinner during the call.
  - **`ConversationService.healGroupDistribution(conversationId)`** — public entry on the
    service. Per-file authorship check (`fileMetadata.originalAuthor ?: senderOdinId ==
    domain`); skips files the user didn't author; full debug logging at every step
    (versionTag, isPending, recipients, enqueue result).
  - **Logging gaps closed.** `uploadAdminFile` and `updateAdminFile` previously discarded
    the `outboxSync.tryEnqueue` boolean; now they `Logger.w` when enqueue returns false
    (UNIQUE conflict / silent drop). `loadTransferHistory` in the ViewModel now logs
    per-recipient state, not just counts, so a `BEFORE heal` / `AFTER heal` snapshot is a
    diff in the log.

  ### Outbox dependency-chain serialization

  #### Service-driven enqueues (commits `8dedb8f3`, `ec78284b`)

  `ConversationService` now produces a strictly-serial chain in three places:

  1. **`createConversation` (group path):** conversation file (no dep) → admin file (dep =
     `conversationId`) → `GroupConversationStarted` status (dep =
     `ChatProtocol.getAdminFileUniqueId(conversationId)`). Previously the admin file had no
     dep at all and the status message chained directly off the conversation file, so admin
     + status raced once the conversation file was acknowledged.
  2. **`updateAdmins`:** admin-file update row (dep = `conversationId`); first
     add/remove-status message seeded with `previousMessageUniqueId =
     adminFileUniqueId`; subsequent status messages chain on the previous one's id.
  3. **`updateGroupMembers`:** trailing removed-status messages → conversation-file update
     (dep = last removed-status `messageId`, or null when there are no removals) → added-
     status messages chain on `conversationId` (existing behavior, preserved).

  #### User-typed messages (commit `b439b3b9`)

  `ChatMessageSenderService.sendMessageInternal` now maintains an in-memory
  `Map<conversationId, lastEnqueuedMessageId>` (mutex-guarded) and falls back through three
  levels when the caller passes `previousMessageUniqueId = null` (which is the common
  case from `ConversationListViewModel.addMessage` and friends):

  effectiveDep = previousMessageUniqueId           // explicit caller wins
    ?: lastEnqueuedPerConversation[conversationId]  // chain on prior in this convo
    ?: conversationId                                // first message in convo: chain on convo file

  This means three messages typed in airplane mode release through the local outbox as
  `A → B → C`, in send order, rather than fanning out in parallel the moment the
  conversation file drains. The `NOT EXISTS` guard in the outbox `checkout` SQL means the
  fallback to `conversationId` resolves immediately for active conversations (the
  conversation file row is gone), so there is no overhead in the common case.

  ### Refactor (commit `b439b3b9`)

  To make `ChatMessageSenderService` testable in isolation, two narrow read-side seams were
  extracted:

  - **`ConversationParticipantLookup`** (new) — `getConversationById` + `getRecipients`.
    Implemented by `ConversationStream`; the sender now depends on the interface.
  - **`MessageLookup`** (existing) extended with `getMessageFile` and `loadFullMessage` so
    the sender's edit/reply paths can also be mocked. `ChatMessageStream` already implements
    `MessageLookup`; the two new methods are marked `override`.
  - DI binding added: `single<ConversationParticipantLookup> { get<ConversationStream>() }`.

  No behavioral change to production wiring — Koin resolves the same singleton.

  ## Tests

  - **`ConversationServiceOutboxOrderingTest`** (new in `380b4a39`) — six static-shape tests
    on the chain produced by `createConversation` (1:1 and group), `updateAdmins`, and
    `updateGroupMembers` (full flow, no removals, only removals). Uses the existing
    `ConversationServiceTestFixture` plus a new `drainOutboxInDependencyOrder()` helper that
    iteratively checks-out-and-deletes so chained rows surface (the existing `drainOutbox`
    only returns chain heads).
  - **`FakeStatusMessageSender`** extended in the same fixture to record `messageUniqueId`
    and `previousMessageUniqueId`. Backwards-compatible: existing tests read only the
    preserved fields.
  - **`ChatMessageSenderServiceOutboxOrderingTest`** (new in `b439b3b9`) — four tests on
    the per-message chain in `sendMessageInternal`:
    1. First message in a brand-new conversation chains on `conversationId`.
    2. Three sequential messages with `previousMessageUniqueId = null` chain on each
       other (`B.dep == A.uniqueId`, `C.dep == B.uniqueId`) — the airplane-mode pile-up
       case.
    3. Caller-supplied `previousMessageUniqueId` overrides the in-memory fallback.
    4. Chains are tracked per-conversation — sends to two parallel conversations don't
       cross-contaminate.
  - **`ChatMessageSenderServiceTestFixture`** (new) — minimal fixture mirroring the
    `ConversationServiceTestFixture` pattern: real offline `OutboxSync` + in-memory
    SQLite + real `OptimisticWriter`; fakes for `ConversationParticipantLookup`,
    `MessageLookup`, `PayloadBundleEncryptor`; `MockEngine`-backed `HttpClient` for the
    `DriveFileProvider` that the send path doesn't actually touch.

  ## Known caveats

  - **App restart with pending chained messages.** The in-memory `lastEnqueuedPerConversation`
    map is cleared on restart; pending outbox rows still carry their persisted deps and
    drain in order among themselves, but a new message after restart starts a fresh chain
    (fallback to `conversationId`). Realistically only matters if the user is offline
    through a restart.
  - **Truly concurrent sends from two coroutines** for the same conversation could read
    the same predecessor under the current mutex granularity (lock wraps map read and map
    write separately, not the entire enqueue). Sequential sends — the common case via
    `viewModelScope.launch` from `addMessage` — are unaffected.
  - **Existing stuck conversations** (e.g. Shelly's `0e684619`) are not auto-fixed by this
    change. Pressing "Heal group" from the original author's app is the empirical
    recovery path; whether it succeeds depends on whether the recipient's server accepts
    Michael's `UpdateFileByUniqueId` despite the local placeholder having a
    recipient-minted `versionTag`. Untested at the server protocol level.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/groupsettings/GroupSettingsScreen.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/groupsettings/GroupSettingsScreen.kt
@@ -1,27 +1,39 @@
 package id.homebase.chat.groupsettings
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.Logout
 import androidx.compose.material.icons.filled.Add
+import androidx.compose.material.icons.filled.Check
+import androidx.compose.material.icons.filled.CheckCircle
 import androidx.compose.material.icons.filled.ChevronLeft
 import androidx.compose.material.icons.filled.DoNotDisturbOn
+import androidx.compose.material.icons.filled.ErrorOutline
+import androidx.compose.material.icons.filled.Healing
 import androidx.compose.material.icons.filled.Key
 import androidx.compose.material.icons.filled.KeyOff
 import androidx.compose.material.icons.outlined.Edit
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.ui.Alignment
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
+import androidx.compose.material3.LocalContentColor
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.ModalBottomSheet
 import androidx.compose.material3.Scaffold
@@ -36,16 +48,24 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalUriHandler
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
 import androidx.compose.ui.window.Dialog
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import id.homebase.api.common.OdinId
 import id.homebase.chat.createconversation.ContactItem
 import id.homebase.chat.services.convo.contact.ContactConnectionState
 import id.homebase.chat.widget.AvatarNameDisplay
 import id.homebase.chat.widget.ErrorInfoItem
 import id.homebase.chat.widget.LoadingListItem
+import id.homebase.core.avatars.AvatarOptions
+import id.homebase.core.avatars.ContactAvatar
 import id.homebase.core.avatars.ConversationAvatarModel
+import id.homebase.core.widget.ContactName
 import id.homebase.core.widget.DialogButtons
 import id.homebase.core.widget.DialogCard
 import id.homebase.core.widget.DialogText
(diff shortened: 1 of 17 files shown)


Seifert69:
The group heal button and diagnostics logs are presumably temporary. Maybe the group member checks should stay?

PR #408, opened 2026-05-02
Chat infra: local-only conversation recovery + outbox permanent-failure drop

Seifert69:
ConversationService.recoverConversation
- Strictly local-only now: writes a placeholder via OptimisticWriter.writeLocalOnlyConversationPlaceholder, no outbox enqueue, no peer transit. Server distribution is reserved for the explicit "Heal Group" button (healGroupDistribution); pushing a placeholder pre-emptively to the server has been observed to make the recipient's local state overwrite later heals, defeating the heal flow.
- When an existing broken local file is present, it's deleteBy()- removed first so the placeholder (updated=ZeroTime) can land despite the modified-timestamp guard in baseUpsertEntryZapZap.
- originalAuthor is now nullable: matches ConversationStream's placeholder behaviour for self-authored group orphans whose stored originalAuthor is null. Hook signature on ConversationStream (onRecoverConversation) widened to match.
- Self-only 1:1 placeholder fallback: when participants reduce to [self] (e.g. recovering a 1:1 the user authored themselves), the placeholder is forced to be a group titled "1:1 repair" so ConversationMapper's group branch maps it cleanly. Without this the mapper's `participants.first { it != domain }` throws.
- MethodAudit instrumentation is preserved through the rewrite.

ConversationService.deleteConversation
- Local-only placeholder branch detects versionTag==null and bypasses the failing UpdateFile call, deletes the local row directly via dbm.driveMainIndex.deleteBy and removes the in-memory entry via the new ConversationLoader.removeConversation hook so the UI updates immediately.
- Without this, clicking delete on a recovered placeholder enqueued an UpdateFile that the server rejects (no such file, or versionTag=null) — the outbox fix below stops the retry storm, but only this branch makes the delete actually take visible effect locally.

OptimisticWriter
- writeLocalOnlyConversationPlaceholder accepts an optional title so recovery placeholders can be labelled (e.g. "1:1 repair") while existing callers keep their blank-title default.

ConversationStream
- onRecoverConversation hook signature widened to OdinId?.
- New removeConversation(conversationId) implementation drops the in-memory entry; called from deleteConversation's local-only placeholder branch.

ConversationServiceCollaborators
- Adds removeConversation() to the ConversationLoader interface.

OutboxSync.isPermanentFailure
- Drops items immediately on: NotFoundException, ClientException with errorCode in {FileNotFound, MissingVersionTag, CannotOverwriteNonExistentFile, UnknownId}, plus message-text fallback for the "Could not find file" / "Missing version tag" 400s the server emits as UnhandledScenario. Avoids ~48h of exponential-backoff retry storms for stale or local-only-placeholder uploads.


Seifert69:
AFAIR the originalAuthor is set once and never changes. The senderId is set by your receiving host when you receive a message over Peer. OriginalAuthor and SenderId might be different if you e.g. receive a forwarded message.

PR #428, opened 2026-05-05
refactor(chat): hoist composer previews into a staged-attachments list

Seifert69:
The chat composer carried link previews via dedicated `linkPreviewData` / `onCancelLinkPreview` parameters threaded through `MessageInputBar`, `MessageTextFieldExpanded`, `MessageTextFieldCompact`, and a `linkPreview: LinkPreview?` field on `ConversationListUiAction.SendMessage`
+ `addMessage`/`replyToMessage`. Adding a second preview kind would have meant repeating that plumbing on every layer.

Replace it with a hoisted `List<StagedAttachment>` slot:

- New `StagedAttachment` sealed interface in `services/staged/`, with a single concrete case today (`StagedLinkPreview`). Stable `id` lets the cancel-X button and the URL-detector deduplicate by identity.
- New `List<StagedAttachment>.toCombinedPayloadBundle(fileOps)` extension — one `when` dispatch maps each kind to its existing `XyzPayloadBuilder`. Replaces the per-kind `?.let { … }` paths in the ViewModel.
- New `StagedAttachmentRow` Compose dispatcher — one `when` branch per kind, delegates to the existing `LinkPreviewCard(isCompact = true, …)`.
- `MessageInputBar` no longer owns `linkPreviewData`. The URL-detector `LaunchedEffect` now produces a `StagedLinkPreview` into the parent-owned list via `onStagedAttachmentsChange`. `cancelledUrls` / `lastFetchedUrl` / 250 ms debounce semantics preserved verbatim.
- `MessageTextFieldExpanded` / `MessageTextFieldCompact` lose their two preview params and gain `stagedAttachments` + `onCancelAttachment(id)`. Render block is a single `forEach { StagedAttachmentRow(…) }`.
- `showSendButton` predicate gains `|| stagedAttachments.any { it !is StagedLinkPreview }` so future user-initiated kinds (location, contact, …) enable send without text. Link-preview-only behaviour is unchanged.
- `ConversationContent` owns `stagedAttachments` at the top level so user-initiated flows (e.g. an attachment-sheet location picker) can append from outside the input bar. Cleared on send.
- `ConversationListUiAction.SendMessage` swaps `linkPreview` for `stagedAttachments: List<StagedAttachment>`. `addMessage` and `replyToMessage` lose their `linkPreview` param and call `toCombinedPayloadBundle` once.

Adding the next attachment kind is now a 3-line change: a sealed-interface case, a `when` branch in the row dispatcher, a `when` branch in the bundle combiner. No new composer params, no new ViewModel branches, no new UiAction fields.

Wire format unchanged. Receiver render unchanged. Behaviour preserved (text-only send, URL-detected link preview, cancel-then-no-refetch, URL replacement, edit-message flow). Builds clean on JVM, Android, and iOS; JVM tests pass.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiAction.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiAction.kt
@@ -2,8 +2,8 @@ package id.homebase.chat.conversationlist
 
 import id.homebase.api.client.KeyHeader
 import id.homebase.api.client.drives.files.PayloadDescriptor
-import id.homebase.api.client.link.LinkPreview
 import id.homebase.api.common.OdinId
+import id.homebase.chat.services.staged.StagedAttachment
 import id.homebase.chat.data.ConversationUiModel
 import id.homebase.chat.data.MessageUiModel
 import id.homebase.core.gallery.GalleryImage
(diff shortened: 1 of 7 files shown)


Seifert69:
DONT MERGE - THIS WAS JUST THE CONCEPT

PR #430, opened 2026-05-05
Chat share location and refactor to make new features easier

Seifert69:
Summary
                                                                                                                                                                                          
  Adds Share Location to chat. Sending is supported on iOS and Android only; Desktop is display-only. The static map preview is fetched server-side (dev stub today; backend swap
  planned), embedded encrypted in the chat payload, and rendered by the existing MediaItem dispatch on every receiver platform.                                                           
  
  This PR has two commits, designed to be reviewed in order:                                                                                                                              
                                                         
  1. refactor(chat): hoist composer previews into a staged-attachments list — pure refactor, no behaviour change. Generalises the link-preview slot so the composer can carry any kind of 
  staged content.                                        
  2. feat(chat): share-location attachment — the new feature, built on top of the abstraction.                                                                                            
                                                                                                                                                                                          
  The split is the point: the second commit measures what the first one bought us. Adding location touches zero new lines on MessageInputBar, ConversationListUiAction, or                
  ConversationListViewModel — the abstraction absorbed all of it.                                                                                                                         
                                                                                                                                                                                          
  Commit 1 — staged-attachments refactor                                                                                                                                                  
  
  Before, link previews were threaded through the composer as dedicated parameters: linkPreviewData: LinkPreview?, onCancelLinkPreview: () -> Unit on three composables (MessageInputBar, 
  MessageTextFieldExpanded, MessageTextFieldCompact), plus a linkPreview: LinkPreview? field on ConversationListUiAction.SendMessage and matching params on addMessage / replyToMessage.
  Adding a second preview kind would have repeated that plumbing on every layer.                                                                                                          
                                                         
  Now there's a single List<StagedAttachment> slot:                                                                                                                                       
  
  - StagedAttachment sealed interface in services/staged/, with one concrete case in this commit (StagedLinkPreview). Stable id lets the cancel-X button and the URL-detector deduplicate 
  by identity.                                           
  - List<StagedAttachment>.toCombinedPayloadBundle(fileOps) extension function — one when dispatch maps each kind to its existing XyzPayloadBuilder.build. Replaces the per-kind ?.let { …
   } paths in the ViewModel.                                                                                                                                                              
  - StagedAttachmentRow Compose dispatcher — one when branch per kind, delegates to the existing LinkPreviewCard(isCompact = true, …).
  - MessageInputBar no longer owns linkPreviewData. The URL-detector LaunchedEffect produces a StagedLinkPreview into the parent-owned list via onStagedAttachmentsChange. cancelledUrls /
   lastFetchedUrl / 250 ms debounce semantics preserved verbatim.                                                                                                                         
  - Inner composables lose their two preview params, gain stagedAttachments + onCancelAttachment(id). Render block is one forEach { StagedAttachmentRow(…) }.                             
  - showSendButton predicate gains || stagedAttachments.any { it !is StagedLinkPreview } so future user-initiated kinds enable send without text. Link-preview-only behaviour unchanged.  
  - ConversationContent owns stagedAttachments at the top level so user-initiated flows (e.g. an attachment-sheet location picker) can append from outside the input bar.                 
  - ConversationListUiAction.SendMessage swaps linkPreview for stagedAttachments: List<StagedAttachment>. addMessage and replyToMessage lose their linkPreview param and call             
  toCombinedPayloadBundle once.                                                                                                                                                           
                                                                                                                                                                                          
  Adding the next attachment kind is now: add a sealed-interface case, a when branch in the row dispatcher, a when branch in the bundle combiner. No new composer params, no new ViewModel
   branches, no new UiAction fields.                     
                                                                                                                                                                                          
  Wire format unchanged. Receiver render unchanged. All link-preview behaviour preserved (text-only send, URL-detected preview, cancel-then-no-refetch, URL replacement, edit-message     
  flow).
                                                                                                                                                                                          
  Commit 2 — Share Location feature                                                                                                                                                       
  
  [iOS/Android picker] tap "Share location" in attachment sheet                                                                                                                           
     │                                                                                                                                                                                    
     ├─> rememberCurrentLocationLauncher → request perm → FusedLocationProvider / CLLocationManager
     │     (one-shot fix, BALANCED_POWER_ACCURACY / kCLLocationAccuracyHundredMeters)                                                                                                     
     │                                                                                                                                                                                    
     ├─> LocationPreviewProvider.getLocationPreview(lat, lon)         [DEV STUB — see Privacy below]
     │     ① GET nominatim.openstreetmap.org/reverse  → { display_name }                                                                                                                  
     │     ② GET staticmap.openstreetmap.de/staticmap.php → PNG bytes                                                                                                                     
     │     ③ return LocationPreview(lat, lon, address, "data:image/png;base64,…", w, h)                                                                                                   
     │                                                                                                                                                                                    
     ├─> stagedAttachments += StagedLocationPreview(preview)                                                                                                                              
     │     (composer renders/cancels via the abstraction — no composer code knows location exists)                                                                                        
     │                                                                                                                                                                                    
     └─> Send → LocationPreviewPayloadBuilder → encrypted payload                                                                                                                         
           { key:  "chat_location"                                                                                                                                                        
             descriptorContent JSON: { lat, lon, address, hasImage, w, h }                                                                                                                
             bytes: encrypted PNG }                                                                                                                                                       
                                                                                                                                                                                          
  Receivers (any platform, including Desktop):                                                                                                                                            
                                                                                                                                                                                          
  MediaItem.kt when → PAYLOAD_KEY_LOCATION → LocationPreviewCard                                                                                                                          
     tap → geo:lat,lon?q=lat,lon(address) → user's own map app                                                                                                                            
                                                                                                                                                                                          
  Wire format                                                                                                                                                                             
                                                                                                                                                                                          
  ┌───────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┐
  │             Field             │                                        Value                                        │
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ appData.fileType              │ 7878 (MessageFileType) — unchanged, location is a regular chat message              │
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ appData.dataType              │ 0 — unchanged                                                                       │                                                                 
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ appData.content               │ only the user's caption (or empty) — no location keys here                          │                                                                 
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ payloads[0].key               │ "chat_location" (PAYLOAD_KEY_LOCATION) — the discriminator                          │
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ payloads[0].descriptorContent │ small JSON: LocationPreviewDescriptor { lat, lon, address, hasImage, w, h }         │
  ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤                                                                 
  │ payloads[0] body              │ the static map PNG, encrypted via the same key-header pipeline as image attachments │
  └───────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────┘                                                                 
                                                         
  This mirrors chat_links exactly. Old clients receiving an unknown payload key fall through to the placeholder bubble; new clients render the rich card. No protobuf changes, no DB      
  migrations, no version negotiation.                    
                                                                                                                                                                                          
  Privacy posture                                        

  ┌──────────────────────────────────────────────────────┬─────────────────────────────────────────┬─────────────────────────────────────────┬─────────────────────────┐                  
  │                                                      │              Sender today               │        Sender after backend swap        │        Receiver         │
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────┤                  
  │ Contacts OSM (Nominatim, staticmap.openstreetmap.de) │ ✅ yes (dev leak)                       │ ❌ no (backend proxies)                 │ ❌ no                   │
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────┤
  │ Contacts Google / Mapbox                             │ ❌ no                                   │ ❌ no                                   │ ❌ no                   │                  
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────┤                  
  │ Static map travels via                               │ inline base64 → encrypted drive payload │ inline base64 → encrypted drive payload │ encrypted drive payload │                  
  └──────────────────────────────────────────────────────┴─────────────────────────────────────────┴─────────────────────────────────────────┴─────────────────────────┘                  
                                                         
  LocationPreviewProvider.getLocationPreview is the only function that needs to change when the backend ships /api/v2/preview/staticmap. Function signature, return type, and every       
  downstream caller stay identical. The receiver pipeline is already final-state — the PNG lives in the encrypted drive payload, never as a third-party URL.
                                                                                                                                                                                          
  A comment in LocationPreviewProvider.kt explicitly marks the dev-stub body and the migration target.                                                                                    
   
  Notable design choices                                                                                                                                                                  
                                                         
  - GPS via expect/actual platform APIs (FusedLocationProviderClient on Android, CLLocationManager on iOS, no-op stub on JVM). LivotovLabs/Anchor was evaluated but is pre-1.0 and lacks a
   documented one-shot API; revisit when the future 24/7 background-tracking feature is built — Anchor's foreground-service + Always-permission model is well-suited to that.
  - Picker is "send my current location" only — no draggable map. Avoids pulling Google/Apple Maps SDKs onto device and matches the project's privacy posture. Future iterations could    
  layer MapLibre on top.                                                                                                                                                                  
  - Permissions: ACCESS_FINE_LOCATION + ACCESS_COARSE_LOCATION on Android (no BACKGROUND_LOCATION); NSLocationWhenInUseUsageDescription on iOS (no Always). The future 24/7 feature will
  need the always-on equivalents.                                                                                                                                                         
  - Location is a "user-initiated" staged attachment, so it enables the send button even with no text caption (link previews alone don't, by design — they ride along with text).
                                                                                                                                                                                          
  Files                                                                                                                                                                                   
                                                                                                                                                                                          
  New (15 files):                                                                                                                                                                         
  - homebase-api/.../client/location/LocationPreview.kt — sender-side data class
  - homebase-api/.../client/location/LocationPreviewProvider.kt — dev-stub OSM fetcher with throttle + in-memory cache                                                                    
  - homebase-chat/.../services/builder/LocationPreviewDescriptor.kt — wire descriptor (small JSON in payload header)  
  - homebase-chat/.../services/builder/LocationPreviewPayloadBuilder.kt — preview → encrypted PayloadBundle                                                                               
  - homebase-chat/.../widget/LocationPreview.kt — sender + receiver LocationPreviewCard composables                                                                                       
  - homebase-chat/.../location/LocationFix.kt — common GPS fix data class                                                                                                                 
  - homebase-chat/.../location/LocationLauncher.kt — expect class + rememberCurrentLocationLauncher                                                                                       
  - homebase-chat/.../androidMain/.../LocationLauncher.android.kt — FusedLocationProviderClient + permission launcher                                                                     
  - homebase-chat/.../nativeMain/.../LocationLauncher.native.kt — CLLocationManager delegate                                                                                              
  - homebase-chat/.../jvmMain/.../LocationLauncher.jvm.kt — no-op stub (button hidden on Desktop via isMobile())                                                                          
  - services/staged/StagedAttachment.kt, StagedAttachmentBundle.kt, widget/StagedAttachmentRow.kt — the abstraction (commit 1)                                                            
                                                                                                                                                                                          
  Modified:                                                                                                                                                                               
  - ChatProtocol.kt — +1 line for PAYLOAD_KEY_LOCATION                                                                                                                                    
  - MediaItem.kt — when branch deserializes descriptor and renders LocationPreviewCard                                                                                                    
  - MessageContentLabel.kt — reply/list-preview label                                 
  - AttachmentOptions.kt — uncomments the existing Location entry, gated by isMobile()                                                                                                    
  - ApiModule.kt — singleOf(::LocationPreviewProvider)                                                                                                                                    
  - ConversationContent.kt — GPS launcher, fetch coroutine, append into the existing stagedAttachments list                                                                               
  - MessageInputBar.kt / ConversationListUiAction.kt / ConversationListViewModel.kt — only the refactor commit touches these; commit 2 leaves them alone                                  
  - homebase-common/.../strings.xml — 4 user-facing strings                                                                                                                               
  - gradle/libs.versions.toml + homebase-chat/build.gradle.kts — play-services-location 21.3.0 on androidMain only                                                                        
  - androidApp/.../AndroidManifest.xml — ACCESS_FINE_LOCATION + ACCESS_COARSE_LOCATION                                                                                                    
  - iosApp/iosApp/Info.plist — NSLocationWhenInUseUsageDescription                                                                                                                        
                                                                                                                                                                                          
  Verification                                                                                                                                                                            
                                                                                                                                                                                          
  - ✅ ./gradlew homebase-chat:compileKotlinJvm homebase-chat:compileAndroidMain homebase-chat:compileKotlinIosSimulatorArm64                                                             
  - ✅ ./gradlew homebase-chat:jvmTest homebase-api:jvmTest
                                                                                                                                                                                          
  Test plan                                              
                                                                                                                                                                                          
  - Link previews still work (regression check on commit 1): paste a URL → preview card appears → Send → message + link bubble arrive on receiver. Cancel-X then retype same URL → preview
   does NOT reappear (cancellation memory works). Modify URL mid-text → old preview hides, new one appears.
  - Android send → Android receive: tap attachment → "Share location" → grant permission → confirm card shows address + map → Send. Bubble appears on receiver; tap opens Google Maps at  
  the coordinate.                                                                                                                                                                         
  - iOS send → iOS receive: same flow via Xcode-installed build.
  - iOS send → Desktop receive (display-only): location bubble renders correctly on Desktop; "Share location" entry is not in the desktop attachment sheet.                               
  - Permission denial: deny location permission on Android/iOS → composer surfaces the in-flight cancellation gracefully; no crash.                                                       
  - Combined link + location: paste a URL → wait for link card → trigger location → both staged → Send → both payloads arrive on the receiver, each rendered by its own MediaItem branch. 
  - Cancel-X on the staged location card: removes only the location, leaves any link preview intact.                                                                                      
  - Privacy spot-check (dev stub phase): adb shell run-as id.homebase.feed.dev cat files/logs/homebase.log | grep -iE 'google\|mapbox\|gms\|googleapis' returns zero hits during a        
  location share. Hits to nominatim.openstreetmap.org and staticmap.openstreetmap.de are expected and documented.                                                                         
                                                                                                                                                                                          
  Explicitly NOT in this PR                                                                                                                                                               
                                                         
  - 24/7 background tracking (would need a foreground service on Android + Always permission flow on iOS — re-evaluate LivotovLabs/Anchor for that feature when it lands; pre-1.0 today,  
  no documented one-shot API).
  - Live-location updates ("share location for 15 minutes"). Static pin only.                                                                                                             
  - Interactive map picker (drag a pin). Picker is "send my current location" with a confirm card.                                                                                        
  - Desktop sending (intentional: Desktop never asks for GPS).                                                                                                                            
  - Backend /api/v2/preview/staticmap endpoint (separate backend repo; this PR's LocationPreviewProvider is the sole client-side swap site).                                              
                                                                                                                                                                                          
  Merge guidance                                                                                                                                                                          
                                                                                                                                                                                          
  Either squash-and-merge (collapses to one commit, simpler git log) or rebase-and-merge (preserves the two-commit story for git blame clarity) work. Avoid the merge-commit option since 
  it adds noise without value here.
                                                                                                                                                                                          
  🤖 Generated with Claude Code (https://claude.com/claude-code)


toddmitchell (commented):
could be good to use a data-type on the chat message when you share location. This would just make it easier to query all messages that have a datatype == 449 so we can see the gps info.

(same goes for other attachemnts too like linkpreview, etc.)

toddmitchell (commented):
On Desktop - found a crash

  Crash: IllegalArgumentException: Destination with route Settings cannot be found in 
  navigation graph. User clicked their own avatar → ConversationListPane.kt:171 →       
  onNavigateToSettingsScreen → navController.navigate(Route.Settings) at                
  AppNavHost.kt:407.   


More details a la claude
****

Root cause: a missing } in AppNavHost.kt. Look at the structure starting at line 558: 
                                                                                        
  composable<Route.ConversationSettings> {                                              
      if (isAuthenticated) {      
          ConversationSettingsScreen(...)                                               
      }                                       // line 567 — closes the if

      composable<Route.GroupSettings>  { ... }     // line 569                        
      composable<Route.GroupAddMembers> { ... }                                         
      composable<Route.GroupEdit>      { ... }                                          
      composable<Route.Examples>       { ... }                                          
      composable<Route.Settings>       { ... }     // line 611                          
      // … NotificationSettings, AppearanceSettings, StorageSettings too              
                                                                                        
  Route.ConversationSettings's lambda body is never closed before the next            
  composable<...> calls. Those nested composable<...> calls compile because composable  
  resolves up to the outer NavGraphBuilder receiver via Kotlin's nested-receiver scoping
   — but they only execute (and therefore only register their destinations on the graph)
   when Route.ConversationSettings's content lambda runs, i.e. only after the user has  
  navigated to ConversationSettings at least once.

  You never visited ConversationSettings in this session, so Route.Settings (and
  GroupSettings, GroupAddMembers, GroupEdit, Examples, NotificationSettings,          
  AppearanceSettings, StorageSettings) were never added to the graph. Clicking the      
  avatar → navigate("Settings") → no such destination → throw.
                                                                                        
  That's also why the error message reports startDestination=app-loading (the outer
  graph root) — it's saying "Settings isn't anywhere in the graph rooted there", which
  is literally true at that moment.                                                     
                                  
  The lambda needs to be closed after line 567 so the rest of the composable<...> calls 
  live as siblings on the NavGraphBuilder and get registered eagerly when the NavHost is
   composed.                                                                            
                                  
  Sanity-check side note: this isn't introduced by the share-location branch (the two   
  commits here only touch chat composer + payload code, not AppNavHost). It's a       
  pre-existing structural bug that just happens to bite when the avatar click path is   
  exercised on desktop.

Seifert69:
Thanks. (I haven't run the code yet, it was the concept around the composer). I'll fix it up. :-)

PR #433, opened 2026-05-06
heal group via status message + receiver-side cleanup

Seifert69:
Reworks the "Heal Group" flow so it can repair recipients whose local conversation/admin file diverged from the canonical author's copy (e.g. clients that mistakenly created their own copy with themselves as originalAuthor and a fresh versionTag — those rejected every subsequent peer push from the true author).

Send side (ConversationService.healGroupDistribution):
  Each click now sends a GroupHealRequested chat status message before
  re-pushing the conversation + admin files. The status carries the
  canonical group identity (originalAuthor, versionTag, title,
  participants, admin uniqueId, admin versionTag) grouped in a nested
  GroupHealInfo. Authorship gate kept — only the original author of the
  main file may announce canonical identity.

Receive side (ConversationService.handleIncomingHealRequest, dispatched from ConversationStream.processMessageBatchIncrementally on live BatchReceived only):
  Forgery / self-loopback / Left-Removed-RejoinPending guards.
  Compares local main + admin files against canonical fields; placeholder
  rows (versionTag == null) are left alone. Hard-deletes the broken
  file(s) on the recipient's own server via DeleteLocalFilesByFileIdRequest
  with recipients=null and hardDelete=true (no peer fan-out, no
  soft-delete row left behind).
  For broken main, writes a versionTag=null placeholder via a new
  writeOrReplaceConversationPlaceholder helper extracted from
  recoverConversation. For broken admin, deletes the local row only —
  getAdmins() falls back to originalAuthor in the gap.
  Surfaces the cleanup as a local-only GroupHealLocalCleanup status so
  the user sees what was removed.

UX: user clicks Heal twice. First click cleans up bad recipient-side files (the file push itself fails for those recipients). Second click's push lands cleanly because the recipient's server slot is empty and the local placeholder has versionTag=null. No hidden retry.

Render side (ChatMessageStream.renderStatusMessage): branches for both new enum values, with three localized cleanup variants (both / main / admin). Strings added in homebase-common.

DI: conversationStream.onIncomingHealRequest wired to conversationService.handleIncomingHealRequest in AppModule.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageStream.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageStream.kt
@@ -51,6 +51,11 @@ import id.homebase.resources.system_group_conversation_member_left
 import id.homebase.resources.system_group_conversation_member_left_you
 import id.homebase.resources.system_group_conversation_started
 import id.homebase.resources.system_group_conversation_started_you
+import id.homebase.resources.system_group_heal_local_cleanup_admin
+import id.homebase.resources.system_group_heal_local_cleanup_both
+import id.homebase.resources.system_group_heal_local_cleanup_main
+import id.homebase.resources.system_group_heal_requested
+import id.homebase.resources.system_group_heal_requested_you
 import kotlinx.collections.immutable.toPersistentList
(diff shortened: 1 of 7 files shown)


Seifert69:
I'll need to get this live on test flight until I can finish debugging

PR #474, opened 2026-05-09
fix(chat): cap reactions at 5 per user, mirror server limit client-side

Seifert69:
The server enforces max 5 reactions per user per message and rejects the upload of any 6th with a 400 / UnhandledScenario / "Too many Reactions count N in ReactionSummary max 5". The client had no matching guard, so it would optimistically write the 6th, enqueue it to the outbox, then surface a generic upload error while local state diverged from server state until the next sync.

Mirror the rule in ConversationListViewModel's ToggleReaction handler: before any optimistic write or service call, if the candidate emoji is not already in the message's ownReactions and ownReactions.size is at the cap, emit a localized info snackbar
(chat_reactions_limit_reached) and return. Removes are never blocked
- they can only reduce the count.

ownReactions on MessageUiModel is decoded by ChatMessageStream into bare emoji strings (the wire form is JSON {"emoji":"X"}); the guard compares action.reaction directly against that list so removes via the ReactionsBottomSheet "slider panel" don't get misread as adds and hit the same snackbar. The KDoc on the field claimed "Raw JSON-encoded reactions" - corrected to match what ChatMessageStream actually produces.

The single source of truth for the cap is a top-level MAX_REACTIONS_PER_USER_PER_MESSAGE = 5 next to
ChatMessageActionService.toggleReaction; no service-layer change since the VM is the only caller. Stale-mirror drift needs no extra defense - the existing rollback path on outbox-enqueue failure plus sync re-alignment is sufficient.

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/QueryBatch.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/QueryBatch.kt
@@ -1,6 +1,7 @@
 package id.homebase.api.sync.database
 
 import app.cash.sqldelight.db.QueryResult
+import co.touchlab.kermit.Logger
 import id.homebase.api.common.IntRange
 import id.homebase.api.common.time.UnixTimeUtc
 import id.homebase.api.common.time.UnixTimeUtcRange
@@ -160,6 +161,21 @@ class QueryBatch(
 
         if (fileSystemType == null) {
             throw IllegalArgumentException("fileSystemType required in Query Batch")
+        }
+
+        // Smoke alarm for "single-uniqueId" misuse: queryBatchAsync sorts and
+        // scans the whole drive index; for a single uniqueId, callers should
+        // use DriveMainIndexWrapper.selectHomebaseFileByUnique instead. We've
+        // had this cost ~500ms latency on the chat reaction/delete/get paths.
+        // Stack trace surfaces the offender directly in shipped logs.
+        if (uniqueIdAnyOf?.size == 1) {
+            Logger.w(tag = "QueryBatch") {
+                val callsite = Throwable().stackTraceToString()
+                    .lineSequence().drop(1).take(6).joinToString("\n")
+                "queryBatchAsync called with a single uniqueId — prefer " +
+                        "DriveMainIndexWrapper.selectHomebaseFileByUnique. " +
+                        "callsite:\n$callsite"
+            }
         }
 
         if (noOfItems < 1) {
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
@@ -51,6 +51,7 @@ import id.homebase.chat.conversationlist.ConversationListUiEvent.ShowInfoMessage
 import id.homebase.chat.data.ConversationState
 import id.homebase.chat.data.MessageUiModel
 import id.homebase.chat.services.ChatMessageActionService
+import id.homebase.chat.services.MAX_REACTIONS_PER_USER_PER_MESSAGE
 import id.homebase.chat.services.ChatMessageSenderService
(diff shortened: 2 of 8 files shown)


Seifert69:
Tested and confirmed working in the Desktop client.

Ran a local /review post-push that revealed a number of issues with the proposed solution. Second commit coming

Seifert69:
Tested on Desktop client — works as expected: 6th-emoji guard fires, removes flip the bubble within a frame, single-emoji adds are no longer ~500 ms,
  delete is visibly snappier, and the group "1→0→2" aggregate flicker is gone.

  This branch picked off a cluster of pre-existing chat-reaction bugs that were always in the code — we fixed several here, the rest are queued as
  follow-ups.

  ### Fixed in this PR (all pre-existing)

  - Client-side cap of 5 reactions per user per message — server already enforced this with `400 / UnhandledScenario / "Too many Reactions"`; client now
  blocks the 6th before the optimistic write or outbox enqueue. Snackbar surfaces the limit.
  - Read-modify-write race in `OptimisticWriter.writeReactionToggle` — five rapid taps had five concurrent coroutines clobbering each other; per-(driveId,
  uniqueId) `Mutex` serializes the read-modify-write. Diagnostic test confirmed it empirically.
  - Optimistic toggle now writes the per-user mirror `localAppData.localReactions`, not just the aggregate `reactionPreview`. Previously the mirror only
  refreshed on the server round-trip, which broke both the cap guard's freshness and a latent group-chat bug.
  - Group-aggregate decrement bug — `isAdding` was driven by aggregate `reactionPreview` count, so toggling an emoji another user already had was misread as
   a remove. Now driven by the user's own `localReactions`.
  - ~500 ms latency on react / delete / `getReactions`. `requireFileId(messageId)` was running `QueryBatch(noOfItems = 1000, NewestFirst)` over the whole
  chat drive just to translate one uniqueId → fileId. Rewritten to a single `selectHomebaseFileByUnique` call. `toggleReaction` additionally reuses the
  optimistic writer's returned `original.fileId`, paying zero extra DB cost.
  - Stale KDoc on `MessageUiModel.ownReactions` (claimed "raw JSON-encoded"; it's bare emoji).

  ### Known follow-ups (pre-existing, NOT introduced here)

  Same bag of issues this branch came out of — queued for separate PRs:

  - **`enrichAllConversationsWithUnreadCounts` fires per-batch** in `ConversationStream.kt:531-539` (fire-and-forget). Desktop logs show ~500 ms × 5 batches
   per reaction ≈ 2.5 s of background CPU contention. Should fire only on `Stopped`, with optional targeted `ChatReadCount.selectUnreadCountForConversation`
   per touched conversation for mid-sync freshness. *Conservative fix recommended; the larger in-memory ReadCount rewrite is explicitly NOT recommended — it
   introduces a parallel cache that has to stay coherent with the DB across mark-as-read, peer-mark-as-read sync, deletion, leave, edit, etc.*
  - **Two `queryBatchAsync(noOfItems = 10_000)` calls** in `ConversationService.kt:1677` and `:1790` (pre/post delete audits). Look debug-only; worth gating
   behind a flag if so.
  - **`DriveMainIndexWrapper` has no `selectHomebaseFileById`** (by drive fileId). Has by-uniqueId, by-uniqueIds, by-globalTransitId. Not needed by this PR,
   but a real gap for future code that needs fileId-keyed lookup.
  - **`OptimisticWriter.reactionToggleLocks` has no eviction.** Bounded by distinct messages toggled per session — small in practice; worth an LRU cap if a
  long-lived client ever shows memory pressure.
  - **Cap guard reads `_messagesUiState.value.messages`** (the visible-list cache). If a tapped message isn't in the cache (rare — scrolled out / evicted),
  the guard is skipped and we fall back on the server rejection. Acceptable today; if logs ever show it firing in normal use, we'll plumb a DB read.
  - **Watch `homebase.log` for new `WARN queryBatchAsync called with a single uniqueId` lines** post-merge. This PR adds the diagnostic WARN; no chat
  operation should fire it. If one does, the stack snippet identifies the offender — likely the same `selectHomebaseFileByUnique` rewrite.

PR #476, opened 2026-05-09
perf(chat): gate unread-count enrich on dirty bit, drop per-batch fan-out

Seifert69:
The reaction PR's `(ConvListPerf)` instrumentation surfaced a steady 12+ calls of `enrichAllConversationsWithUnreadCounts=~500ms mirrored=0 changedRows=0` per chat session — wasted background work that contended with the UI dispatcher and made reactions feel janky.

Cause: `processConversationBatchIncrementally` fired the full enrich on every conversation-file batch (`ConversationStream.kt:530-539`), including drive-sync rounds that deliver many batches back-to-back and conversation files whose only change was metadata (name, participants, lastMessage timestamp). The actual job of that enrich is to mirror peer-device mark-as-read echoes — `localAppData.lastReadTime` into `ChatReadCount.lastReadTime` — which only matters when a conversation file's `lastRead` actually advanced.

Fix: a per-stream `dirtyUnreadIds: Set<Uuid>` accessed via three helpers (`markUnreadDirty`, `hasDirtyUnread`, `clearDirtyUnread`). A conversation is marked dirty exactly when a conversation file arrives and its `lastRead` is later than the prior in-memory entry, or when a brand-new conversation appears with no unread baseline.

Two checkpoints inspect the set:
- `DriveEvent.Stopped` (end of a sync round): gate flipped from `event.totalCount > 0` to `hasDirtyUnread()`. Mid-sync batches no longer fire enrichAll at all — the round's one Stopped picks up everything dirtied in flight.
- `BatchReceived(source = WebSocket)` (no Started/Stopped envelope — this batch is the whole event): runs enrichAll in-place if the batch dirtied anything.

`enrichAllConversationsWithUnreadCounts` now takes a `trigger: String` ("ColdLoad", "Stopped", "WebSocketBatch") that surfaces in the ConvListPerf log so we can attribute frequency without a stack walk. It clears the dirty set at the top — anything dirtied while it's in flight persists for the next checkpoint, preferring a small redundancy to a lost-wakeup race.

Trade-off this commit consciously accepts: today's `Stopped(totalCount > 0)` doubled as periodic reconciliation of the in-memory `++ unreadCount` at `:439-442` against the SQL formula in `ChatReadCount.sq`. After this PR, message-only sync rounds don't trigger enrichAll, so the in-memory increment is the source of truth between peer-mark-as-read events. Cold-load enrich still re-derives every count on next app start. If drift becomes observable in logs, the simple recovery is to OR the Stopped gate with `messageFilesInRound > 0` (separate counter, ~5 lines).

Future work this scaffold cleanly enables (none of it in this PR): debouncing rapid checkpoint hits, swapping in
`selectUnreadCountForConversation` when only one id is dirty, and marking dirty on suspect message arrivals if reconciliation needs to return.

Existing `homebase-chat:jvmTest` suite passes. No new tests this round (no existing ConversationStream test fixture; full integration fixture is ~400 lines of stubs for ContactService / DriveContactService / ConnectionService / image / share-cache deps — deferred). Manual verification path documented at the top of the diff: cold-boot for one ColdLoad line, idle for absence of mid-sync bursts, peer mark-as-read for a single Stopped or WebSocketBatch line.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageActionService.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageActionService.kt
@@ -275,7 +275,10 @@ class ChatMessageActionService(
     ) {
         val msg = messageLookup.getMessage(messageId) ?: return
         val conversation = conversationService.getConversation(msg.conversationId) ?: return
-        val fileId = requireFileId(messageId)
+        // msg.fileId is already populated by messageLookup.getMessage above —
+        // an extra requireFileId(messageId) here would issue a second
+        // selectHomebaseFileByUnique for the exact same row.
+        val fileId = msg.fileId
 
         // Soft-delete only for now — propagates to other clients via sync.
         // Hard-delete will be added as a second phase (user invokes delete again on soft-deleted msg).
@@ -327,6 +330,13 @@ class ChatMessageActionService(
     }
 
     suspend fun requireFileId(messageId: Uuid): Uuid {
+        // In-memory tier first: if the message's conversation is currently
+        // open and loaded, the MessageUiModel already carries fileId — no
+        // DB round-trip needed. Hit rate is effectively 100% for the
+        // remaining caller (getReactions on a visible message).
+        messageLookup.findCachedFileId(messageId)?.let { return it }
+
+        // Cache miss → indexed primary-key lookup.
         val credentials = credentialsManager.requireActiveCredentials()
         val file = dbm.driveMainIndex.selectHomebaseFileByUnique(
             credentials.getIdentityId(), chatDrive, messageId
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageStream.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMessageStream.kt
@@ -176,6 +176,28 @@ class ChatMessageStream(
         )
     }
 
+    /** Walk the open conversations' in-memory message lists for a model whose
+     *  `id == messageId` and return its `fileId`. The hit rate is effectively
+     *  100% for messageId-keyed actions invoked from a visible message
+     *  (toggleReaction's optimistic path is fileId-keyed and bypasses this
+     *  entirely; the remaining call sites — getReactions, deleteMessage's
+     *  fall-through — operate on a message the user just interacted with).
+     *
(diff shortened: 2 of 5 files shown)


Seifert69:
Manually tested on Desktop (Android Studio run) — both fixes behaving correctly.

  ### Cold boot
  Exactly one `enrichAllConversationsWithUnreadCounts=616ms trigger=ColdLoad mirrored=0 changedRows=7 totalRows=80`. 7 conversations had their unread state
  recovered from `ChatReadCount` (totals 1 + 8 + 45 + 6 + 1 + 178 + 2 = 241 unread messages). No mid-startup enrichAll bursts.

  ### During ~2 minutes of interaction (open conversations, mark-as-read, send messages, add a heart emoji)

  10 chat-drive sync rounds via WebSocket file-notifications (`syncDrive(chatDrive)`, not `syncAll()`). Of those:
  - **4 rounds** fired `enrichAllConversationsWithUnreadCounts trigger=Stopped` — each one a conversation file echo where `lastRead` actually advanced (peer
   mark-as-read or self-mark-as-read echo)
  - **5 rounds** correctly **skipped** enrichAll: 3 message-only batches and 2 conversation-file batches whose `lastRead` was unchanged (metadata-only
  echoes)
  - 1 reaction round (the heart emoji): message-only echo from server → no enrichAll

  Pre-fix code would have fired enrichAll on all ~9 of those rounds (under the old `totalCount > 0` gate). Saved roughly 5 × ~500 ms of background enrichAll
   work in this short session.

  ### Subjectively
  Heart emoji was visibly instant — no more perceptible ~400 ms delay. The combination delivers the win: the cap-reactions PR fixed the *direct*
  optimistic-write path (Mutex + `localReactions` write + `requireFileId` rewrite); the dirty-bit work eliminated the *secondary* contention from per-batch
  enrichAll bursts that was bleeding into the UI dispatcher.

  ### In-memory accounting
  - `unread++` correctly bumped count on incoming peer messages (no DB call).
  - `applyLocalAdvance` (single-row `selectUnreadCountForConversation`) correctly recomputed unread for locally-marked-as-read conversations.
  - `mirrored=0` across all enrichAll calls — `ChatReadCount.lastReadTime` and the conversation files agreed throughout.

  ### Dedupe path
  Silent in this trace (no `deleteMessage` or `getReactions` invoked). The cherry-picked dedupe commit compiles and tests pass; the cache-first contract
  just hasn't been exercised live yet.

PR #477, opened 2026-05-09
perf(ws): pure-push WS drive file events; retire per-event syncDrive

Seifert69:
Today every WebSocket file notification triggers
OdinWebSocketClient.handleFileEvent → driveSyncManager.syncDrive(driveId)
which fires an HTTP QueryBatch round-trip just to fetch a file the server already shipped on the WS payload (the encrypted header rides along on ClientDriveNotification.header). In a typical 2-minute chat session that's ~10 chat-drive sync rounds, each ~100-500ms.

The dirty-bit-unread-enrichment PR made each Stopped cheap, so the user-visible perf urgency is gone. What's left is bandwidth and architectural cleanliness. The dead handleFileEvent_manual showed the shape: decrypt header, upsert, emit BatchReceived(source=WebSocket). This commit wires that shape up properly — for the chat drive only — with batching and serialization mirroring DriveSync.

Scope: chat drive only. Other drives keep syncDrive (their cadences aren't part of this PR).

NEW: ChatDriveWebSocketUpsertWorker
  Mirrors DriveSync's job-serialization pattern as closely as
  possible (per user's explicit request). Single-flight Mutex.tryLock,
  atomic killroy bit, job var, finally-unlock-then-recurse-if-killroy.
  Adds a pendingFiles queue so bursts of WS notifications coalesce
  into one baseUpsertEntryZapZap call (one SQLite transaction
  instead of N) and one BatchReceived event.

  No sync cursor — the WS path doesn't manage one. The reconnect
  catch-up below is the recovery mechanism.

MODIFIED: OdinWebSocketClient
  - handleFileEvent now branches: chat drive + non-null header → decrypt + chatWsUpsertWorker.submit(file). Anything else (other drives, null header on chat drive, decrypt failures) falls through to syncDrive(driveId), logged at INFO so we can monitor the rate. A steady stream of chat-drive fallback logs would mean the WS payload is missing headers somewhere we didn't expect.
  - handleReactionEvent and handleAllReactionsDeletedEvent gain INFO logs at their syncDrive call sites too (per user request: visible accounting of every WS-triggered syncDrive).
  - onHandshakeSuccess adds a fire-and-forget syncDrive(SystemDriveConstants.chatDrive.alias) right after the ConnectionOnline emit. One sync per (re)connect — replaces the per-event syncDrive spam with a targeted catch-up backstop. The dirty-bit gating downstream keeps it cheap when the round brings nothing relevant.
  - disconnect() cancels the worker.
  - Dead handleFileEvent_manual removed (its work moved into the worker).

DIAGNOSTIC (Step 0 of the plan, REMOVED before merge):
  A temporary Logger.d at the top of handleNotification prints
  notification.notificationType for every WS message. Lets the user
  observe — on Desktop, in homebase.log — which notification types
  fire for which actions. Critical question this answers: does a
  reaction toggle also produce a fileModified (carrying the updated
  reactionPreview-baked header), or only a reactionContentAdded? If
  both, the reaction handlers' syncDrive calls become redundant for
  the chat drive (a follow-up to drop them). If only the latter,
  reactions remain a separate follow-up PR.

  Tagged "WSDiag" so it's easy to grep, and marked TODO: remove
  before merge.

Existing chat:jvmTest and api:jvmTest suites pass (54 tasks executed, BUILD SUCCESSFUL). No new tests this round; OdinWebSocketClient has no existing test fixture and the manual Desktop log is the proving ground (mirroring how dirty-bit-unread-enrichment shipped).

Manual verification path is in the plan; the headline checks are:
  1. Send/receive a chat message → see "ChatDriveWS: batch upsert rows=N took=Xms" in homebase.log instead of "Synchronizing drive 9ff813af-...".
  2. Reconnect after a forced WS drop → exactly one "Synchronizing drive 9ff813af-..." (the new catch-up) right after handshake.
  3. Watch for "WSFileEvent: syncDrive(...)" — chat-drive entries here are the fallback signal we want to know about.

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/websockets/OdinWebSocketClient.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/websockets/OdinWebSocketClient.kt
@@ -3,7 +3,6 @@ package id.homebase.api.client.websockets
 import co.touchlab.kermit.Logger
 import id.homebase.api.client.SharedSecretEncryptedPayload
 import id.homebase.api.client.auth.CredentialsManager
-import id.homebase.api.client.drives.SystemDriveConstants
 import id.homebase.api.client.drives.TargetDrive
 import id.homebase.api.client.eventbus.BackendEvent
 import id.homebase.api.client.eventbus.EventBus
@@ -11,8 +10,8 @@ import id.homebase.api.common.SecureByteArray
 import id.homebase.api.crypto.AesCbc
 import id.homebase.api.crypto.ByteArrayUtil
 import id.homebase.api.serialization.OdinSystemSerializer
-import id.homebase.api.sync.ChatDriveWebSocketUpsertWorker
 import id.homebase.api.sync.DriveSyncManager
+import id.homebase.api.sync.DriveWebSocketUpsertWorker
 import id.homebase.api.sync.database.DatabaseManager
 import id.homebase.api.sync.database.MainIndexMetaHelpers
 import id.homebase.api.toBase64
@@ -66,11 +65,18 @@ class OdinWebSocketClient(
 
     private var fileHeaderProcessor = MainIndexMetaHelpers.HomebaseFileProcessor(databaseManager)
 
-    // Lazily initialised on the first [connectOnce] call once we have
-    // credentials. Receives WS-pushed file headers for the chat drive
-    // and writes them straight to DriveMainIndex without a Drive.sync()
-    // round-trip. Other drives stay on [driveSyncManager.syncDrive].
-    private var chatWsUpsertWorker: ChatDriveWebSocketUpsertWorker? = null
+    // Per-drive WS upsert workers. Lazily created on the first file
+    // event for a drive (see [getOrCreateWorker]). Each worker
+    // batches incoming files into one DB transaction and emits a
+    // single [BackendEvent.DriveEvent.BatchReceived] per drain;
+    // shape mirrors [DriveSync].
+    //
+    // Cleared in [disconnect]. Mount/unmount of drives is handled
+    // automatically because [AuthConnectionCoordinator.reconnectWebSocket]
+    // destroys the old [OdinWebSocketClient] and creates a fresh one
+    // with an empty map.
+    private val wsUpsertWorkers = mutableMapOf<Uuid, DriveWebSocketUpsertWorker>()
+    private val wsUpsertWorkersMutex = Mutex()
 
     private lateinit var sharedSecret: ByteArray
 
@@ -202,18 +208,9 @@ class OdinWebSocketClient(
         val identity = creds.domain
         sharedSecret = creds.sharedSecret.unsafeBytes
 
-        // Lazily build the chat-drive WS upsert worker once we know the
-        // identity. Idempotent across reconnects — the same worker drains
-        // future bursts. Cancelled in [disconnect].
-        if (chatWsUpsertWorker == null) {
-            chatWsUpsertWorker = ChatDriveWebSocketUpsertWorker(
-                identityId = creds.getIdentityId(),
-                driveId = SystemDriveConstants.chatDrive.alias,
-                databaseManager = databaseManager,
-                eventBus = eventBus,
-                scope = scope,
-            )
-        }
+        // Per-drive WS upsert workers materialise lazily in
+        // [getOrCreateWorker] when a file event for that drive
+        // arrives — no need to predeclare which drives we expect.
 
         _connectionState.value = WebSocketState.Connecting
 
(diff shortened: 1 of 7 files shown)


Seifert69:
## Review — `websocket-pure-push-chat-drive`

  5 commits, +1181 / -74 across 6 files. End-state combines three orthogonal wins built on top of each other:

  1. **Pure-push path for WS file events** (across all drives, not just chat).
  2. **In-memory unread bump actually persists** (pre-existing bug, exposed by the dirty-bit gating).
  3. **Reaction echoes don't multiply unread**.

  Plus a side rename and reaction-handler simplification.

  ### What landed

  | File | Lines | Role |
  | --- | --- | --- |
  | `OdinWebSocketClient.kt` | +101 / −75 | `wsUpsertWorkers` map (lazy per-drive) replaces the `syncDrive` path; reaction handlers become no-ops; reconnect
   catch-up dropped (covered by `AuthConnectionCoordinator.syncAll()`). |
  | `DriveWebSocketUpsertWorker.kt` | new (175) | Mirrors `DriveSync.kt`'s job-serialisation shape (single-flight Mutex + atomic killroy +
  `finally`-recurse) with a pending-files queue. Batches WS bursts into one `baseUpsertEntryZapZap` call per drain, emits one `BatchReceived(source =
  WebSocket)`. |
  | `DriveWebSocketUpsertWorkerTest.kt` | new (390) | 6 tests against real in-memory SQLite + real `EventBus`: single submit, burst batching, WebSocket
  source pin, drive-agnostic, sequential drains, cancel bound. Uses `runBlocking` (real dispatchers) — `runTest`'s `backgroundScope` was unreliable for the
  worker's `scope.launch + Mutex` pattern. |
  | `ConversationMessageBump.kt` | new (91) | Pure top-level `applyIncomingMessageBump(items, …): List<…>?`. All in-memory bump logic. Returns null on
  no-op. Strict-greater `userDate` guard rejects reaction-driven re-emits. |
  | `ConversationStream.kt` | +90 net | `updateConversationFromNewMessage` rewritten to delegate to the pure helper instead of `updateConversation` (which
  silently dropped `unreadCount`). Plus a HACK `associateBy { m.id }` dedupe in `processMessageBatchIncrementally` to defend against server-side WS fan-out
  (3 notifications per logical write today). |
  | `ConversationMessageBumpTest.kt` | new (333) | 10 cases including the regression guards: `sequentialPeerMessages_accumulateUnread` (the original
  dropped-`unreadCount` bug), `reactionEcho_sameUserDate_returnsNull_noUnreadBump` and `peerMessageThenReactionEchoes_doesNotMultiplyUnread` (the `<=` guard
   fix). |

  ### What I'd highlight as good

  - **Worker mirrors `DriveSync` deliberately.** Same Mutex + killroy + finally-recurse-then-relaunch pattern. Anyone who's debugged DriveSync can read this
   in 30 seconds. The choice to copy rather than abstract was the right call.
  - **Defensive fallbacks throughout.** Header-null → `syncDrive`; decrypt fails → `syncDrive`; closed/no-creds → null worker → `syncDrive`. Every fallback
  logs at INFO so production rate is observable.
  - **Mount/unmount works automatically** because `AuthConnectionCoordinator.reconnectWebSocket` destroys the old WS client (workers die with it) and
  creates a fresh one. No special unmount hook needed; the comment explaining this is in the worker class KDoc.
  - **HACK markers are loud.** Both the `processMessageBatchIncrementally` dedupe and the worker's duplicate-detection log are wrapped in big `╔══╗` ASCII
  boxes with `TODO(server)` so they can't be missed at review or grepped past.
  - **Pure-function extraction for the bump logic** keeps the regression test cheap (no DB, no Koin) and the regression name
  (`sequentialPeerMessages_accumulateUnread`) describes exactly the bug it's guarding.
  - **Branch tells a story.** Each commit is a discrete shippable thing: chat-only pure-push → reaction handler no-op → unread bump persistence →
  generalisation → reaction-echo guard. Clean to bisect if anything regresses.

  ### What's worth flagging at merge time

  - **Reaction handler no-op assumes `statisticsChanged` fires alongside `reactionContentAdded` for every drive.** Verified empirically on the chat drive
  only; we accepted "drives are protocol-identical" as the basis for all-drive generalisation. If a future drive type breaks this assumption,
  reaction-preview updates would silently stop on that drive. Cheap follow-up if it ever bites: re-instate `syncDrive` in `handleReactionEvent` for affected
   drives. The receiver-side counterpart (older clients seeing reaction lag) is a release-rollout artefact, not a code bug.
  - **Strict-greater `userDate` guard** in `applyIncomingMessageBump` rejects any message whose userDate equals the conversation's `latestMessageTimestamp`.
   Trade-off documented in the commit and in code: two genuinely-new messages with identical millisecond stamps would only bump once. The reaction-echo case
   (identical userDate) is far more common.
  - **Two HACK dedupe blocks** (worker + ConversationStream) both labelled `TODO(server): stop fan-out-per-write`. They're defensive (idempotent upserts
  already collapse duplicates correctly; HACKs only suppress the redundant per-message side effects). When the server-side fix lands, both can be deleted.
  - **`OdinWebSocketClient` has no unit tests** — fixture cost was assessed as too high (HTTP/WS server stub). Manual Desktop verification on the dispatch
  wiring stands; the worker itself has thorough coverage.

  ### Performance impact

  A 2-minute interactive session that previously fired ~10 `syncDrive` round-trips for the chat drive (one per WS file notification) now fires ~1 (cold-boot
   `syncAll`). Other drives benefit identically. WS bursts now collapse into single DB transactions via the worker's queue. Combined with the prior
  `dirty-bit-unread-enrichment` work on `main`, message- and reaction-related background work is now nearly free.

  ### Verdict

  Shippable. The HACK dedupes are the only items in the diff that explicitly call for a follow-up — both via TODO(server) which makes the trigger condition
  (server-side fan-out fix) explicit. Everything else is either a feature lock or a tested regression guard.

PR #513, opened 2026-05-12
fix(chat): stop leaking Event title in push notifications

Seifert69:
Event chat messages were rendering "Surprise party for Jane in Family Group" in the recipient's OS notification banner — exposing the event title to anyone glancing at the lock screen. The leak was the sender copying EventDescriptor.title into PushNotificationOptions. unEncryptedMessage, which the recipient's NotificationBodyFormer (and the iOS NotificationServiceExtension) renders verbatim.

The fix is a sentinel-prefixed wire format: senders emit "__hb_event__|<startUtcMs>" instead of the title, and the recipient formats a privacy-safe body from its own clock — "Event in 30 minutes in Family Group" / "Event on Thu, May 14". The title, description, location, and meeting URL never leave the sender's host.

Shared across runtimes
----------------------
A new leaf module `homebase-notifshared` holds the parser + formatter. Dependencies are stdlib + kotlinx-datetime only — no Compose, no Ktor, no DI, no logging — so the iOS NotificationServiceExtension can eventually link `HomebaseNotifKit.framework` without blowing its ~24 MB RAM budget. Android, Desktop, and (once linked) iOS all call the same Kotlin function via the framework — one implementation, one test suite, no Swift/Kotlin drift.

Sender
------
- MessageContent gains `notificationLabel` (defaults to `displayLabel`).
- Event overrides it to "__hb_event__|<startUtcMs>"; descriptor stays local for chat-list previews and search.
- ChatMessageSenderService.sendNewTypedMessage passes `notificationLabel` to the outgoing PushNotificationOptions; the existing group-suffix wrap is unchanged.

Receiver (Kotlin — Android + Desktop)
------------------------------------
- NotificationBodyFormer.form() checks the sentinel before the verbatim short-circuit and falls through cleanly for non-event payloads.
- New NotificationBodyFormerTest covers the sentinel path, group suffix, plain-text passthrough, and the missing-message default.

Receiver (Swift — iOS NSE)
--------------------------
- NotificationService.swift calls EventNotificationFormatterKt. tryFormatEventNotificationBodyNow under #if canImport(HomebaseNotifKit).
- A Swift-side fallback strips the sentinel and renders "Event" / "Event in <groupName>" when the framework isn't linked, so iOS users never see the raw token.
- iosApp's "Compile Kotlin Framework" Run Script now also builds :homebase-notifshared:embedAndSignAppleFrameworkForXcode.

Manual step (iOS, on a Mac)
---------------------------
The NotificationServiceExtension target still needs HomebaseNotifKit.framework added to "Link Binary With Libraries" in Xcode UI before the full relative-time format ("Event in 1 hour") takes effect on iOS. The Swift `#if canImport` guard keeps the file compiling either way; until linked, iOS uses the Swift fallback. The helper's KDoc-equivalent in NotificationService.swift documents the wiring.

Verification
------------
- ./gradlew :homebase-notifshared:jvmTest — 14 tests pass (parser round-trips, group name containing " in ", every formatter table branch from "starting now" to absolute past/future).
- ./gradlew :homebase-common:jvmTest — NotificationBodyFormerTest passes plus everything else green.
- ./gradlew :homebase-core:compileKotlinJvm :desktopApp:compileKotlinJvm — full Desktop dep chain compiles clean.
- Pre-existing flake in homebase-chat:jvmTest (LocalAttachmentContextStoreFileExistenceTest, 9 tests, Windows FS) fails identically on clean main with these changes stashed — unrelated.


Seifert69:
Why the iOS NSE has a Swift-side fallback for Event notifications

  Context: the 24 MB budget

  iOS Notification Service Extensions run in their own process, separate from the host app, and Apple caps that process at 24 MB of RAM (historically; newer
   iOS bumps this a little, but 24 MB is the floor you design against). If the NSE goes over, iOS kills it mid-execution and the user sees Apple's default
  fallback — title only, no body, no avatar, no Communication style. That's worse than any of the cases we're designing for, so the budget drives the whole
  structure of this fix on iOS.

  Why a new tiny module instead of reusing ComposeApp

  The natural place to put the formatter would be homebase-common, which is already shipped as part of ComposeApp.framework for the main iOS app. But
  ComposeApp transitively links Compose Multiplatform, Ktor, SQLDelight, Coil, Koin, kmpnotifier — loading that graph into a 24 MB process is a non-starter.
   So the formatter went into a new leaf module, homebase-notifshared, whose only deps are stdlib + kotlinx-datetime. Its HomebaseNotifKit.framework should
  add only a couple of MB to the NSE's working set, leaving headroom for the NSE's own work (parsing the payload, fetching the sender avatar, building the
  INSendMessageIntent).

  Why linking is a manual Xcode step

  The Gradle side is done — the iosApp "Compile Kotlin Framework" Run Script now builds :homebase-notifshared:embedAndSignAppleFrameworkForXcode alongside
  ComposeApp. But adding HomebaseNotifKit.framework to the NSE target's "Link Binary With Libraries" requires Xcode-UI work, which I can't do reliably from
  this branch — hand-editing iosApp.xcodeproj/project.pbxproj for framework linkage is high-risk (Xcode treats a malformed pbxproj as "damaged"), and I
  can't validate the change without a Mac + Xcode in front of me.

  Why the Swift fallback exists

  NotificationService.swift calls EventNotificationFormatterKt.tryFormatEventNotificationBodyNow(raw:) under #if canImport(HomebaseNotifKit). The canImport
  guard makes the file compile in both states.

  Without a #else branch, iOS users would see the raw __hb_event__|1715716200000 in Family Group wire token in their notification banner until someone wires
   up the framework — strictly worse than the title leak we set out to fix. The fallbackEventBody helper parses the sentinel in pure Swift and returns
  "Event" / "Event in " without the relative-time computation, so the iOS experience degrades gracefully during the rollout window.

  Why I'd keep the fallback even after linkage

  Once the framework is linked, canImport succeeds, the #if branch lights up, and iOS gets the same "Event in 1 hour" / "Event on Thu, May 14" output
  Android and Desktop produce. The fallback becomes dead code on the happy path — but I'd leave it in. It's the floor when:

  - A future change to homebase-notifshared accidentally adds a heavy transitive dep and pushes the NSE over 24 MB → process gets killed, the framework call
   never runs.
  - The framework fails to load for any other runtime reason in the NSE.

  In those cases, the pure-Swift fallback still produces a usable string instead of leaving the user with Apple's title-only default. ~20 lines of Swift is
  cheap insurance against violating the budget that motivated this whole module split.

  TL;DR — manual step

  In Xcode, select the NotificationServiceExtension target → General → Frameworks and Libraries → "+" → add HomebaseNotifKit.framework (it'll be in
  BUILT_PRODUCTS_DIR after the next iosApp build). Once that's done, the #if canImport branch starts running and the fallback only matters as
  defense-in-depth.

PR #548, opened 2026-05-17
perf(video): Android-native compress, thumb, rotation, EXIF strip

Seifert69:
Moves the Android video pipeline off FFmpegKit wherever there's a viable platform-native API. Keeps ffmpeg only for HLS segmentation (Signal-Android doesn't do HLS and there is no clean Android equivalent).

Compress (the main thing)
  - Vendor Signal-Android's StreamingTranscoder plus the MediaCodec / MediaExtractor / MediaFormat / MediaMuxer machinery it depends on (~37 files under homebase-api/src/androidMain/.../video/transcoder/). Original license headers preserved verbatim (Apache-2.0 + AGPL-3.0). Local divergences from upstream: two Log.tag(Class) calls inlined as string literals, the second StreamingTranscoder constructor promoted public for the HIGH custom-bitrate path, and the containsLocation re-encode trigger dropped (our Mp4Sanitizer is a no-op stub — re-encoding without strip was pure waste).
  - Replace FFmpegUtils.android.kt:compressVideo body with a wrapper that calls StreamingTranscoder. Hardware -> software codec fallback is handled inside Signal's MediaConverter (REGULAR_CODECS -> ALL_CODECS).
  - JVM/iOS/web actuals: signature accepts the new quality param but bodies unchanged (TODO to map).

Native thumb + rotation
  - grabThumbnail -> MediaMetadataRetriever (via VideoThumbnailExtractor. extractPosterFrame, already platform-native). Side benefit: fixes a SIGABRT in ffmpeg-kit 7.x's sch_add_demux_stream on HLS-remuxed inputs.
  - getRotationFromFile -> MediaMetadataRetriever.METADATA_KEY_VIDEO_ROTATION.

EXIF / location strip
  - New Mp4LocationStripper walks moov/udta and moov/trak/udta, removes AppleGPS-style atoms (©xyz, loci, gpsa) via mp4parser (already a dep for the faststart post-processor). No re-encode -- just rewrite the moov atom and copy the mdat through. Called from compressVideo's short-circuit branch only; the transcoded path doesn't need it because MediaMuxer doesn't propagate user-data atoms by default.

VideoQuality
  - New enum LOW / STANDARD / HIGH threaded through FFmpegUtils.compressVideo and VideoPayloadProcessor.process, default STANDARD. Ready for a future userPreferences.videoQuality wiring; no caller changes required today.

Tests (all @Ignore'd until emulator-on-CI infra lands)
  - androidDeviceTest source set wired via withDeviceTest {}.
  - CompressVideoAndroidInstrumentedTest covers per-quality output and progress-callback continuity.
  - Mp4LocationStripperAndroidTest constructs in-memory MP4s with/without ©xyz via mp4parser and round-trips them through the stripper.

Also bundled (shared the same files)
  - FFmpeg version display in Settings -> Help -> About (FFmpegUtils.kt getFfmpegVersion, FFmpegKitBridge.getFfmpegVersionBanner with Swift impl, HelpUiState/ViewModel/Screen wiring, ffmpeg banner parser test).
  - JVM-side FFmpegPipelineCoverageJvmTest exercises every ffmpeg-touching FFmpegUtils method end-to-end through one pipeline against the bundled desktop binary. Catches structural regressions on future JVM ffmpeg upgrades.

Out of scope
  - ffmpeg-kit Android v7 upgrade reverted earlier in this session (separate concern: SIGABRT on HLS-remuxed input + missing 16KB-page alignment).


Seifert69:
Too little speed gain to justify. But it's working!

PR #656, opened 2026-06-04
build(compose): upgrade to Compose Multiplatform 1.11.1 (Skia m144) — fix blank text

Seifert69:
## Why

**Move 1** of the iOS + web *blank-text* investigation. Symptom: UI **frames render but all glyphs vanish** — the signature of a Skia **GPU glyph-atlas** failure (text is sampled from the atlas; solid frames/borders bypass it, so they survive). It only hits the two Skia-GPU text backends — **iOS (Metal)** and **web/wasm (WebGL)** — while Android and Desktop are fine. The app uses only `FontFamily.Default/SansSerif/Monospace` with **zero bundled fonts**, so this is not a custom-font problem.

The iOS *background-GPU* variant — JetBrains [CMP-9488](https://youtrack.jetbrains.com/issue/CMP-9488) ("dirty font cache when entering background", an inverted `MetalRedrawer.isActive` guard that submitted GPU work from the background) — is **already fixed in Compose 1.10.1**, which we have on `main` (1.10.3). That's why the existing `setNeedsDisplay` nudge (#506/#521) only "helps a little." The **remaining shared failure is the cold-start / first-GPU-frame atlas**.

Compose **1.11** bumps Skia **m138 → m144**, which carries upstream glyph-atlas/GPU fixes. We try the framework bump *before* writing custom per-platform redraw machinery.

## What changed

- `composeMultiplatform` **1.10.3 → 1.11.1** — resolved `skiko` is **0.144.6** (= Skia m144) ✅
- Aligned to JetBrains' **1.11.1 compatibility table** (no stable material3 exists in the 1.11 line; the alpha/beta tags *are* the recommended set):
  - `jetbrains-compose-material3` 1.9.0 → **1.11.0-alpha07**
  - `jetbrains-compose-material3-adaptive` 1.2.0 → **1.3.0-alpha07**
  - `androidx-lifecycle` 2.10.0 → **2.11.0-beta01**
  - navigation stays **2.9.2**; Kotlin stays **2.3.21** (satisfies 1.11's ≥2.3.20 for JS/Wasm) — no Kotlin bump
- **iOS:** `ComposeUIViewController(configure = { parallelRendering = false })` — pins single-threaded rendering so this PR isolates the **Skia bump** from 1.11's new **default-on separate render thread**, which adds atlas upload/present synchronization on the exact surface we're debugging. Flip to `true` only as a deliberate, separately-tested experiment.

## Verification

- Locally compile-verified: `:webApp:compileKotlinWasmJs` + JVM (`:homebase-core:compileKotlinJvm`) **green**.
- iOS not compiled here (needs macOS); the `parallelRendering` API was confirmed against the CMP 1.11 source/docs.
- Landmines checked: only `iosArm64`/`iosSimulatorArm64` targets (dodges 1.11's Apple-x86_64 drop); no `androidx.collection` usage; no raw `org.jetbrains.skia.Shader` (the one 1.11 ABI break doesn't touch us).

### Notes / watch-items
- Pre-existing benign warning `Skiko dependencies' versions are incompatible`: `coil3 3.4.0` / `zoomimage 1.4.0` request old skiko 0.9.x; the Compose plugin force-aligns up to 0.144.6 (same override that already happened on 1.10.3). It's image-decode skiko, not text — **glance that images still load** during smoke test.
- Non-blocking deprecations surfaced by 1.11: `currentWindowAdaptiveInfo(supportLargeAndXLargeWidth=…)` → V2, and `BackHandler` → `NavigationEventHandler`. Left as follow-ups.

## How to test
The decisive test is a **production deploy to the demo env** (the web bug was production-only — it never reproduced on localhost), then check the login screen for text. If text renders → Skia m144 fixed the cold-start atlas; confirm on an iOS device and then delete the now-redundant nudge scaffolding. If still blank → points at glyph/font *availability* on the wasm deploy (next investigation step), not a redraw nudge.


Seifert69:
Closing: the Compose 1.11 upgrade this branch tested was reverted (it was the wrong hypothesis — see PR for details, and it broke CI via material3 1.11.0-alpha07 pinning compose-ui 1.10.2). The web blank-text root cause turned out to be server-side (Compose .cvr string table served as index.html by Kestrel), fixed in odin-core PR #1547. The investigation record (BLANK_TEXT_INVESTIGATION.md) is being carried forward onto the iOS-diagnostics branch. iOS blank-text is a separate issue, now being addressed with GPU/glyph diagnostics logging.

PR #659, opened 2026-06-05
Share into Homebase now switches to the target conversation

2002Bishwajeet:
## Root cause

After sharing content into a conversation, `ShareReceiverActivity` re-opens `MainActivity` on a `homebase-fchat://conversation/{id}` deep link to switch to that chat. Three send paths built that intent inline:

- `sendToMultipleConversations` (text-only, multi-select)
- `sendSharedContent` (text-only, single / direct-share target)
- `sendEditedFiles` (file attachments via the editor)

Each set `data` + `FLAG_ACTIVITY_CLEAR_TOP | FLAG_ACTIVITY_SINGLE_TOP`, but **none** set `ShareShortcutPublisher.EXTRA_FROM_SHARE_SHORTCUT`.

`MainActivity.handleIntent()` uses that extra to choose the deep-link source:

```kotlin
val source = if (fromShareShortcut)
    OpenConversation.Source.ShareIntent
else
    OpenConversation.Source.NotificationTap
```

Without the extra, the share deep link was classified as `Source.NotificationTap`, which resolves via a `PendingNotificationTap` that requires a `messageId` a share never has — so the navigation dead-ended and the conversation never opened. `Source.ShareIntent` is the source `AppNavHost` routes through `selectConversationOnChatList` to actually open the conversation.

## The fix

Extract a single private helper in `ShareReceiverActivity`:

```kotlin
internal fun openConversationIntent(conversationId: Uuid): Intent
```

It builds the URI + the two flags + `putExtra(EXTRA_FROM_SHARE_SHORTCUT, true)` once, and all three send paths now call it. This reuses the **existing** `Source.ShareIntent` routing that the launcher share shortcut already relies on — `MainActivity` / `NotificationService` / `AppNavHost` are untouched.

## Cross-platform notes

Android-only. iOS already does the equivalent: `AppViewModel.handleShareIntent` emits `Source.ShareIntent`. Desktop/Web have no share-receiver entry point.

## Verified

- `./gradlew :androidApp:testDebugUnitTest --tests "id.homebase.feed.share.ShareReceiverActivityIntentTest"` — 3 tests pass (compiles `:androidApp:compileDebugKotlin` as part of the run, BUILD SUCCESSFUL).
- New Robolectric test `ShareReceiverActivityIntentTest` asserts the helper's intent carries `EXTRA_FROM_SHARE_SHORTCUT == true`, targets `MainActivity` with `homebase-fchat://conversation/{id}` data, and keeps `CLEAR_TOP`/`SINGLE_TOP`. It mirrors the existing `ShareShortcutPublisherTest` Robolectric setup and builds the activity via `Robolectric.buildActivity(...).get()` so no Koin graph is needed (the helper only uses the Activity `Context`).

### DIFF
--- a/androidApp/src/main/kotlin/id/homebase/feed/share/ShareReceiverActivity.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/share/ShareReceiverActivity.kt
@@ -345,13 +345,7 @@ class ShareReceiverActivity : ComponentActivity(), KoinComponent {
                 Toast.makeText(this@ShareReceiverActivity, countLabel, Toast.LENGTH_SHORT).show()
 
                 // Open the first conversation in the main app
-                val firstId = conversationIds.first()
-                val mainIntent =
-                    Intent(this@ShareReceiverActivity, MainActivity::class.java).apply {
-                        data = "homebase-fchat://conversation/${firstId}".toUri()
-                        flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
-                    }
-                startActivity(mainIntent)
+                startActivity(openConversationIntent(conversationIds.first()))
                 finish()
             } catch (e: Exception) {
                 Logger.e(tag = "ShareReceiver") { "Failed to send: ${e.message}" }
@@ -392,12 +386,7 @@ class ShareReceiverActivity : ComponentActivity(), KoinComponent {
                 Toast.makeText(this@ShareReceiverActivity, getString(R.string.share_sent), Toast.LENGTH_SHORT).show()
 
                 // Open conversation in main app
-                val mainIntent =
-                    Intent(this@ShareReceiverActivity, MainActivity::class.java).apply {
-                        data = "homebase-fchat://conversation/${conversationId}".toUri()
-                        flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
-                    }
-                startActivity(mainIntent)
+                startActivity(openConversationIntent(conversationId))
                 finish()
             } catch (e: Exception) {
                 Logger.e(tag = "ShareReceiver") { "Failed to send: ${e.message}" }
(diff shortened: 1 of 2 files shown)


2002Bishwajeet:

Verified the fix addresses the real root cause, not a symptom.

**Root cause / correctness — confirmed**
- `MainActivity.handleIntent()` (MainActivity.kt:146-153) branches on `EXTRA_FROM_SHARE_SHORTCUT`: present → `Source.ShareIntent`, absent → `Source.NotificationTap`. The three post-send re-open intents omitted it, so shares were misclassified as NotificationTap.
- `AppNavHost.kt:329-331` routes only `Source.ShareIntent` through `selectConversationOnChatList(id)`; NotificationTap just pops to ChatList and defers to `PendingNotificationTap`, which needs a messageId a share never has — hence the dead-end. Adding the extra flips the classification to the already-tested ShareIntent path. No MainActivity/NavHost change needed.
- All 3 send paths (text-only single :348, text-only multi :389, edited files :564) now funnel through the single `openConversationIntent` helper (:447). Grep confirms no other `MainActivity::class.java` intent construction remains — none missed.
- Helper mirrors the flags/data/extra already used by `ShareShortcutPublisher.buildShortcut` (the launcher-tap path), keeping the two in sync. iOS already emits ShareIntent (`AppViewModel.handleShareIntent`); Desktop/Web have no share entry. Android-only fix brings parity.

**Checklist**
- Konsist: N/A — pure Android Intent plumbing, no Compose. The only `Text(` hits are `Toast.makeText` on unchanged context lines using existing string resources (`countLabel`, `R.string.share_sent`).
- truncateToCodePoints / take / substring: none introduced. The `$conversationId` interpolation is a UUID (ASCII-safe).
- New strings: none → values-da parity N/A.
- M3 / padding / AutoMirrored / collectAsStateWithLifecycle: N/A (no UI).
- Cross-platform: change is confined to `androidApp`; no commonMain touched.
- Wire-format / app-data / sender: untouched.
- Test: `ShareReceiverActivityIntentTest` added (Robolectric, mirroring existing `ShareShortcutPublisherTest` in the same package; `libs.robolectric` already wired). Asserts the extra, the MainActivity component, the deep link, and CLEAR_TOP/SINGLE_TOP. The `buildActivity(...).get()` (no onCreate) approach is valid: all injected deps are lazy `by inject()`, and `openConversationIntent` touches none of them — only Context + Uuid. `@OptIn(ExperimentalUuidApi)` present on the class and the test.

**Non-blocking nits**
- The helper's no-args overload always sets `putExtra(..., true)`; harmless but the helper name doesn't hint at the extra — the thorough KDoc compensates.
- Test could additionally assert there is no stray messageId extra, but the three positive assertions already pin the behavior.

No blocking issues.

PR #660, opened 2026-06-05
Hide archived conversations from the share picker

2002Bishwajeet:
## Root cause

Three share surfaces never filtered `ConversationState.Archived`, so chats the user had deliberately archived still appeared as share targets:

- the Android in-app share picker (`SharePickerScreen.kt`),
- the iOS share-extension conversation picker, and
- the Android Direct Share shortcuts.

The last two both read the share cache built in `ConversationStream.updateShareCache`. The main conversation list already excludes Archived; the share path did not.

## The fix

Two edits, each excluding **only** `Archived` (not the full main-list whitelist), so the change stays literal to the bug:

1. **`androidApp/.../share/SharePickerScreen.kt`** — filter `it.conversationState != ConversationState.Archived` in the `enrichedConversations` `derivedStateOf`, before enriching. recents/contacts/groups/search all derive from `enrichedConversations`, so this single filter covers the whole screen.

2. **`homebase-chat/.../convo/ConversationStream.kt` (`updateShareCache`)** — compute the non-archived list once, map it to `ShareableConversation`, and reuse the same filtered list for the avatar pre-cache loop so archived group avatars are not needlessly decrypted. This single place feeds both the iOS share-extension cache and the Android Direct Share shortcuts. The filter+map was extracted into a pure `internal fun buildShareableConversations(...)` so the cache contents are unit-testable without standing up the full stream.

`ConversationListViewModel` (main list) and the archived screen are untouched.

## What I verified

- `./gradlew :homebase-chat:compileKotlinJvm :homebase-chat:compileTestKotlinJvm` — BUILD SUCCESSFUL (the only warnings are pre-existing, in an unrelated `Dice1d4RoundTripTest.kt`).
- New JVM test `ShareCacheArchivedFilterTest` passes: an `Archived` conversation is absent from the share cache while an `Active` one is present, plus a defensive case asserting a non-Active, non-Archived state (`Left`) is **not** filtered (guards against accidentally widening the filter).

## Cross-platform notes

- Edit 2 lives in `commonMain`, so the single fix covers the iOS share-extension picker and the Android Direct Share shortcuts together; Edit 1 covers the Android in-app picker UI.
- Desktop/Web have no share-entry surface, so no platform-specific work is needed there. Android/iOS share entry points pick up the fix automatically.
- No wire-format, schema, or sender-cap changes.

### DIFF
--- a/androidApp/src/main/kotlin/id/homebase/feed/share/SharePickerScreen.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/share/SharePickerScreen.kt
@@ -50,6 +50,7 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import co.touchlab.kermit.Logger
 import id.homebase.api.client.auth.OwnerSessionRepository
 import id.homebase.api.util.truncateToCodePoints
+import id.homebase.chat.data.ConversationState
 import id.homebase.chat.services.convo.ConversationEnricher
 import id.homebase.chat.services.convo.ConversationStream
 import id.homebase.chat.services.convo.EnrichedConversationUiModel
@@ -94,12 +95,17 @@ fun SharePickerScreen(
     var selectedIds by remember { mutableStateOf(emptySet<Uuid>()) }
     val enricher = remember { ConversationEnricher() }
 
-    // Enrich conversations with contact names — same pattern as ConversationListViewModel
+    // Enrich conversations with contact names — same pattern as ConversationListViewModel.
+    // Archived conversations are excluded from the share picker so the user can't share
+    // into a chat they have deliberately tucked away. recents/contacts/groups/search all
+    // derive from this list, so this single filter covers the whole screen.
     val enrichedConversations by remember {
         derivedStateOf {
             val session = ownerSession ?: return@derivedStateOf emptyList()
             val contactMap = contactsState.associateBy { it.odinId }
-            conversationsData.items.map { enricher.enrich(it, contactMap, session) }
+            conversationsData.items
+                .filter { it.conversationState != ConversationState.Archived }
+                .map { enricher.enrich(it, contactMap, session) }
         }
     }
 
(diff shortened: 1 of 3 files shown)


2002Bishwajeet:

Reviewed against the Task D spec (`docs/superpowers/specs/2026-06-05-chat-kmp-7-task-batch-design.md` §D). Built and **ran** the new test — `./gradlew :homebase-chat:jvmTest` BUILD SUCCESSFUL.

**Root cause — fixed, not patched.** Spec names 3 unfiltered share surfaces; the PR closes all 3 from 2 edits:
- `SharePickerScreen.kt:98-104` — filter `!= Archived` in the `enrichedConversations` derive. recents/contacts/groups/search all derive from this list, so one filter covers the whole Android in-app picker. Verified each section reads `enrichedConversations`.
- `ConversationStream.updateShareCache` — computes `nonArchived` once and reuses it for the cache write (feeds iOS share extension + Android Direct Share shortcuts) **and** the avatar pre-cache loop. Good catch iterating `nonArchived` in the loop — archived group avatars are no longer needlessly decrypted/cached.

**Refactor.** `buildShareableConversations` extracted as a pure `internal` commonMain function; the mapping logic is moved verbatim (no behavioural change), enabling the unit test without standing up the stream. Lives in commonMain, uses only commonMain types (`ShareableConversation` from `id.homebase.core.share`) + stdlib — cross-platform safe; compiled clean for JVM.

**Konsist / strings / unicode.** No new user-facing strings, no `Text("…")` literals, no `take(n)`/`substring` on user text. strings.xml / values-da parity, AutoMirrored, M3 roles — all N/A (no UI strings or icons added).

**Test.** `ShareCacheArchivedFilterTest` added and passing — covers active-present / archived-absent, plus a defensive case asserting a non-Active non-Archived state (`Left`) still appears, locking in the "exclude Archived only" intent.

**Non-blocking — note, not a change request:**
The main list uses a *whitelist* (`ConversationListViewModel.kt:1045-1049`: Active/Left/RejoinPending/Removed) while the share surfaces use a `!= Archived` *blacklist*. So `Deleted`/`Invalid`/`RejoinPending` etc. conversations that the main list hides could still surface as share targets. The spec explicitly chose this ("exclude `Archived` only — literal to the task"), and the second test documents it — so this is intended, just worth a conscious eye if the share picker ever shows an odd target.

Verdict: approve.

PR #661, opened 2026-06-05
Quick-gallery multi-select (long-press, numbered send-order badges)

2002Bishwajeet:
## What & why

The in-chat quick-selection gallery strip (`AttachmentGallery`, mobile-only) only allowed single-select: each thumbnail tap attached exactly one image. This PR adds Signal-style ordered multi-select so a user can stage several photos/videos at once before sending.

This is a **UI-only** change. The whole downstream batch path already existed and is unchanged: `AttachGalleryItem.files` is `List<GalleryImage>`, and `handleAttachGalleryItem` stages a list into the `FullScreenAttachmentEditor` where the user sends. Previously the caller just wrapped a single item in `listOf(it)`.

## The change

- `AttachmentGallery`'s callback changes from `onImageSelected: (GalleryImage) -> Unit` to `onImagesSelected: (List<GalleryImage>) -> Unit`.
- **Interaction (mirrors Signal):**
  - Not in multi-select: a single tap attaches immediately — `onImagesSelected(listOf(item))` — preserving today's fast path.
  - Long-press enters multi-select (`combinedClickable`, `ExperimentalFoundationApi`) and selects the long-pressed item first.
  - While in multi-select, tap toggles membership instead of attaching.
- **Ordered selection:** `selectedIds` is an ORDERED `List<String>` keyed on `GalleryImage.id` (already the LazyRow key). The numbered send-order badge is `selectedIds.indexOf(id) + 1`, computed at render and never stored.
- **Visuals:** selected thumbs get a 2.dp `MaterialTheme.colorScheme.primary` border and a numbered primary-tinted circle badge (top-end); unselected thumbs show an outlined `Icons.Outlined.Circle` marker (`cd_not_selected`) for accessibility parity with `CreateConversationScreen`.
- **Confirm:** a `Send (N)` `ElevatedButton` appears when multi-select is active with a non-empty selection; on confirm it resolves `selectedIds` back to items in selection order (`associateBy { id }` + ordered `mapNotNull`), calls `onImagesSelected(list)`, and resets selection + `multiSelect`.
- **Cap:** a single batch is gated to max 32 items at the toggle.
- **Caller:** `ConversationContent.kt` now forwards the list straight into the existing `ConversationListUiAction.AttachGalleryItem(files = images)` action. `AttachmentHandler` / `UiAction` / send path are untouched.

## Strings

Two new strings added to both `values/strings.xml` and `values-da/strings.xml`, referenced via `stringResource` so the Konsist `ArchitectureTest` stays green (no raw `Text("...")` / `$`-interpolated literals):

- `chat_gallery_selection_index` = `%1$d` (badge number)
- `chat_gallery_send_count` = `Send (%1$d)` / `Send (%1$d)` (DA)

## Verified

- `./gradlew :homebase-chat:compileKotlinJvm` — BUILD SUCCESSFUL.
- Confirmed no raw `Text("...")` literals in the changed file (badge number and send count both go through `stringResource(..., n)`).
- Confirmed `AttachGalleryItem.files` is `List<GalleryImage>`, so the batch flows through the unchanged send path.

## Cross-platform notes

`AttachmentGallery` is gated behind `isMobile()` (Android + iOS only) because it reads the OS photo library via `GalleryCache`. Desktop/Web already multi-pick through FileKit and are not touched by this PR.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/AttachmentOptions.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/AttachmentOptions.kt
@@ -159,6 +159,10 @@ fun AttachmentGallery(
                 }
             }
         } else {
+            // Gallery strip and confirm control stacked in a Column so the Send-(N)
+            // pill sits in its own row BELOW the 160dp thumbnail strip and never
+            // overlaps (occludes) the leading thumbnails.
+            Column(modifier = Modifier.fillMaxWidth()) {
             Box(
                 modifier = Modifier
                     .fillMaxWidth()
@@ -318,16 +322,20 @@ fun AttachmentGallery(
                         }
                     }
                 }
-                // Confirm control: while in multi-select with a non-empty ordered selection,
-                // resolve selectedIds to items (associate by id, ordered by selectedIds) and
-                // hand the batch to the existing send path, then reset.
-                if (multiSelect && selectedIds.isNotEmpty()) {
-                    val byId = galleryItems.associateBy { it.id }
-                    val sendCount = selectedIds.size
+            }
+            // Confirm control: while in multi-select with a non-empty ordered selection,
+            // resolve selectedIds to items (associate by id, ordered by selectedIds) and
+            // hand the batch to the existing send path, then reset. Anchored in its own
+            // row BELOW the 160dp strip so it never covers the thumbnails.
+            if (multiSelect && selectedIds.isNotEmpty()) {
+                val byId = galleryItems.associateBy { it.id }
+                val sendCount = selectedIds.size
+                Row(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(horizontal = 16.dp, vertical = 8.dp)
+                ) {
                     ElevatedButton(
-                        modifier = Modifier
-                            .align(Alignment.BottomStart)
-                            .padding(16.dp),
                         onClick = {
                             val ordered = selectedIds.mapNotNull { byId[it] }
                             resetSelection()
@@ -339,6 +347,7 @@ fun AttachmentGallery(
                         Text(stringResource(MR.string.chat_gallery_send_count, sendCount))
                     }
                 }
+            }
             }
         }
     }
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/AttachmentOptions.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/AttachmentOptions.kt
@@ -3,8 +3,10 @@ package id.homebase.chat.widget
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.slideInVertically
 import androidx.compose.animation.slideOutVertically
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
-import androidx.compose.foundation.clickable
+import androidx.compose.foundation.border
+import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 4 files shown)


2002Bishwajeet:
Automated review (Wave 1) — Quick-gallery multi-select (task E)

**Verdict: approve** (one non-blocking layout nit + two minor i18n notes)

Reviewed the diff via `gh pr diff` against the spec (docs/.../2026-06-05-chat-kmp-7-task-batch-design.md, section E) without touching the working tree.

### Root cause / approach — correct, not a symptom patch
The batch path already existed end-to-end (`AttachGalleryItem.files: List<GalleryImage>` → `AttachmentHandler.handleAttachGalleryItem` appends `newFiles` into the `FullScreenAttachmentEditor` overlay). This PR is purely the missing UI affordance on `AttachmentGallery`, exactly as the spec scoped it ("No ViewModel/handler/send-path change"). The single-tap fast path is preserved; long-press enters multi-select; tap toggles; ordered `List<String>` keyed on `GalleryImage.id` drives `indexOf+1` send-order badges computed at render (never stored); confirm re-resolves ids against live `galleryItems` via `associateBy` before dispatch. Matches the Signal-derived design (ordered list, numbered badge, Send-(N) pill, max 32). 

### Checks
- **KONSIST literal Text:** PASS. Both new `Text(...)` go through `stringResource` — badge `chat_gallery_selection_index` (`%1$d`) and `chat_gallery_send_count` (`Send (%1$d)`). No raw / `$`-interpolated literals introduced.
- **take(n)/substring on user text:** PASS. None. The only values formatted into strings are `Int`s (selection number, send count) — no surrogate risk.
- **maxLines/overflow/platform branching:** PASS. Feature stays inside the existing `isMobile()` gate (Android/iOS only); caller in ConversationContent.kt:1568 is also `isMobile()`-gated. Desktop/Web (FileKit multi-pick) untouched.
- **Cross-platform parity:** PASS. `combinedClickable` + `@OptIn(ExperimentalFoundationApi::class)` are already used across commonMain in this repo (GroodleBubble, EventBubble, MessageBubbleRaw, etc.), so no Android/iOS/Desktop/Web breakage. All-commonMain change.
- **No wire/app-data/sender change:** PASS. Only 2 .kt + 2 strings.xml changed; no model/descriptor/protocol/handler files. Send path unchanged.
- **M3 / padding / AutoMirrored / lifecycle:** PASS. `colorScheme.primary`/`onPrimary`, `typography.labelSmall`, start/end-safe `padding()`, no directional icons needing AutoMirrored, `collectAsStateWithLifecycle()` retained. Unselected marker mirrors CreateConversationScreen's `Icons.Outlined.Circle` + `cd_not_selected`.
- **New strings in BOTH values + values-da:** PASS for the two NEW strings — both present in values/strings.xml and values-da/strings.xml.
- **Test added where spec called for one:** N/A. Spec E does not request a test (UI-only, no handler change); tests are mandated for C and D only. Acceptable.

### Non-blocking
1. **Send-(N) button overlaps the thumbnail row.** The gallery `Box` is a fixed `height(160.dp)` and thumbnails are `size(160.dp)`; the confirm `ElevatedButton` is `align(BottomStart).padding(16.dp)`, so it floats *on top of* the first thumbnails rather than below the row. Functional, but the button covers the very items being selected. Consider reserving vertical space (or anchoring the confirm pill outside the 160dp strip) so it doesn't occlude thumbnails. (AttachmentOptions.kt — confirm-control block.)
2. **`cd_not_selected` missing in values-da.** Pre-existing gap (string predates this PR; already used by CreateConversationScreen), but this PR becomes its second consumer. The Danish build will fall back to the English content description. Cheap to add `Ikke valgt` while here. (values-da/strings.xml)
3. **`chat_gallery_send_count` DA value is an untranslated copy** (`Send (%1$d)` == English). "Send" is valid Danish so harmless, but flag it as not-actually-localized. (values-da/strings.xml:302)

Nice touch: the `else if (selectedIds.size < MaxGallerySelection)` gate correctly caps adds at 32 while still allowing removal at the cap.

PR #662, opened 2026-06-05
Collapse long message bodies with an inline Read more expander

2002Bishwajeet:
## Root cause / motivation

Plain-text message bodies are rendered in `MessageBubbleRaw`'s bespoke `Layout` with **no `maxLines`** on either render path — the search-highlight `Text` and the richeditor `RichText`. A body that fits inside the message header (< 7 KB, so `hasMore = false` and there is no payload spill) but is much taller than a screenful therefore renders fully expanded on mobile, pushing surrounding messages far off-screen.

This is **distinct** from the existing `show_more` affordance (gated on `message.hasMore`), which is the payload-spill download for bodies over 7 KB. Task F targets the header-resident case and is a pure UI/layout improvement — no wire format, message app-data schema, or sender-cap change.

## The fix (all scoped inside `MessageBubbleRaw`, no caller-signature change)

- `private const val CollapsedBodyMaxLines = 10` (top-level). It is only ever passed to Compose `maxLines` (clamped internally by text layout) — never used as an array index or loop bound, per the CLAUDE.md `Int.MAX_VALUE` warning.
- Transient state: `var bodyExpanded by remember(message.id) { mutableStateOf(false) }` — `remember`, **not** `rememberSaveable`, keyed on message identity so recycled bubbles don't inherit a stale expanded flag.
- `val bodyMaxLines = if (isMobile() && !bodyExpanded) CollapsedBodyMaxLines else Int.MAX_VALUE` — **desktop always shows the full body and never renders an expander**.
- Both body render paths now pass `maxLines = bodyMaxLines` and `overflow = TextOverflow.Ellipsis`. The richeditor `RichText` supports `maxLines`/`overflow`/`softWrap`/`onTextLayout` (verified against the 1.0.0-rc14 sources). The **emoji-only branch is left untouched**.
- Overflow detection reuses the existing `onTextLayout` capture — no screen-height measurement:
  `val isBodyTruncated = isMobile() && !bodyExpanded && (textLayoutResult?.hasVisualOverflow == true)`.
- **Layout integration (the load-bearing part):** the Read more/less chip renders **inside the existing show-more `Layout` slot** (the `Box` that previously held only the `show_more` `Text`), choosing payload-spill `show_more` vs body `read_more`/`read_less` by message state. The custom `Layout`'s **child count and the `textIndex` / `showMoreIndex` / `infoIndex` math are unchanged**, so the timestamp-on-last-line measurement (which reads the now-capped `textLayoutResult` and the `showMorePlaceable` height) keeps working.
- The chip carries its own `Modifier.clickable { bodyExpanded = !bodyExpanded }`, which **consumes the tap** so it does not bubble to the bubble's `combinedClickable` / long-press overlay.

## Strings

New strings, kept separate from the payload-spill `show_more`:
- `chat_message_read_more` ("Read more" / da "Læs mere")
- `chat_message_read_less` ("Read less" / da "Læs mindre")

Added to both `values/strings.xml` and `values-da/strings.xml`. No literal `Text("…")` is introduced — the chip label comes from `stringResource(...)`, satisfying the Konsist `ArchitectureTest` gate.

## What I verified

- `:homebase-chat:compileKotlinJvm` — **BUILD SUCCESSFUL**, no warnings.
- `:homebase-chat:compileAndroidMain` — **BUILD SUCCESSFUL**.
- Confirmed via the rc14 sources that `RichText` exposes `maxLines`, `overflow`, `softWrap`, and `onTextLayout`.
- Confirmed the custom-`Layout` child count is unchanged (the chip lives in the pre-existing show-more `Box`), so `textIndex`/`showMoreIndex`/`infoIndex` and the timestamp placement math are untouched.

## Cross-platform notes

- iOS/wasmJs were not compiled locally (CI `build-check.yml` / `test.yml` cover them). The change is pure common Compose using only APIs already present on every target; `isMobile()` gates the cap so Desktop (JVM) and Web behave as before — full body, no expander.
- Risk area is the bespoke `Layout` (called out in the task as highest-risk). Mitigated by not changing the child set: the chip reuses the existing slot, so index/measure math is identical to before. The only new measured content is the chip text inside a slot that already existed.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
@@ -209,7 +209,14 @@ sealed class MessageListContentModel(val id: String) {
     data class Message(
         val message: MessageUiModel,
         val clusterPosition: MessageClusterPosition = MessageClusterPosition.ALONE,
-    ) : MessageListContentModel(message.id.toString() + message.versionTag.toString() + message.hasMore)
+    // NOTE: do NOT add message.hasMore to this key. When "Read more" downloads a
+    // spilled body, hasMore flips true->false; if it were in the key the LazyColumn
+    // item would be recreated mid-tap, resetting the bubble's bodyExpanded state and
+    // re-showing "Read more" on the now-full body (a double "Read more"). The body
+    // already refreshes on download because textState is remember(message.content)
+    // (MessageBubbleRaw) — the historical hasMore-in-key reset (commit 297c4830) was
+    // only needed back when RichTextState was keyless.
+    ) : MessageListContentModel(message.id.toString() + message.versionTag.toString())
 
     data object UnreadSeparator : MessageListContentModel("unread-separator")
 
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/MessageBubbleRaw.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/MessageBubbleRaw.kt
@@ -47,6 +47,7 @@ import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextLayoutResult
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.mohamedrejeb.richeditor.annotation.ExperimentalRichTextApi
@@ -76,6 +77,8 @@ import id.homebase.core.util.isMobile
 import id.homebase.resources.MR
 import id.homebase.resources.chat_message_deleted
 import id.homebase.resources.chat_message_edited
+import id.homebase.resources.chat_message_read_less
+import id.homebase.resources.chat_message_read_more
 import id.homebase.resources.show_more
 import kotlinx.collections.immutable.ImmutableMap
 import kotlinx.collections.immutable.persistentListOf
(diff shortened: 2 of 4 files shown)


2002Bishwajeet:

Reviewed against the Task F design spec. The change fixes the actual root cause (header-resident bodies <7KB rendered with no `maxLines` on either render path), not a symptom. End-to-end wiring verified.

**Verified**
- **Root cause / not a symptom patch:** adds `maxLines = bodyMaxLines` + `TextOverflow.Ellipsis` to both body paths (search-highlight `Text` + `RichText`); overflow detected via the existing `onTextLayout` capture (`textLayoutResult?.hasVisualOverflow`), no screen-height measurement. Emoji-only path correctly left alone.
- **Cross-platform parity (highest risk):** confirmed against the rc14 sources that `material3.RichText` (the imported overload, MessageBubbleRaw.kt:54) accepts `maxLines`/`overflow`/`onTextLayout` in commonMain and `BasicRichText` threads all three into `BasicText` (BasicRichText.kt:147,149 + onTextLayout forwarding at :145). So `hasVisualOverflow` genuinely flips when the 10-line cap clips — the feature is not a silent no-op, and it compiles on Android/iOS/Desktop/Web (pure common Compose).
- **Platform branching:** `isMobile() && !bodyExpanded` gates the cap; desktop/web fall through to `Int.MAX_VALUE` (full body, no expander) — matches the spec default.
- **Konsist:** both new `Text` calls use `stringResource(...)`; `CollapsedBodyMaxLines = 10` is only ever passed to `maxLines` (documented, no array-index/loop-bound use per the CLAUDE.md `Int.MAX_VALUE` warning).
- **No surrogate-splitting:** no `take(n)`/`substring` on user text introduced — truncation is done by Compose text layout.
- **Strings:** `chat_message_read_more` / `chat_message_read_less` added to BOTH `values/strings.xml` and `values-da/strings.xml`; kept separate from the payload-spill `show_more`.
- **No wire/app-data/sender change.** Custom `Layout` child count unchanged — the chip reuses the existing show-more `Box` slot, so `textIndex`/`showMoreIndex`/`infoIndex` and the timestamp-on-last-line math are untouched.
- **Tap consumption:** the chip carries its own `Modifier.clickable`; as a later sibling it wins pointer dispatch over the long-press overlay Box, so the toggle does not bubble to the bubble overlay.

**Non-blocking**
- The Box-clickable region changed from `if (onShowMoreClick != null)` to `onShowMoreClick?.takeIf { message.hasMore }` (MessageBubbleRaw.kt:~496 base). This is a *tightening* (the whole-Box click is now only live when `hasMore`, which previously was a no-affordance latent click region) and is required so the body-expander region isn't double-claimed. Intentional and correct, but worth a glance.
- No unit test added. The spec did not mandate one for this UI-only layout change (cross-cutting constraints only require per-target compile verification), so non-blocking — but a small assertion that `isMobile() == true → maxLines == 10` / desktop → `MAX_VALUE`, or a render test that the chip appears on overflow, would lock the behavior.
- Cosmetic: when the body is truncated, the last visible line ends with the ellipsis glyph and `getLineRight(lastLineIndex)` includes it, so the on-last-line timestamp could sit close to the `…`. Pre-existing layout behavior (same as the `show_more` path), not a regression.

No blocking issues.

PR #663, opened 2026-06-05
Add animated GIF rendering on Android (Coil coil-gif)

2002Bishwajeet:
## Root cause

The image data path already loads full GIF payloads — `image/gif` is in
`THUMBLESS_CONTENT_TYPES` (`HomebaseImageLoader.kt`) so the whole animated
file reaches Coil. But **no animated decoder was registered on any platform**,
so Coil decoded GIFs with its default static path and rendered a single frame.

## The fix

Register Coil's GIF decoder on **Android**:

1. `gradle/libs.versions.toml` — add a `coil3-gif` alias (Coil `3.4.0`, matching the existing `coil3` version ref).
2. `homebase-common/build.gradle.kts` — add `coil-gif` to **`androidMain`** (see note below on why not `commonMain`).
3. `AppModule.android.kt` — in the `ImageLoader` `components { }` block, register
   `AnimatedImageDecoder.Factory()` **unconditionally**. It uses the platform
   `android.graphics.ImageDecoder`, which requires API 28+. Our effective
   `minSdk` is **28** (`gradle/libs.versions.toml` `android-minSdk = "28"`), so
   the earlier `Build.VERSION.SDK_INT < 28` guard with a `GifDecoder.Factory()`
   fallback was **dead code** and has been removed (along with the now-unused
   `android.os.Build` and `coil3.gif.GifDecoder` imports).

No change to the transport, thumbnail, or `THUMBLESS_CONTENT_TYPES` pipeline.
No new user-facing strings (no Konsist/strings.xml impact).

## Scope: Android-only — coil-gif has no non-Android animated decoder at any Coil version

The earlier hope was that `coil-gif` ships android+jvm+ios+wasmJs (like `coil-svg`)
and that the other targets could register an `AnimatedSkiaImageDecoder.Factory()`.
**That is not the case at any published Coil version.** Verified evidence:

- **Published Gradle module metadata** for `io.coil-kt.coil3:coil-gif` lists **only
  Android variants** (`releaseVariantReleaseApiPublication` /
  `...RuntimePublication` / `...SourcePublication`, `.aar`) — checked at both the
  latest stable `3.4.0` **and** the latest beta `3.5.0-beta01` on Maven Central.
  There is no `coil-gif-jvm`, native, or `wasmJs` publication.
- **Empirical resolve** confirms it: in `:homebase-common`'s `jvmCompileClasspath`,
  `coil-core`, `coil-compose`, and `coil-svg` all resolve their `-jvm` Skiko
  variants, while `coil-gif` does not appear at all (it lives only in `androidMain`).
- **`coil3.gif.AnimatedSkiaImageDecoder` does not exist** in any Coil version. The
  `coil3.gif` package contains only `AnimatedImageDecoder`, `GifDecoder`,
  `MovieDrawable`, `AnimatedTransformation`, `PixelOpacity`. The official
  [GIFs docs](https://coil-kt.github.io/coil/gifs/) state: *"This feature is only
  available on Android."*
- Upstream [coil-kt/coil#2347](https://github.com/coil-kt/coil/issues/2347)
  (animated images on iOS / non-Android) is **still open** and tagged
  *help wanted* — the feature is unimplemented.

Because no non-Android animated decoder exists, registering one on
Desktop/iOS/Web would not compile. `coil-gif` is therefore added to
**androidMain only**, and the desktop/native/web `AppModule`s are intentionally
left untouched.

**Net effect on non-Android targets:** Desktop, iOS, and Web continue to render
the **first frame** of a GIF (via coil-core's static `SkiaImageDecoder`) — same
as before this PR, no regression.

## What would unlock Desktop / iOS / Web

Nothing in this repo's control today. It requires **upstream Coil** to ship a
multiplatform/Skia animated decoder (the work tracked in coil-kt/coil#2347). No
Coil version through `3.5.0-beta01` provides it, and the in-flight Compose /
Skia upgrade (PR #656) does not change this — the missing piece is a Coil
artifact, not a Skia/Compose version. Revisit when Coil publishes a non-Android
`coil-gif` variant or an `AnimatedSkiaImageDecoder`.

## What I verified

- `./gradlew :homebase-core:compileAndroidMain` — **BUILD SUCCESSFUL**.
- `./gradlew :homebase-core:compileKotlinJvm` — **BUILD SUCCESSFUL** (untouched JVM path still compiles).
- Confirmed all `coil-gif` / `coil3.gif.*` references are confined to `androidMain` + build files.

## Cross-platform smoke-test status (UNVERIFIED — needs manual post-merge testing)

- **Android** — the real path here. Animated GIF playback should work on API 28+
  via `AnimatedImageDecoder`. **Smoke-test on a device:** send/receive a GIF and
  confirm it animates inline.
- **Desktop (JVM) / iOS (Skia) / Web (wasm-Skia)** — **no animation in this PR**;
  they render the first frame as before. Confirm there's no visual regression
  (GIFs still show their first frame).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -183,6 +183,7 @@ coil3-compose = { group = "io.coil-kt.coil3", name = "coil-compose", version.ref
 coil3-network = { group = "io.coil-kt.coil3", name = "coil-network-ktor3", version.ref = "coil3" }
 coil3-video = { group = "io.coil-kt.coil3", name = "coil-video", version.ref = "coil3" } # Android only
 coil3-svg = { group = "io.coil-kt.coil3", name = "coil-svg", version.ref = "coil3" }
+coil3-gif = { group = "io.coil-kt.coil3", name = "coil-gif", version.ref = "coil3" }
 zoomimage-compose-coil3 = { group = "io.github.panpf.zoomimage", name = "zoomimage-compose-coil3", version.ref = "zoomimage" }
 androidsvg = { module = "com.caverock:androidsvg-aar", version = "1.4" }
 kotlin-stdlib = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib", version.ref = "kotlinStdlib" }
--- a/homebase-common/build.gradle.kts
+++ b/homebase-common/build.gradle.kts
@@ -121,6 +121,14 @@ kotlin {
             implementation(libs.accompanist.permissions)
             implementation(libs.firebase.crashlytics)
             api(libs.coil3.video)
+            // coil-gif is an Android-only artifact in Coil 3.4.0 (com.android.library,
+            // packaging=aar, no jvm/ios/wasmJs variants — verified against the published
+            // Gradle module metadata and the coil-gif/build.gradle.kts at the 3.4.0 tag).
+            // It supplies coil3.gif.AnimatedImageDecoder / coil3.gif.GifDecoder. There is
+            // NO AnimatedSkiaImageDecoder in Coil 3.4.0 (nor 3.5.0-beta01), so animated GIF
+            // decoding is Android-only at this Coil version; the other targets fall back to
+            // coil-core's static SkiaImageDecoder (first-frame still).
+            api(libs.coil3.gif)
             implementation(libs.kermit.io)
             // kmpnotifier has no wasmJs artifact — must stay off commonMain.
             // `api` so the dep cascades transitively to androidApp (which
--- a/homebase-core/src/androidMain/kotlin/id/homebase/core/di/AppModule.android.kt
+++ b/homebase-core/src/androidMain/kotlin/id/homebase/core/di/AppModule.android.kt
@@ -1,8 +1,11 @@
 package id.homebase.core.di
 
+import android.os.Build
 import co.touchlab.kermit.Logger
 import coil3.ImageLoader
 import coil3.SingletonImageLoader
+import coil3.gif.AnimatedImageDecoder
+import coil3.gif.GifDecoder
 import coil3.memory.MemoryCache
 import coil3.video.VideoFrameDecoder
 import id.homebase.api.file.AndroidFileOperationsProvider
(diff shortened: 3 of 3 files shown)


2002Bishwajeet:
Automated review (Wave 1) — **changes-requested**

Verified the core claims against the *actual* pinned Coil 3.4.0 artifacts in the Gradle cache (not from memory). The deviation from the brief is real and correct; two accuracy issues should be fixed before merge.

### Root cause — addressed correctly (not a symptom patch)
The decoder really was missing. `image/gif` is already in `THUMBLESS_CONTENT_TYPES` so the full payload reaches Coil; registering `AnimatedImageDecoder.Factory()` is the genuine fix, not a workaround. Confirmed `coil3/gif/AnimatedImageDecoder` + `GifDecoder` ship in the `coil-gif-3.4.0.aar`.

### The brief's "all platforms" plan was infeasible at Coil 3.4.0 — independently verified
- `coil-gif:3.4.0` `.module` lists **only** `releaseVariantRelease*` AAR variants; POM `packaging=aar`, Android-only deps (`vectordrawable-animated`, `core-ktx`). No `-jvm`/`-ios*`/`-wasm-js` sibling (contrast: `coil-svg` has all of them — verified).
- `AnimatedSkiaImageDecoder` (the class the brief named for Desktop/iOS/Web) exists in **zero** Coil 3.4.0 JAR/klib — both class-listing and raw-symbol grep. Only the static `coil3.decode.SkiaImageDecoder` is in `coil-core`. Registering the brief's class would not compile.
- So Android-only is the right call; non-Android stays first-frame (no regression). Good, accurate PR writeup on this.

### BLOCKING
1. **`else { GifDecoder.Factory() }` is dead code; the "minSdk 27" rationale is wrong.** `gradle/libs.versions.toml` has `android-minSdk = "28"`, and every module (incl. `androidApp`) consumes it — effective minSdk is **28**. `Build.VERSION.SDK_INT < 28` is therefore unreachable. The code comment in `AppModule.android.kt`, the commit message, and the PR body all assert "our minSdk is 27" (that figure comes from a stale line in CLAUDE.md, not the authoritative version catalog). Either drop the `else` branch and the API guard (just `add(AnimatedImageDecoder.Factory())`), or — if you want to keep the guard defensively — fix the comment to say minSdk is 28 and the fallback is unreachable today. Leaving an incorrect rationale baked into a permanent comment is exactly what the repo's debugging guidance warns against.
2. **PR title says "across all platforms"** but the change is Android-only. Retitle to reflect reality (e.g. "Add animated GIF rendering on Android (Coil coil-gif)" — which is what the commit headline already says) so history/changelog isn't misleading.

### NON-BLOCKING
- No automated test added. Confirmed a JVM/common test is *impossible* here — `coil3.gif.*` is absent from all non-Android artifacts, so the `HeicDecoderFactoryTest` JVM pattern can't be mirrored. A Robolectric/instrumented Android test asserting `AnimatedImageDecoder.Factory().create(...)` returns non-null for `image/gif` would be the only automated option; otherwise the manual smoke-test plan in the PR body is acceptable per the brief.
- Consider a one-line `// TODO: revisit non-Android animation when Coil ships a Skia animated decoder` near the androidMain dep so the parity gap is discoverable in code, not just the PR description.

### Clean
- Scope is exactly 3 files; no strings.xml / values-da, no `Text("…")`, no `take`/`substring`, no `MessageContent`/`ChatProtocol`/`dataType`/wire-format/sender change. Konsist & i18n: N/A (no new strings). No Compose/M3 surface touched.

PR #664, opened 2026-06-05
Sticker support: render transparent images cut-out via an explicit per-image flag

2002Bishwajeet:
## Goal / root cause

Transparent PNG/WebP cut-outs ("stickers") used to render with an opaque `surfaceContainerHigh` card behind them plus a rounded-card clip, so with the default `ContentScale.Fit` the transparency was flattened onto the bubble surface and they never looked cut out. Alpha is already preserved end-to-end on every platform (ARGB_8888 / N32Premul); the only defect was the render-side backdrop.

This adds an **explicit per-image sticker flag** — a real stored flag, not content-type inference at render — auto-detected at send time, and drops the backdrop + clip for solo stickers in the bubble.

## Change per file

**Storage (additive, backward-compatible, no wire break)**
- `PayloadDescriptor.kt` — new `@Serializable DescriptorContent.ImageFile(isSticker: Boolean = false)`, parsed in `descriptorInfo()` for `image/*` content type. Blank / null / legacy / malformed JSON all fall back to `ImageFile(false)` (try/catch, never throws — mirrors the audio/video parse-failure contract). `filename()` returns `null` for `ImageFile` so the download name still falls back to `payload.key` exactly as it did for the legacy `""` case. New `descriptorContentFromImage(isSticker)` companion helper. **No `MessageAppData` version bump, no message-body change** — this rides entirely on the per-payload descriptor slot that image payloads otherwise leave `""`.

**Detection (send-side)**
- `ImageUtil.kt` — new `expect ImageUtils.hasNonOpaquePixels(srcBytes): Boolean`.
- `ImageUtil.android.kt` — Android actual: `BitmapFactory` ARGB_8888, `Bitmap.hasAlpha()` fast-path then a coarse grid (≤ ~4096 samples) + the 4 corners + centre; `getPixel` alpha scan; try/catch → false; recycles the bitmap.
- `ImageUtil.skia.kt` — skiaMain actual (covers **Desktop + iOS + Web**): `alphaType == OPAQUE` fast-path then BGRA `readPixels` coarse-grid alpha-byte scan; try/catch → false.
- Threshold **alpha < 250** (not 255) to avoid false positives from WebP/PNG compression fringe on near-opaque photos.
- `MessageThumbnailGenerator.kt` / `ThumbnailResult.kt` — thread the original source bytes out of the generator (it already reads them) so the probe runs on the **pre-thumbnail** bytes with **no second file read**.
- `MessageAttachmentBuilder.kt` — in the image branch, gate to `image/png` + `image/webp` via `ImageFormatDetector`, run the probe, and write `{"isSticker":true}` only when transparent (or `AttachmentInput.forceSticker` is set). Opaque photos keep `""`.
- `AttachmentInput.kt` — new `forceSticker: Boolean = false` hook for a later manual "Send as sticker" toggle (v1 is auto-detect only).

**Render (100% commonMain, one path for all platforms)**
- `MediaMessage.kt` — `isSticker = payloads.size == 1 && (payloads[0].descriptorInfo() as? ImageFile)?.isSticker == true`; drops the `surfaceContainerHigh` background for the solo-image branch only (multi-image grids keep the letterbox fill); passes `isSticker` to `MediaItem`.
- `MediaItem.kt` — `isSticker: Boolean = false` param; applies `.clip(shape)` only when `!isSticker`; keeps `ContentScale.Fit`.
- `FullScreenMediaViewer.kt` — no functional change; added a comment noting the deliberate divergence (bubble = transparent, full-screen = solid surface backdrop).
- `PendingMessageBubble.kt` — unchanged; the optimistic placeholder keeps the normal fill for v1 (covered by the upload overlay, brief lifetime).

## Compile results (all targets)

| Target | homebase-api | homebase-chat |
|---|---|---|
| JVM (`compileKotlinJvm`) | PASS | PASS |
| Android (`compileAndroidMain`) | PASS | PASS |
| iOS (`compileKotlinIosSimulatorArm64`) | PASS | PASS |
| Web (`compileKotlinWasmJs`) | PASS | PASS |

The `expect`/`actual` alpha probe resolves on all four targets (two actuals only: androidMain + skiaMain).

## Tests

JVM (`homebase-api:jvmTest` + `homebase-chat:jvmTest`), all green:
- **PayloadDescriptor round-trip** — `ImageFile(true)` round-trips; blank / null / malformed all → `isSticker=false` with no throw; unknown keys ignored; `filename()` == null for image descriptors.
- **hasNonOpaquePixels** (skiaMain actual on JVM) — opaque PNG → false, fully-transparent → true, single transparent corner → true, alpha-252 fringe → false, opaque JPEG → false, real `shirt_transparent.png` fixture → true, corrupt/empty bytes → false (no throw).
- **MessageAttachmentBuilder** — transparent PNG → sticker descriptor, opaque JPEG → `""`, `forceSticker=true` on an opaque image → sticker; descriptor round-trips to `ImageFile(isSticker=true)`.
- Konsist `ArchitectureTest` passes (no new user-facing strings).

## Cross-platform + wire-compat notes

- Render change is 100% commonMain — identical on Android/iOS/Desktop/Web. The only per-platform code is the alpha probe (androidMain + skiaMain).
- **No wire break**: additive per-payload descriptor field; older clients ignore `descriptorContent` for images and a blank/legacy descriptor parses to `isSticker=false`. `MessageAppData.version` and the markdown/RichText body contract are untouched.
- Behaviour change worth calling out: `descriptorInfo()` for an `image/*` payload now returns `ImageFile` instead of `File`. Production never wrote a filename to an image descriptor (image payloads always wrote `""`), and `filename()` still returns `null` → download name falls back to `payload.key`. One existing test that asserted the old `image/jpeg`→`File("filename.jpg")` shape was retargeted to `application/pdf` to preserve its "generic file branch is unaffected" intent.
- Manual cross-platform smoke (recommended post-merge): send a cut-out transparent PNG and confirm no grey card behind it on Android + Desktop + iOS + wasmJs; confirm an ordinary photo still shows the letterbox backdrop; tap to full-screen and confirm the solid surface backdrop.

### DIFF
--- a/homebase-api/src/androidMain/kotlin/id/homebase/api/image/ImageUtil.android.kt
+++ b/homebase-api/src/androidMain/kotlin/id/homebase/api/image/ImageUtil.android.kt
@@ -292,6 +292,59 @@ actual object ImageUtils {
         return ImageSize(options.outWidth, options.outHeight)
     }
 
+    actual fun hasNonOpaquePixels(srcBytes: ByteArray): Boolean {
+        var bitmap: Bitmap? = null
+        return try {
+            val options = BitmapFactory.Options().apply {
+                inPreferredConfig = Bitmap.Config.ARGB_8888
+            }
+            bitmap = BitmapFactory.decodeByteArray(srcBytes, 0, srcBytes.size, options)
+                ?: return false
+            // Fast path: the bitmap has no alpha channel at all → fully opaque.
+            if (!bitmap.hasAlpha()) return false
+
+            val w = bitmap.width
+            val h = bitmap.height
+            if (w <= 0 || h <= 0) return false
+
+            // Coarse grid (≤ ~ALPHA_PROBE_MAX_SAMPLES) so a huge image stays cheap.
+            val cols = minOf(w, ALPHA_PROBE_GRID)
+            val rows = minOf(h, ALPHA_PROBE_GRID)
+            for (gy in 0 until rows) {
+                val y = (gy.toLong() * (h - 1) / maxOf(1, rows - 1)).toInt()
+                for (gx in 0 until cols) {
+                    val x = (gx.toLong() * (w - 1) / maxOf(1, cols - 1)).toInt()
+                    if ((bitmap.getPixel(x, y) ushr 24) < ALPHA_OPAQUE_THRESHOLD) return true
+                }
+            }
+            // Always include the 4 corners + centre — where cut-out stickers
+            // carry their transparency — in case the grid stepped over them.
+            val probes = intArrayOf(
+                bitmap.getPixel(0, 0),
+                bitmap.getPixel(w - 1, 0),
+                bitmap.getPixel(0, h - 1),
+                bitmap.getPixel(w - 1, h - 1),
+                bitmap.getPixel(w / 2, h / 2),
+            )
+            probes.any { (it ushr 24) < ALPHA_OPAQUE_THRESHOLD }
+        } catch (e: Exception) {
+            Logger.w(throwable = e, tag = "hasNonOpaquePixels") { "Android alpha probe failed" }
+            false
+        } finally {
+            bitmap?.recycle()
+        }
+    }
+
+    /**
+     * Alpha below this counts as non-opaque. 250 (not 255) tolerates the
+     * compression fringe that WebP/PNG lossy re-encode adds to near-opaque
+     * photos, so an ordinary photo isn't misdetected as a transparent sticker.
+     */
+    private const val ALPHA_OPAQUE_THRESHOLD: Int = 250
+
+    /** Grid side length → ALPHA_PROBE_GRID² ≤ ~4096 samples regardless of image size. */
+    private const val ALPHA_PROBE_GRID: Int = 64
+
     @RequiresApi(Build.VERSION_CODES.R)
     actual fun warpAffine(
         srcBytes: ByteArray,
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
@@ -79,9 +79,15 @@ data class PayloadDescriptor(
                 // or corrupt descriptor never throws at render and never strips a photo's
                 // backdrop (opaque is the safe default). Mirrors the audio/video
                 // parse-failure contract above.
-                if (descriptorContent.isBlank()) {
-                    DescriptorContent.ImageFile(isSticker = false)
-                } else {
+                //
+                // Single-payload link-preview (chat_links) and location-preview (chat_loc)
+                // bubbles ALSO carry an image/* contentType but store a JSON *array* in
+                // descriptorContent — not an {"isSticker":...} object. Only attempt the
+                // ImageFile deserialize when the content actually looks like a JSON object,
+                // so the expected blank/array case is silent (no redundant parse, no warning
+                // log on the recomposition hot path) while genuine "looks like an object but
+                // failed to parse" corruption is still logged.
+                if (descriptorContent.trimStart().startsWith("{")) {
                     try {
                         OdinSystemSerializer.deserialize<DescriptorContent.ImageFile>(
                             descriptorContent
@@ -90,6 +96,8 @@ data class PayloadDescriptor(
                         Logger.w("PayloadFile.descriptorInfo", e)
                         DescriptorContent.ImageFile(isSticker = false)
                     }
+                } else {
+                    DescriptorContent.ImageFile(isSticker = false)
                 }
             }
 
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
@@ -72,6 +72,27 @@ data class PayloadDescriptor(
                 }
             }
 
+            contentType?.startsWith("image/") == true -> {
+                // Image payloads historically wrote descriptorContent = "" (no info).
+                // We now optionally store a tiny {"isSticker":true} object here. Blank /
+                // legacy / malformed all fall back to a non-sticker ImageFile so an older
+                // or corrupt descriptor never throws at render and never strips a photo's
+                // backdrop (opaque is the safe default). Mirrors the audio/video
+                // parse-failure contract above.
+                if (descriptorContent.isBlank()) {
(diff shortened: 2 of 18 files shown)


2002Bishwajeet:
## Automated review (Wave 2) — B1 Sticker support

**Verdict: approve** (one non-blocking nit). Scope confirmed B1-only — no markdown/G or F-merge changes in this PR, so the G review criteria don't apply here.

### Verified
- **Implements the plan as designed.** Explicit per-image `isSticker` flag stored in `PayloadDescriptor.descriptorContent` (no content-type inference at render), alpha auto-detect at send, render drops `surfaceContainerHigh` backdrop + rounded-card clip for solo stickers. `forceSticker` hook + FullScreenMediaViewer "keep solid surface" divergence both match the plan. Fixes the real defect (render-side backdrop), not a symptom.
- **Wire-compat: OK.** Additive `DescriptorContent.ImageFile(isSticker=false)`; blank/null/legacy/malformed all parse to `isSticker=false` (no throw — mirrors audio/video contract). Non-sticker images keep `descriptorContent=""` **byte-identical** to before. No `MessageAppData` bump, body untouched. `filename()`→null for images keeps the `payload.key` download-name fallback (actually fixes the old `File(name="")`→`""` path). The retargeted `image/jpeg`→`application/pdf` test is justified + documented.
- **Cross-platform: OK.** Built **all four targets for both `homebase-api` and `homebase-chat`** locally — JVM / Android / wasmJs / iosSimulatorArm64 all BUILD SUCCESSFUL. Two actuals (androidMain + skiaMain) correctly cover every target (skiaMain is parent of jvm/native/wasmJs). The PR's compile-results table is accurate.
- **Tests: present, meaningful, green.** Ran `PayloadDescriptorImageTest`, `HasNonOpaquePixelsTest`, `MessageAttachmentBuilderStickerTest`, `PayloadDescriptorVideoTest` — all pass. Good coverage: corner-probe, alpha-252 fringe, corrupt/empty bytes, forceSticker, real `shirt_transparent.png` fixture (tracked), unknown-keys forward-compat.
- **Conventions: OK.** No new user-facing strings (Konsist not at risk; no values-da needed). No `take/substring` on user text. No hardcoded `Color`. Probe index math is safe for 1×1 / huge images (`maxOf(1, rows-1)` guards div-by-zero; ≤~4096 samples + corners + centre).

### Non-blocking
- **`MediaMessage.kt:91`** — `val isSticker = payloads.size == 1 && (payloads[0].descriptorInfo() as? ImageFile)?.isSticker == true` is computed **un-`remember`ed in the composable body**, so it re-runs `descriptorInfo()` on every recomposition. For single-payload **link-preview (`chat_links`) and location-preview (`chat_loc`)** bubbles — which carry `image/*` contentType + a JSON-**array** `descriptorContent` — `deserialize<ImageFile>` throws (array ≠ object) → caught → emits `Logger.w("PayloadFile.descriptorInfo", e)` and returns `ImageFile(false)` **every recomposition**. Render stays correct (isSticker=false), but it's redundant parse + warning-log spam on a hot path. Suggest `remember(payloads)` around the derivation and/or skipping the `chat_links`/`chat_loc` keys. (Send side is unaffected — those previews go through their own builders, never `MessageAttachmentBuilder`'s image branch.)

crossPlatformOk: true · conventionsOk: true · wireCompatOk: true

2002Bishwajeet:
Automated review (bare sticker component) — APPROVE / no blocking issues.

Verified the new commit `491264d` ("Render stickers via dedicated bare StickerMessage like emoji-only"). It does exactly what it claims and there is no gray anywhere in the sticker path.

**1. No longer routes through MediaMessage / MediaTimestampOverlay — no gray survives.**
`MessageBubbleRaw` now branches `if (isStickerBubble) StickerMessage(...) else if (mediaOnly) { MediaMessage + MediaTimestampOverlay }`. A solo transparent image (`isSticker && mediaOnly`) takes the StickerMessage branch and never touches MediaMessage's letterbox `.background(surfaceContainerHigh)` nor `MediaTimestampOverlay` (which paints the `Color.Black @ 0.6` vertical-gradient scrim). The outer `Surface` is `Color.Transparent` + `RectangleShape` + no clip for stickers. I checked the MediaItem image branch (image path, lines 233-312) for any residual fill/scrim/letterbox: there is none — every `Color.Black`/`surfaceContainer` paint lives in the video branch, which a sticker never hits. **noGrayAnywhere: true.**

**2. Renders via the existing MediaItem(isSticker=true) encrypted-image path, bare, Fit, Dimens.Sticker.**
StickerMessage calls `MediaItem(isSticker = true, preserveAspectRatio = true, modifier = sizeIn(max = Dimens.Sticker.maxSize), imageSize = THUMB_MEDIUM)`. `isSticker=true` drops the rounded clip (`baseModifier = modifier`); `preserveAspectRatio=true` → `ContentScale.Fit`. New `Dimens.Sticker.maxSize = 160.dp`. Same encrypted `previewThumbnail ?: payload.previewThumbnail?.toEmbeddedThumb()` + key-header path MediaMessage's solo image uses; download/upload wiring identical.

**3. Timestamp/delivery shown the lightweight emoji-only way (no scrim).**
The tucked footer (`labelSmall` Text + `DeliveryStatus`, both `contentColor.copy(alpha = 0.7f)`, custom `Layout` tucking it under the bottom-end corner) is byte-for-byte the emoji-only footer styling. `contentColor` is folded into the emoji-only `onSurface` branch so it stays wallpaper-readable. No `background()` anywhere in the component.

**4. Non-sticker media unchanged.** The `else if (mediaOnly)` body is identical to the prior `if (mediaOnly)` (MediaMessage + MediaTimestampOverlay); only the guard changed. Multi-image grids, reply-media, sticker-with-text, and photos all fall through untouched.

**5. commonMain, no wire change, compiles.** StickerMessage.kt is pure commonMain (only common Compose + kotlinx.collections.immutable + kotlin.time.Instant / kotlin.uuid.Uuid imports — all matching this module's conventions). No descriptor/protocol change in this commit. All referenced symbols resolve (MediaItem, DeliveryStatus signature, EmbeddedThumb, persistentListOf).

**6. Strings.** No new user-facing string is introduced (the only `Text` is `text = messageInfoText`, a variable) — so values/ + values-da/ correctly have nothing to add; Konsist passes.

Non-blocking nits (optional):
- The `@param` KDoc says `payloads` is rendered but the body silently `return`s on empty and only ever reads `payloads[0]`; a one-line note that extra payloads are ignored would match MediaMessage's solo contract.
- The `isSticker` derivation now lives in three places (MessageBubbleRaw, MediaMessage, and implicitly the StickerMessage route) with the same `descriptorInfo() as? ImageFile`?.isSticker logic — fine for now, but a shared `fun List<PayloadDescriptor>.isSoloSticker()` would prevent the three copies drifting.

PR #665, opened 2026-06-05
Beautiful markdown: mikepenz renderer + richeditor editor (full authoring, drop underline)

2002Bishwajeet:
## Root cause / goal

The chat app used `com.mohamedrejeb.richeditor:richeditor-compose:1.0.0-rc14`
for **both** editing and read-only display of markdown bodies. Using it as a
renderer is the source of the in-tree fragility the spec flagged:

- `applyMarkDownContent` wraps `setMarkdown` in a try/catch plus a hand-rolled
  `fixProblematicMarkdownText` workaround (documented "space after a blank
  line" crash).
- An infinite-recomposition footgun in the conversation list (`RichTextState`
  mutates Compose state during composition unless `remember(text)`-wrapped).
- Search-highlight could not coexist with markdown styling, so a **separate**
  plain-`AnnotatedString` fork was used while searching — and it crashed
  `String.subSequence` when the styled string and the highlight string had
  different lengths.
- Two divergent markdown→plain strip paths whose `[*_`#>\-]` regex mangled
  mid-word hyphens/underscores and dropped trailing punctuation.

Goal: a beautiful, professional editor **and** renderer ("like TipTap"). In CMP
there is no single library that is a true TipTap at all four targets, so this
ships a **hybrid**: keep richeditor as the WYSIWYG **editor** (the only CMP
inline editor that works on wasmJs/iOS), and adopt **mikepenz
multiplatform-markdown-renderer (m3 + coil3) 0.41.0** as the robust read-only
**renderer**. Both speak CommonMark, so the markdown wire string round-trips
between them and to other Homebase clients.

## Change per file

**Dependencies**
- `gradle/libs.versions.toml`: `markdownRenderer = "0.41.0"`; aliases
  `markdown-renderer-m3`, `markdown-renderer-coil3`, and `jetbrains-markdown`
  (0.7.3, the CommonMark engine — same version richeditor & mikepenz use, no
  conflict). richeditor kept at rc14 (newest; no stable exists).
- `homebase-chat/build.gradle.kts`: add the two mikepenz deps to commonMain.
- `homebase-api/build.gradle.kts`: add `org.jetbrains:markdown` for the strip
  helper.

**Renderer**
- **NEW `homebase-chat/.../widget/MarkdownRenderer.kt`** — single
  `ChatMarkdown(content, color, style, searchQuery, isCurrentSearchResult,
  maxLines, overflow, onTextLayout, modifier)` wrapping mikepenz `Markdown(...)`
  with `markdownColor()/markdownTypography()/markdownDimens()` using only M3
  roles (inline code = monospace `surfaceVariant` chip; fenced code container;
  blockquote accent bar; headings on the M3 type scale; list spacing). A custom
  `paragraph` component surfaces the last text layout via `onTextLayout` for the
  bubble's timestamp-fit math. **Search-highlight is folded in here**: when a
  query is active, it renders the markdown's *rendered plain* form with
  highlight spans, so offsets are computed against exactly the drawn string —
  the stale-offset `subSequence` crash class is structurally impossible.
- `MessageBubbleRaw.kt` — body renders through `ChatMarkdown`; deleted the
  `RichTextState`/`applyMarkDownContent` block and the separate `highlightedText`
  fork. Emoji-only branch and the show-more slot untouched.
- `FullScreenMediaViewer.kt` — caption renders through `ChatMarkdown`.
- `ConversationItem.kt` (list preview) and `MessageSearchItem.kt` (search row)
  are 1-line sites → strip to plain via `markdownToPlainPreview` + plain `Text`
  (the block renderer would blow up a one-liner with headings/lists). The search
  row folds highlight into that plain string. The search query is threaded from
  `ConversationListPane.kt`.

**Composer (`MessageInputBar.kt`)**
- Re-enabled inline code; added heading (cycles Normal→H1→H2), blockquote,
  fenced code block, a link-insert dialog, and a nested-list (indent) button.
  Each new button has a `stringResource` content description.
- **Dropped the underline button** — underline has no CommonMark form, so
  `toMarkdown()` (the wire string) would silently lose it and it would not
  round-trip to other Homebase clients or render in mikepenz.
- Replaced the hardcoded `Color(0xFF393B3D)` toolbar dividers with
  `MaterialTheme.colorScheme.outlineVariant`.
- `RichTextStyleButton.kt` gains an optional localized `contentDescription`.

**Unified strip path**
- **NEW `homebase-api/.../util/MarkdownPlain.kt`** —
  `markdownToPlainPreview(markdown, maxCodePoints)` parses with CommonMark and
  walks the AST (denylist of markup tokens, keep link/image alt, collapse
  whitespace, `truncateToCodePoints`). `ChatMessageSizer.preview(400)` and
  `String.stripMarkdownForPreview()` now delegate to it.

**RichTextExtensions.kt**
- `applyMarkDownContent` + `fixProblematicMarkdownText` **KEPT** — they remain
  the editor-load path for the Vault note editor (`VaultNoteEditorScreen` seeds
  the `RichTextState` editor from a saved markdown body, which still needs the
  `setMarkdown` safety net). Only the renderer-side callers were removed.
  `applyDefaultStyling` stays (the editor uses it). *(This deviates from the
  plan's "remove applyMarkDownContent": it has a live editor caller, so removal
  would break the Vault note editor.)*

## Compile results — all four targets pass

| Target | Task | Result |
|---|---|---|
| JVM/Desktop | `:homebase-chat:compileKotlinJvm` | ✅ |
| Android | `:homebase-chat:compileAndroidMain` | ✅ |
| iOS (sim arm64) | `:homebase-chat:compileKotlinIosSimulatorArm64` | ✅ |
| Web (wasmJs) | `:homebase-chat:compileKotlinWasmJs` | ✅ |

`homebase-api` (the new `org.jetbrains:markdown` dep) and `homebase-core`
(Vault note editor path) also compile on these targets. **mikepenz m3 + coil3
resolve on wasmJs and iOS** — the decisive dependency gate.

## Tests (JVM, all green)

- **NEW `MarkdownPlainTest`** (7): mid-word `-`/`_` preserved, `#`/`*`/`` ` ``/`>`
  stripped, link + image alt kept, newlines collapsed, code-point-safe emoji
  truncation, empty input.
- `RichTextRenderingTest` repointed to `ChatMarkdown`: bold/italic renders, the
  `gg\n\n  gg` space-after-blank-line input renders **without crashing** (proves
  the parser swap fixed the documented crash), and a heading+quote+code+list
  body renders.
- `SearchHighlightTest`: keeps the 4 existing invariants and adds a combined
  styling+highlight case proving highlight offsets stay valid against the
  rendered plain text.
- `MessageSearchItemTest` / `ConversationMessagePreviewTest` still pass against
  the new strip path; the Konsist `ArchitectureTest` (no-literal-`Text` gate)
  passes with the new strings.

## Cross-platform + wire-compat notes

- Render path is 100% commonMain; mikepenz and richeditor both sit in commonMain
  with no expect/actual.
- **Wire format unchanged**: the message body stays a markdown string
  (`MessageAppData` v≥1; legacy RichText JSON for v==null via the untouched
  `getPlainTextFromRichText`). Dropping underline is what keeps the markdown
  round-trip to other Homebase clients clean.
- New strings added to **both** `values/strings.xml` and `values-da/strings.xml`.

## Task F merge awareness

Task F (PR #662, branch `collapse-long-message-body-read-more`) adds a read-more
cap (`maxLines` + `hasVisualOverflow` + inline expander) to `MessageBubbleRaw`'s
body render. This PR branches off `main` **without** F. `ChatMarkdown` already
exposes `maxLines` / `overflow` / `onTextLayout`, so F integrates cleanly at the
bubble body render at merge time — note this overlap when merging the two.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -149,6 +149,7 @@ kermit-io = { module = "co.touchlab:kermit-io", version.ref = "kermit" }
 okio = { module = "com.squareup.okio:okio", version.ref = "okio" }
 okio-fakefilesystem = { module = "com.squareup.okio:okio-fakefilesystem", version.ref = "okio" }
 richeditor-compose = { module = "com.mohamedrejeb.richeditor:richeditor-compose", version.ref = "richeditorCompose" }
+markdown-renderer = { module = "com.mikepenz:multiplatform-markdown-renderer", version.ref = "markdownRenderer" }
 markdown-renderer-m3 = { module = "com.mikepenz:multiplatform-markdown-renderer-m3", version.ref = "markdownRenderer" }
 markdown-renderer-coil3 = { module = "com.mikepenz:multiplatform-markdown-renderer-coil3", version.ref = "markdownRenderer" }
 jetbrains-markdown = { module = "org.jetbrains:markdown", version = "0.7.3" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -47,9 +47,14 @@ okio = "3.17.0"
 richeditorCompose = "1.0.0-rc14"
 # Read-only markdown RENDERER (mikepenz). Built on org.jetbrains:markdown:0.7.3
 # (the CommonMark engine richeditor already pulls in — no version conflict).
-# Publishes m3 + coil3 artifacts for android/jvm/iosArm64/iosSimulatorArm64/wasmJs
-# at 0.41.0 (verified on Maven Central). Drives every read-only display site.
-markdownRenderer = "0.41.0"
+# Publishes m3 + coil3 artifacts for android/jvm/iosArm64/iosSimulatorArm64/wasmJs.
+# Pinned to 0.40.2 (NOT 0.41.0): 0.41.0's bytecode references the synthetic
+# `androidx.compose.ui.graphics.painter.Painter.$stable` field, which Compose 1.10
+# removed — so 0.41.0 throws NoSuchFieldError at runtime on our Compose 1.10.3 the
+# moment it renders an inline markdown image (MarkdownInlineImageWithSize). 0.40.2
+# is the newest release whose bytecode carries no removed-field reference (verified
+# via javap), and 0.39.x release notes confirm the 0.39–0.40 line targets CMP 1.10.
+markdownRenderer = "0.40.2"
 sqldelight = "2.3.2"
 sqlcipher = "4.13.0"
 kotlinx-coroutines = "1.10.2"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -40,7 +40,16 @@ junit = "4.13.2"
 kotlin = "2.3.21"
 browser = "1.10.0"
 okio = "3.17.0"
+# richeditorCompose: 1.0.0-rc14 is the newest published build (2026-04-26); no
+# stable release exists, so there is no upgrade target. Kept as the WYSIWYG
+# composer/editor (the only CMP library that does true inline editing on all
+# four targets, incl. wasmJs). Read-only rendering uses mikepenz below.
 richeditorCompose = "1.0.0-rc14"
+# Read-only markdown RENDERER (mikepenz). Built on org.jetbrains:markdown:0.7.3
+# (the CommonMark engine richeditor already pulls in — no version conflict).
+# Publishes m3 + coil3 artifacts for android/jvm/iosArm64/iosSimulatorArm64/wasmJs
+# at 0.41.0 (verified on Maven Central). Drives every read-only display site.
+markdownRenderer = "0.41.0"
 sqldelight = "2.3.2"
 sqlcipher = "4.13.0"
 kotlinx-coroutines = "1.10.2"
@@ -140,6 +149,9 @@ kermit-io = { module = "co.touchlab:kermit-io", version.ref = "kermit" }
 okio = { module = "com.squareup.okio:okio", version.ref = "okio" }
 okio-fakefilesystem = { module = "com.squareup.okio:okio-fakefilesystem", version.ref = "okio" }
 richeditor-compose = { module = "com.mohamedrejeb.richeditor:richeditor-compose", version.ref = "richeditorCompose" }
+markdown-renderer-m3 = { module = "com.mikepenz:multiplatform-markdown-renderer-m3", version.ref = "markdownRenderer" }
+markdown-renderer-coil3 = { module = "com.mikepenz:multiplatform-markdown-renderer-coil3", version.ref = "markdownRenderer" }
+jetbrains-markdown = { module = "org.jetbrains:markdown", version = "0.7.3" }
 sqldelight-runtime = { module = "app.cash.sqldelight:runtime", version.ref = "sqldelight" }
 sqldelight-coroutines-extensions = { module = "app.cash.sqldelight:coroutines-extensions", version.ref = "sqldelight" }
 sqldelight-android-driver = { module = "app.cash.sqldelight:android-driver", version.ref = "sqldelight" }
--- a/homebase-api/build.gradle.kts
+++ b/homebase-api/build.gradle.kts
@@ -216,6 +216,10 @@ kotlin {
             implementation(libs.kotlinx.immutableCollections)
             implementation(libs.coil3)
             implementation(libs.okio)
+            // CommonMark AST parser for markdownToPlainPreview (MarkdownPlain.kt).
+            // Same engine the chat renderer (mikepenz) and editor (richeditor)
+            // use, so the preview strip grammar mirrors the rendered output.
+            implementation(libs.jetbrains.markdown)
         }
         commonTest.dependencies {
             implementation(libs.kotlin.test)
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/util/MarkdownPlain.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/util/MarkdownPlain.kt
@@ -6,6 +6,7 @@ import org.intellij.markdown.MarkdownTokenTypes
 import org.intellij.markdown.ast.ASTNode
 import org.intellij.markdown.ast.getTextInNode
 import org.intellij.markdown.flavours.commonmark.CommonMarkFlavourDescriptor
+import org.intellij.markdown.flavours.gfm.GFMFlavourDescriptor
 import org.intellij.markdown.parser.MarkdownParser
 
 /**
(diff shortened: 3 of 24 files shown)


2002Bishwajeet:
Automated review (Wave 2) — **APPROVE**

Verdict: **approve**. Implements the planned Option-D hybrid (richeditor stays the WYSIWYG **editor**, mikepenz `0.41.0` becomes the read-only **renderer**) and fixes the real root causes — not symptom patches. All 4 compile targets independently re-verified in a clean worktree; all JVM tests + Konsist pass.

### Verified
- **Real fix, planned approach.** `ChatMarkdown` (MarkdownRenderer.kt) is the single renderer; the richeditor `RichTextState` round-trip, the `fixProblematicMarkdownText` workaround for display, the recomposition footgun (ConversationItem) and the separate search-highlight fork are all gone. The space-after-blank-line parser crash is now a passing regression test.
- **Wire compat (G).** Body stays a markdown string (`textFieldState.toMarkdown()` at MessageInputBar.kt:272). Underline fully removed (no `FormatUnderlined`/`TextDecoration.Underline`) so nothing emits non-CommonMark. Legacy `getPlainTextFromRichText` (RichText.kt:54, v==null path) untouched. No older-client break.
- **4 read-only sites routed.** Bubble (MessageBubbleRaw.kt:450) + caption (FullScreenMediaViewer.kt:278) → `ChatMarkdown` block render with highlight folded in; the two 1-line sites (ConversationItem.kt, MessageSearchItem.kt) strip to plain via the shared `markdownToPlainPreview` (correct — block render would blow up a one-liner). `RichText`/`BasicRichText` only remain on the Vault note editor + dev example screen, both out of scope.
- **3 strip paths unified.** `ChatMessageSizer.preview` + `String.stripMarkdownForPreview` now delegate to `markdownToPlainPreview` (MarkdownPlain.kt, CommonMark AST walk — same engine the renderer uses). `getPlainTextFromRichText` correctly left alone.
- **Cross-platform gate PASSES.** Re-compiled `:homebase-chat` on `compileKotlinJvm`, `compileKotlinWasmJs`, `compileKotlinIosSimulatorArm64` — all green (the mikepenz `MarkdownText` custom-paragraph component, `annotatorSettings()`, `buildMarkdownAnnotatedString`, `markdownColor/Typography/Dimens/Padding` all resolve). m3 + coil3 + `org.jetbrains:markdown` artifacts confirmed HTTP 200 on Maven Central for wasm-js/iosarm64/jvm/android. No expect/actual needed.
- **Konsist + strings.** `ArchitectureTest` ran and passed (no-literal-`Text` gate). 11 new `chat_markdown_*` strings present in **both** values/ and values-da/. New `Text(...)` all wrap `stringResource`.
- **Conventions.** No `take(n)`/`substring`/`subSequence` on user text (all truncation via `truncateToCodePoints`). No hardcoded `Color(0x…)` (the `Color(0xFF393B3D)` divider and `0xFFFF8C00`/`0xFFFFEB3B` highlight colors are now `outlineVariant`/`tertiary`/`secondary`). `Icons.AutoMirrored.Outlined.FormatIndentIncrease` for the indent button. No left/right padding. No non-lifecycle `collectAsState()`.
- **Tests meaningful & green.** MarkdownPlainTest (7: mid-word `-`/`_` preserved, markers stripped, link/image alt kept, code-point-safe emoji truncation), RichTextRenderingTest (3, incl. the parser-crash regression + block-element render), SearchHighlightTest (6, incl. offsets-valid-against-rendered-plain). 0 skipped.
- **F-merge note present** in the PR body — `ChatMarkdown` already exposes `maxLines`/`overflow`/`onTextLayout` so PR #662 integrates at the bubble body.
- **Plan deviation handled correctly.** `applyMarkDownContent` + `fixProblematicMarkdownText` are KEPT (plan said delete) because there is a live editor caller (`VaultNoteEditorScreen.kt:157`); the doc comment documents this accurately. Deleting would have broken the Vault note editor.

### Non-blocking
- `buildSearchHighlightedText` (MessageBubbleRaw.kt:765) lowercases both strings for matching; `String.lowercase()` can change length for a few Unicode code points (e.g. `İ`), which can drift the highlight span vs. the drawn `plain`. Guarded by `coerceAtMost(plain.length)`, and this is pre-existing — not a regression introduced here. Worth a follow-up if precise highlight offsets ever matter for such input.
- Blockquote / fenced-code-block buttons insert literal CommonMark markers via `state.addTextAfterSelection("\n> ")` / `"\n```\n\n```\n"` rather than a structural editor toggle (richeditor has no paragraph-level toggle for these). Round-trip through `toMarkdown()` is plausible and the live editor renders them, but only manual cross-client verification confirms the emitted wire string. Recommend a manual round-trip smoke per the PR's own test plan.
- During active search the bubble renders the stripped-plain form (inline emphasis dropped) — an intentional, plan-sanctioned trade-off; flagging for visibility, not a defect.

crossPlatformOk: true · conventionsOk: true · wireCompatOk: true

2002Bishwajeet:
## Fix: Desktop hover crash — keep the block renderer out of the bubble's timestamp Layout

### Root cause (confirmed by a real Desktop hover stack trace + decompile)
The crash was `java.lang.IllegalStateException: layout state is not idle before measure starts`, thrown during a mouse-hover (`mouseEntered`) `measureAndLayout` on Desktop.

This PR made `ChatMarkdown`'s default body render through mikepenz `Markdown()` — a **multi-node interactive block `Column`**. In `MessageBubbleRaw` the body `Row` lives inside a bespoke custom `Layout` that, *during its own measure/place pass*:
1. reads the `textLayoutResult` `MutableState` (written by the body via `onTextLayout = { textLayoutResult = it }`) to tuck the timestamp onto the last text line, and
2. force-remeasures the reply child to converge bubble width.

A single stable `Text` node (what the old richeditor renderer effectively produced) is safe in there. The block `Column`, with native link/hover handling, re-fires `onTextLayout` and invalidates layout on Desktop hover — so the custom Layout reads freshly-written state and remeasures a child *during the layout pass* → "layout state is not idle".

The bug is the block `Column` being coupled to the timestamp-Layout's `textLayoutResult` read + force-remeasure — not hover itself. So the fix removes the block renderer from that coupling (no try/catch swallow, no remember-dedupe, no disabling hover).

### Fix
- **`homebase-api`**: new `markdownHasBlockElements(content)` next to `markdownToPlainPreview`, parsing with the same `org.jetbrains:markdown` engine (GFM flavour, matching mikepenz). `true` for headings / lists / quotes / fenced+indented code / tables / thematic breaks / HTML blocks / multi-paragraph; `false` for a single inline run; unparseable → `false`. New JVM unit test (15 cases).
- **Inline path** in `ChatMarkdown`: when the body has no block elements and no active search, render **one** `Text` over mikepenz `buildMarkdownAnnotatedString` + `annotatorSettings` (bold/italic/strike/inline-code/links). Links ride as **native Compose `LinkAnnotation`s** opened via `LocalUriHandler` — no custom `pointerInput` reading a layout result. The link `TextLinkStyles` carries **only `style`** (no hovered/pressed variant), so hover yields a structurally-identical `AnnotatedString` and `TextAnnotatedStringNode.updateText` short-circuits → `onTextLayout` never re-fires on hover, keeping this node stable inside the timestamp-tuck Layout (same cursor-only hover behaviour the old richeditor path had).
- **Block path**: the mikepenz `Markdown()` Column no longer reports `onTextLayout`, and `MessageBubbleRaw` routes block bodies to a plain `Column` with the timestamp placed below — exactly the below-placement the custom Layout already falls back to when `textLayoutResult == null`. The block Column is now entirely **outside** the force-remeasure / `textLayoutResult` path.
- Search path, wire format, strip-path unification, and composer untouched.

### Verification
- Compiles on all targets for both modules: `:homebase-api` and `:homebase-chat` × `compileKotlinJvm`, `compileAndroidMain`, `compileKotlinWasmJs`, `compileKotlinIosSimulatorArm64`.
- Tests green: new `MarkdownHasBlockElementsTest` (15), `RichTextRenderingTest` (4, incl. a new inline-link render test), `SearchHighlightTest` (6).

⚠️ The crash is a Desktop **runtime** hover reentrancy that can't be reproduced headlessly. The structural change is correct by construction (block bodies no longer feed `textLayoutResult` and live outside the force-remeasure path; inline links can't mutate text on hover) — please confirm by hovering links/text in block and inline messages on the Desktop app.

PR #666, opened 2026-06-05
Animate GIF/WebP on Desktop, iOS & Web (in-house skiko decoder)

2002Bishwajeet:
## Problem

Animated GIFs (and animated WebP) play on Android via Coil's `coil-gif` `AnimatedImageDecoder` (PR #663), but show only their **first frame, frozen** on Desktop (JVM), iOS (native), and Web (wasmJs). Coil ships **no** animated decoder for those targets — `coil-gif` is Android-only, and there is no `coil3.gif.AnimatedSkiaImageDecoder` at any version. So this implements one in-house using **skiko** (`org.jetbrains.skia.Codec`), which JVM/native/wasmJs all bundle.

## Approach: a custom Coil `Decoder` in the shared `skiaMain` source set

`AnimatedSkiaDecoder` is a `coil3.decode.Decoder` registered next to the existing `HeicDecoder` in the three non-Android `AppModule`s. Mirroring `HeicDecoder`'s structure, it lives in a new shared `skiaMain` source set (same pattern as `homebase-api`'s `skiaMain`) so one copy serves Desktop + iOS + Web.

Because it sits **inside the existing encrypted Coil pipeline**, it inherits decrypt/fetch/cache/placeholder/sizing for free:

- The bytes it decodes are the already-fetched, already-decrypted Homebase payload that `HomebaseImageFetcher` hands Coil (GIF already loads the full payload via `THUMBLESS_CONTENT_TYPES`).
- **No** second image loader, **no** re-fetch, **no** re-decrypt — the decoder only decodes bytes Coil gives it.
- It claims GIF/WebP sources (mime or magic bytes), asks the codec for `frameCount`, and returns **null for single-frame** images so static decoding falls through to Coil's default Skia decoder completely unchanged.

### How it animates without touching `HomebaseImage`

Verified against the Coil 3.4.0 bytecode: on non-Android, `coil3.Image.draw(org.jetbrains.skia.Canvas)` is the contract, and Coil wraps a non-`BitmapImage` custom `coil3.Image` in its internal `coil3.compose.ImagePainter`, whose `onDraw(DrawScope)` calls `image.draw(canvas)` **inside the Compose draw phase**.

So `SkiaAnimatedImage.draw()` reads a Compose `MutableIntState` (the current frame). Because that read happens during the draw phase, Compose's snapshot observer records it and re-invalidates the draw when it changes. A shared `SkiaFrameDriver` on `Dispatchers.Main` advances that state on the GIF's **own per-frame timeline**, and the painter repaints in lock-step. This drives animation through the normal `SubcomposeAsyncImage`/`AsyncImagePainter` path — `MediaItem`/`HomebaseImage` are not modified at all.

`shareable = false` keeps the animated image out of Coil's memory cache (verified: `MemoryCacheService` skips caching when `shareable` is false), so frames are never flattened into a stale static snapshot.

## Correctness & cost

- **Animated GIF and animated WebP** both detected (`Codec.frameCount > 1`).
- **Per-frame duration** honored (skiko `framesInfo[i].duration`), with a 20ms floor so a 0ms-delay GIF can't spin at thousands of fps.
- **Disposal / delta frames** composite correctly: frames decode into one reused scratch bitmap and Skia's 3-arg `readPixels(bitmap, frame, requiredFrame)` blends deltas; on a loop wrap or seek we replay forward from the keyframe so the composite base is correct.
- **Decode on demand** — never pre-decode all frames; decoded frames cache in a small bounded LRU (24) so large GIFs don't blow up memory.
- **Bounded list-scroll cost** — the driver only ticks while an image has `draw()`n within a 500ms grace window. GIFs scrolled out of a `LazyColumn` stop drawing, stop renewing their lease, get reaped, and the driver parks itself — so off-screen GIFs cost nothing.
- **Static images unchanged**, transparency preserved (no opaque background fill — respects the B1 sticker work), native resolution.

## Per-target compile results

All green; Android proven untouched:

| Target | `homebase-common` | `homebase-core` |
|---|---|---|
| JVM / Desktop (`compileKotlinJvm`) | PASS | PASS |
| iOS (`compileKotlinIosSimulatorArm64`) | PASS | PASS |
| Web (`compileKotlinWasmJs`) | PASS | PASS |
| Android (`compileAndroidMain`) | PASS | PASS |

## Tests

New JVM unit test `AnimatedSkiaDecoderTest` (runs on CI via `homebase-common:jvmTest`, since `skiaMain` is on the JVM classpath). Uses two hand-built minimal GIF89a fixtures:

- `isAnimatableSource` gate (GIF/WebP by mime + magic bytes; rejects JPEG/PNG).
- Decoder returns an animating `SkiaAnimatedImage` for the 2-frame GIF and **null** for the single-frame GIF (proving the static fall-through).
- `frameForElapsed` timeline math validated against the real codec's reported 500ms/750ms delays, including loop wrap.

## Relation to #663

Complementary and non-overlapping: #663 makes GIFs animate on **Android** via `coil-gif`'s `AnimatedImageDecoder`. This PR makes them animate on **Desktop, iOS, and Web** via the in-house skiko decoder. Android is not wired to this decoder (`skiaMain` excludes Android by design).

### DIFF
--- a/homebase-common/build.gradle.kts
+++ b/homebase-common/build.gradle.kts
@@ -69,6 +69,17 @@ kotlin {
     }
 
     sourceSets {
+        // Shared Skia-backed implementations for the targets that bundle skiko:
+        // Desktop/JVM, iOS/native, and Web/wasmJs. Android is intentionally
+        // excluded — it uses Coil's android.graphics-backed decoders (coil-gif
+        // animates GIFs there via AnimatedImageDecoder; see PR #663). Lets the
+        // in-house animated-GIF/WebP Skia decoder live in one place instead of
+        // three near-identical per-platform copies.
+        val skiaMain by creating { dependsOn(commonMain.get()) }
+        jvmMain.get().dependsOn(skiaMain)
+        nativeMain.get().dependsOn(skiaMain)
+        wasmJsMain.get().dependsOn(skiaMain)
+
         commonMain.dependencies {
             api(project(":homebase-api"))
             api(project(":homebase-notifshared"))
--- a/homebase-common/src/commonMain/kotlin/id/homebase/core/image/AnimatedImageDetection.kt
+++ b/homebase-common/src/commonMain/kotlin/id/homebase/core/image/AnimatedImageDetection.kt
@@ -1,0 +1,61 @@
+package id.homebase.core.image
+
+import coil3.fetch.SourceFetchResult
+
+/**
+ * Shared "could this be an animated image?" gate for the in-house Skia
+ * animated decoder (Desktop/JVM, iOS/native, Web/wasmJs).
+ *
+ * This is intentionally a *container* check, not an "is animated" check: it
+ * answers "are these bytes a GIF or WebP whose frames the Skia codec might
+ * animate?". A GIF/WebP with a single frame still passes this gate — the
+ * [AnimatedSkiaDecoder] then asks the Skia codec for the real frame count and
+ * returns null for single-frame images so Coil falls back to its default
+ * (static) Skia decoder. Keeping the cheap container sniff here means the
+ * expensive codec parse only runs for GIF/WebP, never for JPEG/PNG/HEIC.
+ *
+ * Animated GIF and animated WebP are the only animated raster formats Skia's
+ * [org.jetbrains.skia.Codec] exposes multi-frame; APNG is not reliably
+ * multi-frame across skiko versions, so we don't claim PNG here.
+ */
+fun isAnimatableSource(result: SourceFetchResult): Boolean {
+    when (result.mimeType) {
+        "image/gif", "image/webp" -> return true
+        // Any other declared image type is not an animated container we handle.
+        // (mimeType is null only when the fetcher couldn't determine it; fall
+        // through to the magic-byte sniff below.)
+        null -> Unit
+        else -> return false
+    }
+    return try {
+        val peek = result.source.source().peek()
+        val header = ByteArray(HEADER_PEEK_BYTES)
+        val read = peek.read(header)
+        if (read < MIN_HEADER_BYTES) return false
+        isGifHeader(header) || isWebpHeader(header)
+    } catch (_: Exception) {
+        false
+    }
+}
+
+/** GIF magic: "GIF87a" or "GIF89a" — only the first three bytes ("GIF") matter here. */
+private fun isGifHeader(b: ByteArray): Boolean =
+    b.size >= 3 &&
+        b[0] == 0x47.toByte() && // G
+        b[1] == 0x49.toByte() && // I
+        b[2] == 0x46.toByte() //    F
+
+/**
+ * WebP magic: "RIFF" .... "WEBP" — bytes 0..3 == RIFF, bytes 8..11 == WEBP.
+ * Bytes 4..7 are the little-endian RIFF chunk size (skipped).
+ */
+private fun isWebpHeader(b: ByteArray): Boolean =
+    b.size >= 12 &&
+        b[0] == 0x52.toByte() && b[1] == 0x49.toByte() && // RI
+        b[2] == 0x46.toByte() && b[3] == 0x46.toByte() && // FF
+        b[8] == 0x57.toByte() && b[9] == 0x45.toByte() && // WE
+        b[10] == 0x42.toByte() && b[11] == 0x50.toByte() // BP
+
+// Enough to cover the WebP RIFF/WEBP split header (12 bytes); GIF only needs 3.
+private const val HEADER_PEEK_BYTES = 12
+private const val MIN_HEADER_BYTES = 3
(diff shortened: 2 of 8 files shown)


2002Bishwajeet:
Automated review (GIF cross-platform) — VERDICT: APPROVE (no blocking issues)

Verified against skiko 0.9.37.4 and Coil 3.4.0 bytecode in the Gradle cache, plus the existing pipeline source.

**(1) It actually animates (not first-frame).** Confirmed the mechanism end-to-end against bytecode, not just the prose:
- `coil3.Image` (non-Android) interface is exactly `draw(org.jetbrains.skia.Canvas)` + `shareable` — `SkiaAnimatedImage` overrides match.
- `coil3.compose.ImagePainter.onDraw(DrawScope)` calls `getNativeCanvas(...)` then `coil3.Image.draw(skia.Canvas)` inside the Compose draw phase. So reading the `MutableIntState displayedFrame` in `draw()` registers a snapshot read → `SkiaFrameDriver` (Dispatchers.Main) advances it on the GIF's own per-frame timeline → painter re-invalidates. Real frame advance.
- Per-frame durations honored (`framesInfo[i].duration`, 20ms floor). Disposal/delta via skiko's real 3-arg `Codec.readPixels(Bitmap, frame, requiredFrame)` (signature confirmed present), with forward replay from the keyframe on seek/loop-wrap. `frameForElapsed` timeline math is self-consistent and unit-tested incl. loop wrap.

**(2) Pipeline reuse — PASS.** GIF (`image/gif` ∈ `THUMBLESS_CONTENT_TYPES`) → `loadFullPayload` → `FullPayloadByteCache.getOrLoad` (decrypt-once, coalesced) → `HomebaseImageFetcher` hands those exact bytes to Coil as `SourceFetchResult(mimeType=image/gif)` → `AnimatedSkiaDecoder` decodes them. No second image loader, no re-fetch, no re-decrypt — the decoder only decodes bytes Coil gives it. `shareable=false` keeps it out of the memory cache (`MemoryCacheService` does check `getShareable()` before caching — confirmed in bytecode). Registered only on the 3 non-Android AppModules; Android module is untouched (still only `VideoFrameDecoder`). `isAnimatableSource`/`AnimatedImageDetection.kt` mirrors `isHeicSource`/`HeicDetection.kt` precisely; `skiaMain` source-set wiring mirrors `homebase-api`'s existing `skiaMain`. Single-frame GIF/WebP → `decode()` returns null → falls through to Coil's default static Skia decoder unchanged.

**(3) Cross-platform — PASS (plausible).** All API used (`Codec.makeFromData/frameCount/framesInfo/repetitionCount/readPixels(3-arg)`, `Image.makeFromBitmap`, `Bitmap.allocPixels(imageInfo)`) exists in skiko 0.9.37.4. `mutableIntStateOf` (compose-runtime is a commonMain dep), `Dispatchers.Main`, and `TimeSource.Monotonic` are already used by commonMain code that compiles for wasmJs (ObserveAsEvents, MainThreadWatchdog) — so resolution on jvm+ios+wasm is credible. Couldn't run the compiles myself (branch not checked out; review is read-only), so the green-compile table is taken on trust but is consistent with the APIs.

**(4) Cost — PASS.** Frames decoded on demand (never pre-decoded), bounded LRU of 24 immutable frames per image; driver ticks only while an image has drawn within a 500ms grace window and parks itself when none remain → off-screen GIFs in a LazyColumn cost nothing. Static path untouched.

**(5) B1 sticker transparency — PASS.** Scratch bitmap allocated from `codec.imageInfo` (preserves source alpha type); frames drawn via `drawImageRect` with no opaque fill / `erase` / `drawColor` / background. Transparency preserved.

**(6) Konsist / Color / padding — N/A.** No new user-facing strings, no `Text("...")`, no hardcoded `Color`, no `.dp`/padding — pure decoder/image/build code. Nothing for ArchitectureTest to flag; no values/values-da additions needed.

**(7) Test — PASS.** `AnimatedSkiaDecoderTest` (jvmTest, runs on CI via unqualified `./gradlew jvmTest`): gate sniffing, factory, animated-vs-static decode + single-frame null fall-through, and `frameForElapsed` timeline incl. loop wrap. Good coverage for what's testable headless.

**Non-blocking notes:**
- Animated **WebP won't animate inline.** `THUMBLESS_CONTENT_TYPES` is `{image/gif}` only, so inline/bubble WebP loads a server thumbnail (static) and never reaches this decoder. WebP only animates where the call site sets `loadFullPayload=true` (full-screen viewer, link/location preview, vault zoomable, prefetch). The PR title says "GIF/WebP" — accurate for full-screen, but inline WebP stays static. Consider adding `image/webp` to `THUMBLESS_CONTENT_TYPES` if inline animated WebP is desired (weigh against larger inline payloads).
- PR description says "Android animates via coil-gif (PR #663)" — but **coil-gif is not in this repo/base** (no dependency, Android module has no `AnimatedImageDecoder`). This PR correctly leaves Android untouched; just note Android GIF animation is contingent on #663 actually landing, not a current fact.
- Same GIF shown at two call sites decodes twice (memory-cache bypass via `shareable=false`); acceptable and intentional.

PR #667, opened 2026-06-05
Clipboard paste: iOS image paste + preserve animated GIFs

2002Bishwajeet:
## State today

Clipboard image paste is already wired end to end through the shared composer (`MessageInputBar.onPasteImage` -> `ConversationListUiAction.AttachClipboardImage` -> `AttachmentHandler.handleAttachClipboardImage` -> `FullScreenOverlay.AttachmentData`). Whether bytes actually arrive is per-platform:

- **Android** — works (soft-keyboard `Modifier.contentReceiver`).
- **Desktop** — works (Cmd/Ctrl+V via AWT `getImageFromClipboard()`), but **animated GIFs are flattened** to a single PNG frame by `ImageIO.write(...,"png")`.
- **iOS** — does nothing: both clipboard actuals are no-op stubs, there's no Ctrl+V, and `contentReceiver` is Android-only.
- **Web** — no-op stubs (out of scope for this PR).

A separate correctness gap on **all platforms**: `handleAttachClipboardImage` hardcoded a `.png` temp-file suffix, so even when raw GIF bytes survived (Android), the wire content-type came out `image/png`.

### iOS image paste
- Real `getImageFromClipboard()` (`ClipboardImageReader.native.kt`) reads `UIPasteboard.generalPasteboard`, preferring raw bytes via `dataForPasteboardType("com.compuserve.gif")` -> `("public.png")` -> `("public.jpeg")` to preserve the original encoding (esp. animated GIF), falling back to `pasteboard.image` -> `UIImagePNGRepresentation` only as a last resort. NSData -> ByteArray via the existing `usePinned`/`memcpy` idiom, guarding the empty case. Returns null when no image is present.
- **iOS paste trigger**: iOS has no Ctrl+V and `isDesktopOrWeb()` is false, so the existing key interceptor never fires. A new **Paste** item in the attachment options sheet calls `getImageFromClipboard()` and, if non-null, dispatches `AttachClipboardImage` — the same existing pipeline, no new send path. Shown on iOS and desktop; hidden on Android (keyboard captures paste) and web (`getImageFromClipboard()` is a no-op there). Label + contentDescription are `stringResource`s added to `values/` and `values-da/`.

### Animated-GIF fidelity (end to end)
- **How the SENT content-type is derived** (investigated): for a `FileImage`, `MessageActionsHandler` builds the wire content-type via `resolveContentType(fileName = file.name, platformMimeType = file.mimeType())`. A bare temp file exposes no rich platform MIME, so it falls back to **extension-based** lookup. So the content-type comes from the **file extension**, not from the bytes. A GIF written to a `.png` temp file was therefore sent as `image/png`.
- **Fix**: `handleAttachClipboardImage` now sniffs the magic bytes (`ImageFormatDetector.detectFormat`) and picks the suffix (`.gif` / `.jpg` / `.webp` / `.png`) instead of hardcoding `.png`. The bytes are never re-encoded. A pasted GIF now lands in a `.gif` temp file -> resolves to `image/gif` -> `image/gif` is in `THUMBLESS_CONTENT_TYPES`, so receivers load the full payload and the GIF animates (per #663 Android / #666 skia).
- **Desktop raw bytes**: `getImageFromClipboard()` (jvm) now tries raw `image/gif` then `image/png` `InputStream` `DataFlavor`s before falling back to the `ImageIO` re-encode, so an animated GIF pasted on desktop keeps its frames. The `ImageIO` fallback is retained for in-memory bitmaps (screenshots) that expose no raw byte flavor.
- **Android**: already wrote raw clipboard bytes; the magic-byte suffix fix corrects the label/content-type there too. Verified.

## Tests
- `ClipboardImageSuffixTest` (homebase-chat jvmTest): GIF87a/GIF89a -> `.gif`, PNG -> `.png`, JPEG -> `.jpg`, unknown -> `.png`, plus the end-to-end assertion that the chosen suffix resolves to `image/gif` / `image/png` through the same extension lookup the send path uses.
- `ClipboardImageReaderJvmTest` (homebase-common jvmTest): a fake clipboard holding a raw `image/gif` flavor returns the original bytes unmodified (no PNG re-encode); an empty clipboard returns null.

## Per-target compile results
All green for both `:homebase-common` and `:homebase-chat`:
- `compileKotlinJvm` — pass
- `compileAndroidMain` — pass
- `compileKotlinIosSimulatorArm64` — pass (the key new target)

JVM tests for both new suites pass; `homebase-common` Konsist `ArchitectureTest` passes (new UI strings use `stringResource`).

## Overlap note
`AttachmentOptions.kt` / `MessageInputBar.kt` / `ConversationContent.kt` are also touched by #661 (gallery multi-select) and #665 (markdown). Edits here are minimal and localized (one new optional `onPasteClick` param + one LazyRow item, plus the existing onPasteImage call site) to ease the merge.

## Web — intentionally out of scope
Web paste needs an async DOM `paste`/`navigator.clipboard.read()` seam (the synchronous `getImageFromClipboard()` can't reach the browser clipboard) and a non-throwing `PlatformFile` path. Not built in this PR.

### DIFF
--- a/androidApp/src/main/kotlin/id/homebase/feed/MainActivity.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/MainActivity.kt
@@ -160,6 +160,16 @@ class MainActivity : AppCompatActivity() {
                 return
             }
 
+            // Share-into-"New Moment" deep link: homebase-fchat://moment-compose
+            // The share flow already seeded MomentCreateFlowState with the chosen
+            // media; this only steers the nav stack into the composer.
+            if (data.host == "moment-compose") {
+                Logger.i(tag = "MainActivity") { "Deep link: navigating to moment compose" }
+                notificationService.navigateToMomentCompose()
+                intent.data = null
+                return
+            }
+
             // Owner-console "Extend Permissions" return URL.
             // Path: homebase-fchat://permission-callback?status=[canceled|...]
             if (data.host == "permission-callback") {
--- a/androidApp/src/main/kotlin/id/homebase/feed/share/SharePickerScreen.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/share/SharePickerScreen.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.text.input.TextFieldState
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Check
 import androidx.compose.material.icons.filled.Close
+import androidx.compose.material.icons.outlined.AutoAwesome
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.FilledTonalButton
@@ -65,6 +66,8 @@ import id.homebase.resources.groups
 import id.homebase.resources.number_of_members
 import id.homebase.resources.recents
 import id.homebase.resources.sending_to_conversations
+import id.homebase.resources.share_picker_new_moment
+import id.homebase.resources.share_picker_new_moment_subtitle
 import id.homebase.resources.share_picker_next
 import id.homebase.resources.share_picker_send
 import id.homebase.resources.share_to
@@ -84,7 +87,8 @@ fun SharePickerScreen(
     sharedContent: SharedContent,
     hasFiles: Boolean,
     isSending: Boolean,
-    onSendToConversations: (Set<Uuid>) -> Unit,
+    showNewMomentOption: Boolean,
+    onTargetSelected: (ShareTarget) -> Unit,
     onCancel: () -> Unit,
 ) {
     val conversationsData by conversationStream.conversations.collectAsStateWithLifecycle()
@@ -186,7 +190,7 @@ fun SharePickerScreen(
                 ShareSendBar(
                     count = selectedIds.size,
                     buttonText = if (hasFiles) stringResource(MR.string.share_picker_next) else stringResource(MR.string.share_picker_send),
-                    onSend = { onSendToConversations(selectedIds) },
+                    onSend = { onTargetSelected(ShareTarget.Conversations(selectedIds)) },
                 )
             }
         },
@@ -212,6 +216,15 @@ fun SharePickerScreen(
             )
 
             HorizontalDivider()
+
+            // A "New Moment" destination sits above the conversation list (only
+            // when files are present and Moments is activated). It's an action,
+            // not a selectable checkbox row, so it dispatches immediately rather
+            // than joining the multi-select set.
+            if (showNewMomentOption && !isSending && conversationsData.dataReady) {
+                NewMomentRow(onClick = { onTargetSelected(ShareTarget.NewMoment) })
+                HorizontalDivider()
+            }
 
             if (isSending) {
                 Box(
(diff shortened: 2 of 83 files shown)


2002Bishwajeet:
## Automated review (clipboard paste) — APPROVE

Verified the iOS clipboard-image paste + animated-GIF fidelity claims against the code, compiled all touched targets, and ran the new + Konsist tests. No blocking issues.

### What I confirmed

**(1) iOS `getImageFromClipboard()` reads UIPasteboard, prefers raw bytes — ✅**
`ClipboardImageReader.native.kt` now reads `UIPasteboard.generalPasteboard` and prefers raw byte flavors in order `com.compuserve.gif` → `public.png` → `public.jpeg`, only falling back to `UIImagePNGRepresentation(pasteboard.image)` for in-memory bitmaps. No re-encode on the raw path, so an animated GIF keeps every frame. The `NSData -> ByteArray` copy uses the established `usePinned { memcpy(addressOf(0), this.bytes, this.length) }` pattern (same as `NSDataConversions.kt`, `PdfThumbnailGenerator.native.kt`) and is zero-length safe (`length <= 0 -> null`), so a zero-byte pasteboard entry can't produce a junk attachment.

**(2) Routes through the EXISTING pipeline — ✅**
The new iOS/Desktop Paste affordance in `AttachmentOptions` / `ConversationContent` dispatches `ConversationListUiAction.AttachClipboardImage(conversationId, bytes)` — the *same* action Android's keyboard `contentReceiver` already used. No new send path; bytes land in `AttachmentHandler.handleAttachClipboardImage` → temp file → `AttachmentPendingFile.FileImage` → existing overlay/send. The option is gated `if (isIos() || isDesktop())` (Android uses the keyboard receiver, web's reader is a no-op), which is correct.

**(3) GIF fidelity END-TO-END — ✅**
Traced the full chain:
- Read raw GIF bytes (iOS `com.compuserve.gif` / Desktop `image/gif` InputStream flavor — replacing the old `ImageIO`→PNG flatten).
- `clipboardImageSuffix()` sniffs magic bytes via `ImageFormatDetector.detectFormat` → `.gif`.
- Send path: `MessageActionsHandler` `FileImage` branch calls `resolveContentType(fileName=file.name, ...)` → `.gif` → `image/gif`.
- `ThumbnailGenerator` documents GIF = N=0 thumbnails (original animated payload ships, only the tiny embedded preview is generated).
- `image/gif ∈ HomebaseImageLoader.THUMBLESS_CONTENT_TYPES` → receiver loads full payload → animates.
No re-encode anywhere on the GIF path.

**(4) Cross-platform compiles — ✅** (run in the existing PR worktree)
`:homebase-common`/`:homebase-chat` `compileKotlinJvm`, `compileAndroidMain`, and `compileKotlinIosSimulatorArm64` all BUILD SUCCESSFUL. (The one `VideoPlayerSurface.native.kt` "No cast needed" warning is pre-existing, unrelated.)

**(5) Konsist / Material3 / a11y — ✅**
Both new strings (`chat_message_attachment_paste`, `cd_paste_clipboard_image`) are present in `values/` AND `values-da/` (the only two locale dirs). Labels use `stringResource(...)`; new `contentDescription` is provided on the paste icon; `Icons.Default.ContentPaste` from M3. `ArchitectureTest` (Konsist `scopeFromProject()`, scans all modules incl. homebase-chat) passes — no `Text("…")` literal introduced.

**(7) Tests — ✅**
`ClipboardImageSuffixTest` (GIF87a/GIF89a→.gif, png/jpeg, unknown→.png, plus suffix→`image/gif` via the real `detectContentTypeFromExtensionOrHint`) and `ClipboardImageReaderJvmTest` (fake `Clipboard` with raw `image/gif` flavor asserts original bytes survive; empty→null) both pass on JVM. iOS native paste isn't unit-testable here, which is a reasonable gap.

### Non-blocking

- **(6) Merge overlap not noted.** PR **#661** (`chat-gallery-multi-select`) edits the *same* `AttachmentOptions.kt` (+143/-5), `ConversationContent.kt`, and both `strings.xml` — real conflict surface with this PR's `AttachmentOption` signature change + option-list insertion. PR **#665** edits `MessageInputBar.kt` heavily, which this PR does *not* touch (so low collision there), but it also adds to both `strings.xml`. The overlap is mechanical (merge conflicts), not a correctness bug, but it isn't called out anywhere — worth a heads-up on merge order with #661.
- **Comment/code mismatch (cosmetic).** `clipboardImageSuffix`'s KDoc says "everything else (PNG, unknown raster) defaults to .png", but the code has an explicit `"image/webp" -> ".webp"` branch. Harmless (webp resolves correctly and isn't thumbless, which is fine), but the comment is slightly stale.
- **Minor reuse.** `homebase-api`'s `NSDataConversions.toByteArray()` already exists (internal to that module); the native reader reimplements an equivalent. Not worth exporting just for this.

Verdict: ship it. Blocking: none.

PR #668, opened 2026-06-05
Show 'Sticker' label in the conversation tile for sticker messages

2002Bishwajeet:
## Mechanism

The conversation-list tile's last-message preview is **computed live**, not from a denormalized stored string. `ConversationItem.kt` calls the shared `messageContentLabel(...)` Composable (`MessageContentLabel.kt`), passing `enrichedData.conversation.lastMessageFirstPayload` — the actual `PayloadDescriptor?` of the last message's first payload (populated in `ConversationUiModel`/`ConversationStream`/`ConversationMessageBump` from `msg.payloads?.firstOrNull()`). Because the real descriptor is in hand, `isSticker` is reachable at label time; nothing needs to be re-derived from stored text.

`messageContentLabel` is a **shared helper** used by three call sites — `ConversationItem` (tile), `ReplyPreviewBar` (reply input preview), and `MessageBubble` (reply quote) — so fixing it once makes the label consistent everywhere.

## The change

In `messageContentLabel`, the solo-image branch now first checks for a sticker. `hasMultiplePayloads` is already `false` at that point, so this matches the same single-image-payload condition `MediaMessage` uses to render the sticker bubble:

```kotlin
firstPayload.contentType?.startsWith("image/") == true &&
    (firstPayload.descriptorInfo() as? DescriptorContent.ImageFile)?.isSticker == true
```

When true, the label is `chat_preview_sticker` ("Sticker") with an `Icons.AutoMirrored.Filled.StickyNote2` icon. Ordinary images/photos/videos/audio/links/locations are unchanged. Detection reuses the existing `DescriptorContent.ImageFile.isSticker` check — no new detection logic, **no wire-format change**.

New string `chat_preview_sticker` added to `homebase-common` `values/strings.xml` (`Sticker`) and `values-da/strings.xml` (`Klistermærke`). All user-facing text goes through `stringResource`.

## Cross-platform

Compiled clean on all four targets:
- `:homebase-chat:compileKotlinJvm` ✅
- `:homebase-chat:compileAndroidMain` ✅
- `:homebase-chat:compileKotlinIosSimulatorArm64` ✅
- `:homebase-chat:compileKotlinWasmJs` ✅

## Tests

No unit test added: `messageContentLabel` is a `@Composable` whose output is a `stringResource`, so it isn't unit-testable on JVM without a UI test harness. Behavior verified by reusing the already-tested sticker-detection idiom (`MediaMessage`, `PayloadDescriptorImageTest`, `MessageAttachmentBuilderStickerTest`).

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/MessageContentLabel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/MessageContentLabel.kt
@@ -6,9 +6,11 @@ import androidx.compose.material.icons.filled.Description
 import androidx.compose.material.icons.filled.Image
 import androidx.compose.material.icons.filled.LocationOn
 import androidx.compose.material.icons.filled.PhotoLibrary
+import androidx.compose.material.icons.automirrored.filled.StickyNote2
 import androidx.compose.material.icons.filled.PlayArrow
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.vector.ImageVector
+import id.homebase.api.client.drives.files.DescriptorContent
 import id.homebase.api.client.drives.files.PayloadDescriptor
 import id.homebase.chat.services.ChatProtocol
 import id.homebase.resources.MR
@@ -20,6 +22,7 @@ import id.homebase.resources.chat_message_link
 import id.homebase.resources.chat_message_location
 import id.homebase.resources.chat_message_multiple_media
 import id.homebase.resources.chat_message_video
+import id.homebase.resources.chat_preview_sticker
 import org.jetbrains.compose.resources.stringResource
 
 /**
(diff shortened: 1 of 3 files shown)


2002Bishwajeet:
Automated review (sticker tile label) — APPROVE.

Verified against the code on `conversation-tile-sticker-label`:

1. Sticker detection reuses the existing idiom, not a heuristic. The new branch is `firstPayload.contentType?.startsWith("image/") == true && (firstPayload.descriptorInfo() as? DescriptorContent.ImageFile)?.isSticker == true` — byte-for-byte the same check `MediaMessage.kt:96` and `MessageBubbleRaw.kt:249` use to render the sticker bubble. The solo-image requirement is guaranteed by the `hasMultiplePayloads` early-return above it, so by the time this branch runs it is the single-payload case. Good.

2. Ordinary previews unchanged. Non-sticker images parse to `ImageFile(isSticker=false)` (default) and fall through to the unchanged `chat_message_image` branch; video/audio/link/location/file branches are untouched. Note: link/location previews that carry an `image/*` contentType already resolved to "Image" on `main` (the image branch sat above the key checks) — the new sticker branch only diverts `isSticker==true`, so it neither fixes nor worsens that pre-existing ordering. No regression.

3. String present in BOTH locales (`values` -> "Sticker", `values-da` -> "Klistermærke"); those are the only two locale dirs. Rendered via `stringResource(MR.string.chat_preview_sticker)` — no literal Text, Konsist-safe.

4. Label is consistent across the reply preview: the single shared `messageContentLabel(...)` helper feeds all three call sites (ConversationItem tile, ReplyPreviewBar, MessageBubble reply quote), each passing a real PayloadDescriptor.

5. No wire-format change — read-only consumption of the existing `descriptorContent` slot; no serialization touched.

6. Compiles (static verification): `Icons.AutoMirrored.Filled.StickyNote2` is already used in CreateConversationScreen.kt; the `DescriptorContent` import path is correct; `descriptorInfo()` is public and already called cross-module from homebase-chat.

7. The reported preview mechanism is accurate: `lastMessageFirstPayload = msg.payloads?.firstOrNull()` is populated identically in all three paths (ConversationUiModel:164, ConversationMessageBump:86, ConversationStream:517), so the tile preview is computed live from the real descriptor — isSticker is reachable, nothing re-derived from stored text.

Non-blocking nit: the new `automirrored.filled.StickyNote2` import is placed between the `filled.*` imports (alphabetically out of order). CI lint is Android Lint only (no ktlint/detekt import-order rule), so it will not break the build — cosmetic only.

No blocking issues.

PR #672, opened 2026-06-05
Add sticker picker / saved "My Stickers" tray (synced Stickers drive)

2002Bishwajeet:
## What

A v1 **saved-stickers library** for chat — a flat, synced "My Stickers" tray modeled exactly on Vault. Saved stickers are stored as one HomebaseFile each on a **dedicated, synced Stickers drive**, so the library follows the user across Android / iOS / Desktop for free via the existing sync engine. Reuses PR #664's sticker wire format end-to-end (`forceSticker` → `{"isSticker":true}` payload descriptor) — no new protocol, no hand-rolled crypto, no new SQLDelight table.

## How it maps to Vault (the precedent)

| Vault | Stickers |
|---|---|
| `vaultLabeledDrive` | `stickerLabeledDrive` (placeholder GUID) |
| `VaultStream` (StateFlow from DriveMainIndex + EventBus) | `StickerStream` |
| `VaultUploaderService` (outbox + optimistic write) | `StickerService` |
| `HomebaseFile.toVaultEntry()` | `HomebaseFile.toSavedSticker()` |
| Moments on-demand `mountDrive` | `StickerService.ensureDriveMounted()` |

## Scope

**Storage / sync (commonMain):** `StickerProtocol` (`STICKER_FILE_TYPE = 7060`, single `stk_00` payload), tiny `StickerFileContent`, `SavedSticker` + mapper (reserved nullable `groupId` for future packs), `StickerStream` (in-memory `StateFlow<List<SavedSticker>>`, EventBus-driven), `StickerService` (`saveSticker(bytes, contentType)` single entry point, `resolveForSend`, `deleteSticker`, lazy `ensureDriveMounted`).

**Save → pick → send:** `forceSticker` added to `AttachmentPendingFile.FileImage` and threaded into `AttachmentInput.forceSticker` (one line). `StickerHandler` handles: tap a saved sticker → re-stage bytes as a `FileImage(forceSticker=true)` → `addMessageWithFiles` (reuses the whole send pipeline); long-press a received sticker → "Save sticker" (decrypt via `getPayloadBytes` → `saveSticker`); import a transparent PNG/WebP from the tray (**alpha-gate** via `ImageUtils.hasNonOpaquePixels`, opaque rejected with a snackbar); per-tile Remove.

**UI:** `StickerTray` (LazyVerticalGrid + Import tile + long-press Remove) inside a `StickerTraySheet` (ModalBottomSheet + FileKit image picker); new sticker tile in the composer `AttachmentOptions` row; "Save sticker" item in the full-screen media menu **gated strictly to `ImageFile(isSticker = true)`** so it never appears on ordinary photos.

**Wiring:** DI singles + post-auth reset/start; new strings in both `values/` and `values-da/`; `SavedStickerTest` (JVM unit test for the mapper, sticker flag, groupId reservation, IV decode).

**Deferred:** packs/reorder, emoji tagging, animated stickers, search.

## ⚠️ Before shipping (must-fix)

- **The Stickers drive type GUID in `AppConfig.stickerLabeledDrive` is a PLACEHOLDER** (`// TODO: replace with the server-provisioned Stickers drive type GUID`) — the same open item the Vault placeholder GUID has. The server must **provision a Stickers drive type** before mount + upload will work; until then, saving/loading stickers will fail at the drive layer. Alternative lower-friction fallback: store stickers on the existing Chat drive under the sticker fileType (couples sticker storage to chat sync; avoids a new mount/permission).

## Build verification note

This worktree exhibits a **pre-existing, environment-level Gradle/Compose-resources defect**: any *newly added* `homebase-chat/commonMain` file cannot resolve `homebase-common`'s `MR.string.*` accessors — a brand-new control file referencing the long-existing `chat_dice_share` fails identically on a clean `main` tree (caches disabled, daemons killed, fresh JVM, out-of-process compiler). `homebase-common` (with the new `AppConfig` + strings) compiles cleanly, and every error is exclusively a generated-accessor reference (zero errors in the new Kotlin logic). A clean CI checkout that builds the full module graph in one pass resolves this. Please confirm `:homebase-chat:compileKotlinJvm` / `:compileAndroidMain` are green in CI.


2002Bishwajeet:
## Automated review (picker) — saved "My Stickers" tray

**Verdict: REQUEST CHANGES — does not compile (all 4 CI jobs red).**

Scope note: this PR (#672) is the *picker / saved-stickers* half. The background **remover** (expect/actual ML Kit + Vision, transparent-PNG output, jvm/wasm null actuals) is **not in this diff** — that's a separate PR, so the remover items in the brief don't apply here.

### What's done well (verified against `origin/main`)
- **Reuses PR #664 end-to-end, no reinvention.** `forceSticker` is threaded through the *existing* `AttachmentInput.forceSticker` → `MessageAttachmentBuilder` path; the payload marker is the existing `DescriptorContent.descriptorContentFromImage(isSticker = true)` (`{"isSticker":true}`). No hand-rolled detection/crypto. Crypto/outbox is reused (`PayloadBundleEncryptionService`, `KeyHeader.newRandom16`, `OutboxSync`, `OptimisticWriter`, `DriveFileProvider.getPayloadBytesDecrypted`).
- **Picker is a synced drive, not a local DB.** `stickerLabeledDrive` is a real `LabeledDrive`/`TargetDrive`, mounted via `AuthConnectionCoordinator.mountDrive(...)`, hydrated from DriveMainIndex via `QueryBatch` + kept live by `EventBus` — a faithful structural mirror of `VaultStream`/`VaultUploaderService`. Placeholder drive GUID is clearly flagged (TODO in `AppConfig.kt` + the PR body's "before shipping" section).
- **No wire-format break / foundation field is clean.** `STICKER_FILE_TYPE = 7060` is unused on `main` (no collision). `forceSticker: Boolean = false` added to `FileImage` and threaded in one line — backward-compatible.
- **Strings present in BOTH locales.** All 12 keys exist once each in `values/` and `values-da/`; both files are well-formed XML; UI uses `stringResource()`.

### Blocking
- **The PR does not compile.** CI is red on **Build Android & Desktop**, **Build iOS**, **Run unittests**, and **Lint & Format** — all from the same root cause: `homebase-chat:compileKotlinJvm` / `:compileAndroidMain` / `:compileKotlinIosArm64` **FAILED** with `Unresolved reference 'chat_sticker_save_failed'` (and every other new `MR.string.*` / `cd_*` accessor) across `StickerHandler.kt`, `AttachmentOptions.kt`, `Menu.kt`, `StickerTray.kt`.
  - The PR body attributes this to "a pre-existing, environment-level Gradle/Compose-resources defect" that "a clean CI checkout … resolves," and asks the reviewer to "confirm CI is green." **CI *is* that clean checkout, and it reproduces the failure** — so the environment rationalization doesn't hold.
  - I checked the source side: the new strings are well-formed, single (no duplicate keys), correctly placed inside `<resources>`, in both locales, and `homebase-common:generateResourceAccessorsForCommonMain` + `homebase-common:compileKotlinJvm` ran **fresh and green** in CI (so common's `MR` *does* contain the new accessors). Yet `homebase-chat` — which consumes `id.homebase.resources.MR` transitively from `homebase-common`, exactly as the long-existing `chat_dice_share` does — can't see them. The CI log also shows `homebase-chat:compileAndroidMainLibraryResources FROM-CACHE` while common regenerated, which points at a **cross-module build/incremental-cache inconsistency**, not the author's machine. This must be root-caused and the build made green before merge (per CLAUDE.md: capture the cause, don't ship around it). Because compilation dies first, the Konsist `stringResource` gate never actually ran on these strings — "Konsist passes" is unconfirmed, masked by the compile break.

### Non-blocking
- **Two hardcoded user-facing strings** in `StickerHandler.kt`: `ShowErrorMessage("Failed to send sticker: ${e.message}")` (L65) and `ShowErrorMessage("Failed to remove sticker: ...")` (L144). These are in a handler (not a Composable), so Konsist won't flag them, but they're untranslated English inconsistent with every other message here (which use `MR.string.*`). Prefer a `chat_sticker_*_failed` resource.
- **Two dead string keys:** `chat_sticker_import` and `chat_sticker_remove` are defined in both locales but referenced nowhere in `.kt` (the import tile uses `cd_import_sticker`, removal uses `chat_sticker_removed`). Drop them or wire them up.
- **Permission-extension flow not wired (relevant once the real GUID lands).** `getStickerPermissionExtensionConfig()` / `stickerTargetDriveAccessRequest` are added to `AppConfig` but, unlike the Vault precedent (`getVaultPermissionExtensionConfig()` wired into an `ExtendPermissionViewModel` qualifier in `AppModule`), they're referenced nowhere. `mountDrive` only subscribes the WS/registry — it does not grant server-side Read/Write. So even with a server-provisioned drive type, uploads/queries would 403 until this extend-permissions flow is wired and triggered. Worth calling out alongside the GUID TODO (the PR body mentions only the GUID).
- `saveSticker(name)` param is always passed `null` (dead for now) — fine as reserved surface, just noting.

PR #685, opened 2026-06-08
Fix post-login sync: skip ungranted drives, hold loading screen until sync stops

Seifert69:
## Symptom

On WASM (and any platform), logging in sometimes **skips the initial loading/sync screen** and lands straight on the chat list, which then syncs in the foreground.

## Root cause

On login, `AuthConnectionCoordinator` mounts, WebSocket-subscribes, and syncs **every** drive returned by the cross-device `DriveRegistry`. The registry is the user's "activated drives" list and is **not permission-aware** — a drive activated on another device (or before a permission change) can appear there even though *this* app token has no read grant for it. In the captured session that drive was **Vault** (`f47ac10b-58cc-4372-a567-0e02b2c3d479`); the app even logs `Missing permissions detected: drives=1` for it.

The server enforces per-drive read permission on the notify WebSocket. In `AppNotificationHandler.ProcessCommand` (`EstablishConnectionRequest`), a single unauthorized drive throws `OdinSecurityException`, which sends an `AuthenticationError` and then `throw new CloseWebSocketException()` — closing the **entire** socket. Sequence from the WASM console:

```
WS[1] negotiated Sec-WebSocket-Protocol=odin.notify.v1
Sent WebSocket command: establishConnectionRequest
WebSocket closed by server                 ← server rejects the unauthorized (Vault) drive
AuthCC: onDisconnected fired for WS[1]
(LoginViewModel) AuthState: Authenticated  ← navigates Home here, before any sync
... WS[1] reconnects (without Vault), handshake successful, DriveSync starts ...
WebSocket auth error (non-fatal): [Unauthorized to read to drive [f47ac10b…]]
drive f47ac10b… returned 403 Forbidden — unmounting for this session
```

The client's WS layer recovers as designed (exclude the rejected drive, reconnect ~1s later), but that brief disconnect flips `AuthConnectionCoordinator.isConnecting=false`, and `LoginViewModel.observeAuthState` maps `Authenticated + !isConnecting` → `StartupState.Authenticated` → `handleAuthenticatedUser()` → `NavigateToHome`. So the login navigates **before the first sync even starts**. The same un-granted drive also returns `403` on the REST `query-batch` every login.

The fix targets the trigger, not the symptom: never offer the server a drive this token can't read.

## Fix

Resolve the app token's **readable** drives once per `Authenticated` transition and filter un-granted drives out of both the mount loop and the WebSocket subscription. The grant source is `SecurityContextProvider.getSecurityContext()` (`/auth/context`) — the same source `ExtendPermissionViewModel` already uses to compute missing permissions.

- **`refreshGrantedDriveAliases()`** — caches the normalized (lowercased, hyphen-stripped) aliases of drives that carry a `Read` grant. On fetch failure it leaves the filter **disabled** (`null`), so a transient `/auth/context` error degrades to the previous behaviour rather than hiding a granted drive.
- **`List<LabeledDrive>.retainGrantedDrives()`** — drops un-granted optional drives. Applied in the login **mount loop** (stops the REST 403 churn) **and inside `connect()`** (stops the drive ever reaching the WS subscription — covers mid-session reconnects and the headless `promoteToForeground` replay). Mandatory drives (Chat, Contacts) are never filtered.
- The registry-change **observer baseline stays the full registry**, so a drive we intentionally skipped isn't later mistaken for a brand-new mount.
- **DI:** inject `SecurityContextProvider` into `AuthConnectionCoordinator`.

UUID comparison reuses the same normalization as `compareStringUuId` so aliases match regardless of hyphen/case format.

## Why not patch the login navigation instead

Gating `LoginViewModel` on a "real connect" signal would hide the premature navigation but leave the client subscribing to a drive it knows it can't read — re-rejected and re-flapped on every login, plus a 403 per login. Removing the unauthorized subscription fixes the WS close, the 403 churn, and the skipped loading screen at once.

## Testing

- Compiles on **JVM**, **Wasm**, and **Android** against current `main`. iOS unchanged (not compiled — no macOS host here).
- Runtime verification pending: rebuild/restart the WASM run and log in; expect no `Unauthorized to read to drive`, no Vault `query-batch` 403, `WS[1] handshake successful` on the **first** attempt, and the loading/sync screen shown before navigating Home.

### DIFF
--- a/homebase-auth/src/commonMain/kotlin/id/homebase/auth/login/LoginViewModel.kt
+++ b/homebase-auth/src/commonMain/kotlin/id/homebase/auth/login/LoginViewModel.kt
@@ -8,8 +8,10 @@ import id.homebase.api.exception.AuthInProgressException
 import id.homebase.api.isIos
 import id.homebase.api.sync.DriveState
 import id.homebase.api.sync.DriveSyncManager
+import id.homebase.api.sync.SyncState
 import id.homebase.api.youauth.UsernameStorage
 import id.homebase.api.youauth.YouAuthFlowManager
+import id.homebase.api.youauth.YouAuthState
 import id.homebase.core.auth.AuthConnectionCoordinator
 import id.homebase.core.config.AUTO_CONNECTIONS_CIRCLE_ID
 import id.homebase.core.config.AppConfig
@@ -18,8 +20,6 @@ import id.homebase.core.config.appPermissions
 import id.homebase.core.config.circleDriveTargetRequest
 import id.homebase.core.config.targetDriveAccessRequest
 import id.homebase.core.notifications.NotificationService
-import id.homebase.core.util.StartupState
-import id.homebase.core.util.mapToStartupState
 import id.homebase.resources.MR
 import id.homebase.resources.error_unknown
 import id.homebase.resources.login_error_generic
@@ -238,10 +238,11 @@ class LoginViewModel(
         didHandleAuthenticated = false
         authStateJob = viewModelScope.launch {
             combine(
+                youAuthFlowManager.authState,
                 authConnectionCoordinator.connectionState,
-                youAuthFlowManager.authState
-            ) { connectionState, authState ->
-                authState.mapToStartupState(connectionState.isConnecting)
+                driveSyncManager.syncState
+            ) { authState, connectionState, syncState ->
+                Triple(authState, connectionState.isConnecting, syncState)
             }
                 .distinctUntilChanged() // Ensures only unique combined results are emitted
                 .catch { error ->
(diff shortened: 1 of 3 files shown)


Seifert69:
Added a second, independent fix on this branch (commit `655f964`): **hold the initial loading screen until the drive sync actually stops.**

### Why
Verified live on web (`michael.seifert.page`, all drives granted, so the drive-grant fix above was clean — no 403/WS-close): the loading screen still vanished ~85s early. Trace:
- navigated Home at `12:26:42.611`
- `DriveSyncManager.syncState` only reached `Completed` at `12:28:08.306`

Root cause (independent of drive grants): navigation was gated on `isConnecting`, which flips false in `AuthConnectionCoordinator.onConnected`'s `finally` as soon as the **first** `syncAll()` pass returns. A drive-mount-triggered WS refresh (`WS[1]→WS[2]`) makes that first pass race through near-empty query-batches and return in ~0.4s.

### Fix
`LoginViewModel.observeAuthState` now gates the login→home transition on the drive sync's terminal state (`DriveSyncManager.syncState`) while `Authenticated`:
- `Completed` or `Failed` → leave the loading screen (either outcome — a premature churn like a network drop still yields a terminal state; the running client resumes sync on WS reconnect).
- `Idle && !isConnecting` → sync can't start → skip the loading screen.
- otherwise (`Syncing`/connecting) → keep showing it.

`syncState` is computed over the live mounted drive set, so it waits for exactly the drives that actually sync (varies per session); 403-denied drives are unmounted and excluded so it still reaches `Completed`. Pure commonMain → identical on all platforms.

Verified on web after the fix: loading screen now holds through `syncState → Completed`. Compiles on JVM/Wasm/Android; auth tests pass.

(Happy to split this into its own PR if you'd prefer to keep #685 scoped strictly to the drive-grant change — they're independent mechanisms.)

PR #686, opened 2026-06-08
Honest upload % via payload-relative socket send-buffer cap

Seifert69:
## Problem

Sending a photo in chat shows the upload % race to 100% almost instantly, then a **"Finalizing…" spinner that hangs 10–50 s**. Diagnosed from `homebase.log`: every slow upload (8–51 s) is a media send (`payloads ≥ 1`); every text send (`payloads = 0`) is < 1 s.

**Root cause:** Ktor *streams* the multipart body (a large video shows a wire-paced % climb), but a curated image fits entirely inside the OS TCP send buffer (Android `tcp_wmem` autotunes to several MB). Ktor's `onUpload` counts bytes *accepted into the socket*, not bytes on the wire — so for a buffer-sized payload it reports 100% the instant the kernel swallows it, and the real transmission happens during "Finalizing". (If it were server-side, the video would hang at "Finalizing" too — it doesn't.)

## Fix

Cap `SO_SNDBUF` **below the payload size** so TCP backpressure forces `onUpload` to track the wire. The cap is relative to payload size so large uploads aren't over-throttled. **Only the multipart drive-upload path** uses the capped client; every other request keeps the shared client untouched. No compression (payloads are curated), and "Finalizing" is kept (it's genuinely one step — awaiting the POST response).

### What changed
- **`UploadHttpClientPool`** — one cached `HttpClient` per power-of-two tier `{64K…2M}`; `buffer = (totalUploadSize / 12).coerceIn(64 KiB, 2 MiB)`. `/12` → ~12 wire-paced steps; the 2 MiB ceiling stays above the bandwidth-delay product of typical mobile/Wi-Fi links; the 64 KiB floor bounds the penalty for tiny payloads. Tiering (vs. per-request sizing) is required because the OkHttp `SocketFactory` is fixed at client-build time and runs on OkHttp's own threads.
- **`createUploadHttpClient` (expect/actual)** — Android (OkHttp) installs a delegating `SocketFactory` that calls `setSendBufferSize` **pre-connect** (required to influence TCP window scaling; also disables Linux autotuning). Darwin/CIO/Js are **honest no-ops** (no `SO_SNDBUF` lever).
- **`HttpClientProvider`** — extracted `applyOdinDefaults()` so upload clients are plugin-identical to the shared one.
- **`OdinApiProviderBase`** — `plainPostMultipart`/`plainPatchMultipart` take an optional `client` (defaults to the shared client → non-upload callers unaffected).
- **`DriveUploadProvider`** — threads the already-computed `totalUploadSize` into `pureUpload`/`pureUpdate` to select the tier. **Fixes `updateFileByUniqueId`**, which previously neither computed the size nor reported progress against the true total. Adds a `(UploadTiming)` log line (`bytes/bufferBytes/handoffMs/waitMs/throughputKBps`).
- **`ApiModule`** — `single { UploadHttpClientPool() }`.

## Platform support (honest)
| Platform | Engine | Status |
|---|---|---|
| **Android** | OkHttp | ✅ real `SO_SNDBUF` cap |
| iOS | Darwin/NSURLSession | ⚪ no-op — NSURLSession exposes no send-buffer control; `didSendBodyData` may already be more wire-accurate. **Measure first** via the timing log. |
| Desktop | CIO | ⚪ no-op — no supported lever |
| Web | Js/fetch | ⚪ no-op — fetch can't report upload progress (would need XHR) |

## Testing
- Compiles on JVM, Android, wasmJs (iOS skipped on a Linux host).
- `homebase-api` + `homebase-chat` JVM test suites pass.

### How to verify on-device (Android)
Send a photo, then:
```
adb shell run-as id.homebase.feed.dev cat files/logs/homebase.log | grep "(UploadTiming)"
```
Before: `handoffMs` tiny, `waitMs` huge. After: `handoffMs` ≈ real transfer time, `waitMs` small. Throttle to ~1 Mbps and the chat % should climb in steps instead of snapping to 100% then hanging.

## Regression notes
- The cap trades peak throughput for honesty; the 2 MiB max tier bounds this — raise it if fast-link upload times regress.
- Cross-tier sends reopen TCP+TLS (sequential outbox, acceptable); the shared client's connection reuse is unchanged.

### DIFF
--- a/homebase-api/src/androidMain/kotlin/id/homebase/api/client/UploadHttpClientFactory.android.kt
+++ b/homebase-api/src/androidMain/kotlin/id/homebase/api/client/UploadHttpClientFactory.android.kt
@@ -1,0 +1,53 @@
+package id.homebase.api.client
+
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.okhttp.OkHttp
+import java.net.InetAddress
+import java.net.Socket
+import javax.net.SocketFactory
+
+/**
+ * Delegating [SocketFactory] that caps each socket's `SO_SNDBUF` to [sendBufferBytes].
+ *
+ * OkHttp opens connections via the no-arg [createSocket] (an *unconnected* socket it then
+ * connects), so setting the send buffer there applies **pre-connect** — required to influence
+ * TCP window-scaling negotiation, and it disables Linux send-buffer autotuning for that socket
+ * (which is exactly what forces the backpressure that makes Ktor's `onUpload` track the wire).
+ * The connected overloads are delegated and set defensively even though OkHttp doesn't use them.
+ *
+ * `setSendBufferSize` is a hint the kernel clamps to `net.core.wmem_max` (>= 4 MiB on Android),
+ * so the [UploadHttpClientPool] range (64 KiB..2 MiB) is honored.
+ */
+private class SndBufSocketFactory(
+    private val sendBufferBytes: Int,
+    private val delegate: SocketFactory = SocketFactory.getDefault(),
+) : SocketFactory() {
+
+    private fun Socket.capped(): Socket = apply {
+        runCatching { sendBufferSize = sendBufferBytes }
+    }
+
+    override fun createSocket(): Socket = delegate.createSocket().capped()
+
+    override fun createSocket(host: String?, port: Int): Socket =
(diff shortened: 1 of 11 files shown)


Seifert69:
It is almost guaranteed a server side issue

PR #712, opened 2026-06-12
CI: fit the iOS release link into the 7 GB macos-15 runner

Seifert69:
## Problem

Run [27400310875](https://github.com/homebase-id/chat-kmp/actions/runs/27400310875) failed in `build-ios` with `java.lang.OutOfMemoryError: Java heap space` inside `:homebase-core:linkReleaseFrameworkIosArm64` after ~44 minutes. Even the *passing* run the day before spent ~66 minutes in the Kotlin framework phase — the build has been GC-thrashing at the ceiling of the free 7 GB `macos-15` runner for a while.

Three things stack up:

1. **Two 8 GB daemons on a 7 GB box** — `org.gradle.jvmargs` and `kotlin.daemon.jvmargs` are both `-Xmx8g`, while `xcodebuild`/`SWBBuildService` are also resident. The Gradle daemon only orchestrates this build.
2. **Concurrent K/N tasks in one heap** — `kotlin.native.gradle.workers.max=3`, and the archive builds two frameworks (`homebase-core` + `homebase-notifshared`).
3. **The link inherits a dirty heap** — with the compiler daemon enabled, the release link runs in the same long-lived JVM that just compiled every module in the project.

## Change

A single CI step in the `build-ios` job appends overrides to `gradle.properties` at build time (last key wins) — **repo defaults are untouched for local dev**:

- `org.gradle.jvmargs=-Xmx2g`, `kotlin.daemon.jvmargs=-Xmx3g` — daemons stop competing with the K/N compiler
- `kotlin.native.gradle.workers.max=1`, `kotlin.incremental.native=false` — serialize K/N work, drop daemon-resident caches that are dead weight on a one-shot runner
- `kotlin.native.disableCompilerDaemon=true` + `kotlin.native.jvmArgs=-Xmx6g` — each K/N compilation gets a fresh short-lived process, so the release link starts with a clean 6 GB heap

Plus a failure-only `ios-jvm-oom-diagnostics` artifact (heap dumps + GC logs, `-XX:+HeapDumpOnOutOfMemoryError`), so if it still OOMs we learn what the live set actually is instead of guessing.

The override step runs **after** the Kotlin/Native cache step on purpose — that cache key hashes `gradle.properties` and should see the pristine file.

## Rollback

Delete the `Apply iOS CI memory overrides` and `Upload JVM OOM diagnostics` steps (or revert this one-file commit). Nothing outside the workflow file changes.

## Validation

`Build Mobile Release Production` dispatched on this branch (branch content = main + this CI-only change).


Seifert69:
**Attempt 1 (run [27407420160](https://github.com/homebase-id/chat-kmp/actions/runs/27407420160)) failed — but produced the evidence we wanted.**

From the `ios-jvm-oom-diagnostics` artifact (7 GB heap dump + GC logs from 9 separate K/N compiler processes):

- The 6 GB cap was honored, and at death a maximal full compaction clearing soft references reclaimed **nothing**: `5122M->5122M(6144M)`. ~5.1 GB is genuinely live during `linkReleaseFrameworkIosArm64`, including 477 humongous regions (~1.9 GB of large arrays). Not config failure — real heap demand.
- `User=10.61s Sys=31.84s Real=44.68s` for a single full GC: the runner was deep in swap, which also explains the SIGKILLed (exit 137) sibling java processes.
- Upside: fresh-process compilation cut the Kotlin compile phase from ~40 min to **~3 min**.

**Attempt 2 (`fa1409e1`):** raise the K/N compiler process to `-Xmx8g` (the heap size yesterday's green run completed with), and pre-build both release frameworks in a dedicated step before fastlane, so the link doesn't share the 7 GB box with xcodebuild. The Xcode script phase then finds the link UP-TO-DATE and only copies.

If 8 GB + a quiet box still isn't enough, the next lever is shrinking the ObjC export surface (`export()` of homebase-api/common/image-editor-ui in the ComposeApp framework) — the heap dump is saved locally for a class histogram if we need to see what the 5.1 GB actually is.

PR #723, opened 2026-06-12
feat(ios): shake recovery v2 — recreate the Compose surface (root cause: first frame rendered while backgrounded)

Seifert69:
## 🎯 Root cause found (3 field captures, 2 devices, identical fingerprint)

Every confirmed blank onset shares one signature:

| occurrence | pattern | cache at first Foreground |
|---|---|---|
| 06-08 11:55 | ColdStart 10:30, **no immediate Foreground** → first Foreground 85 min later, no `bgMs` | **6889 / 4** |
| 06-10 18:18 | ColdStart 18:16, **no immediate Foreground** → push → first Foreground, no `bgMs` | **6889 / 4** |
| 06-12 20:47 | ColdStart 20:28, **no immediate Foreground** → first Foreground 19 min later, no `bgMs` | **6889 / 4** |
| healthy cold starts | ColdStart → Foreground within ~50 ms (user-tap launch) | 0 / 0 |

**Mechanism:** iOS launches the process into the **background** (prewarm / push relaunch after multitasking memory pressure). The SwiftUI scene connects and `MainViewController()` runs (`promoteToForeground` in the logs proves it) — so **Compose renders its first frame while `applicationState == .background`**. iOS rejects background Metal submissions (the CMP-9488 error family), so the glyph-atlas texture created by that frame is **born dead**: CPU strikes rasterize fine (the constant `6889/4` = the same first composition every time), Skia's bookkeeping marks them resident, but the GPU memory never receives pixels. **Every later glyph — old or new — lands in the poisoned atlas** and samples nothing. Frames/images don't touch the glyph atlas → they render.

**Why shake v1 (purge+recompose) couldn't work — falsified 3× in the field:** after `Graphics.purgeAllCaches()` + full recomposition, fresh strikes rebuilt within ~1.3 s and the screen stayed blank. The poisoned atlas lives on the per-context `GrDirectContext` Compose owns internally; no public API reaches it.

## What this PR does

**Shake = recreate the Compose surface.** `ContentView` bumps `.id()` on `ComposeView` → SwiftUI dismantles + remakes the `ComposeUIViewController` → fresh `GrDirectContext` → empty atlas bookkeeping → every glyph re-rasterizes and re-uploads (in the foreground this time). This replicates the one known-good fix (app relaunch) without losing the process — Koin/DB/sync survive.

- **Acknowledge toast** (native SwiftUI — renders even while Compose is blank): testers now *know* the shake fired, and it reminds them to share the log without clearing it.
- **Before/after logging** to `homebase.log` (`shake-recreate/before`, `shake-recreate/after-recreate`), each with the CAMetalLayer state of the old/new view.
- **Debounced 5 s** — field logs show double-shakes; we must not recreate twice.
- `onBlankTextShake` (v1) removed; `TextRecovery` kept as a generic lever with corrected docs.

**Cost:** a shake rebuilds the composition — navigation resets, in-progress typing lost. Acceptable on a deliberate shake during an unusable screen.

## What this PR intentionally does NOT do (next steps, pending this experiment)

- **Trigger v2:** #703's `peak > 200 KB` gate excludes exactly this onset (first-ever foreground — nothing rendered before). The data hands us the fix: fire on first Foreground with no `bgMs` AND `0 < fontCacheUsed ≤ ~50 KB` (3-for-3 in our data, 0 false positives — healthy first foregrounds show exactly 0). Lands once the recovery is proven.
- **Prevention:** gate `ComposeView` on the scene's first `.active` phase so Compose never renders while backgrounded — would stop the atlas being poisoned at all.
- **Repro recipe:** kill app → silent push (`content-available`) relaunches it in background → wait → open. Possibly deterministic via `xcrun simctl push`.

## Verification

- Common: `:homebase-common:jvmTest` + JVM/metadata compiles green.
- `nativeMain` (`logShakeRecovery`) + Swift (`.id()` recreation, toast, debounce) validated by **CI (iosArm64 / Xcode build)**.

### DIFF
--- a/BLANK_TEXT_INVESTIGATION.md
+++ b/BLANK_TEXT_INVESTIGATION.md
@@ -42,9 +42,13 @@ anything rendered. Trigger v2 should fire on the fingerprint instead: first Fore
 first frame was drawn in background). 3-for-3 in our data; 0 false positives observed (healthy
 first foregrounds show exactly 0).
 
-**Prevention idea (untested):** never let Compose render while backgrounded — gate `ComposeView`
-on the scene's first `.active` phase (show a native placeholder until then). Would stop the atlas
-being poisoned at all.
+**Prevention — BUILT (same branch as shake v2):** `ContentView` defers `ComposeView` creation until
+the scene's first `.active` phase (native `systemBackground` placeholder until then; latched forever
+after so later backgrounding never tears it down). Compose can no longer render its first frame
+while the app is backgrounded → the atlas is never poisoned. Logged as `prevention: first .active —
+ComposeView built now` in homebase.log. **Validation signal:** background-launched sessions
+(ColdStart with no immediate Foreground) should now show `0/0` at the first Foreground instead of
+the `6889/4` fingerprint — measurable even without a user report. Shake v2 stays as the backstop.
 
 **Repro recipe (untested):** kill app → deliver a silent push (`content-available`) so iOS
 relaunches it in background → wait → open the app. Simulator: `xcrun simctl push` with a
--- a/BLANK_TEXT_INVESTIGATION.md
+++ b/BLANK_TEXT_INVESTIGATION.md
@@ -5,6 +5,56 @@ Skia-backed targets (iOS and Web). Update as we learn more.
 
 ---
 
+## 🎯 iOS ROOT CAUSE FOUND (2026-06-12) — first frame rendered while app BACKGROUNDED
+
+Three field captures (2026-06-08, 06-10, 06-12 — two devices, two builds) share one exact onset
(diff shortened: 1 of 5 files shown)


Seifert69:
Added the **prevention** to this PR (commit aa17aa9e): `ContentView` defers `ComposeView` creation until the scene's first `.active` phase, so Compose can never render its first frame while the app is backgrounded — the confirmed atlas-poisoning condition. Latched in both the `.active` onChange and `onAppear` so later backgrounding never tears the view down. Validation signal even without a user report: background-launched sessions should now show `0/0` at first Foreground instead of the `6889/4` fingerprint. Shake v2 remains as the backstop.

PR #739, opened 2026-06-16
Global crash handler with native recovery screens (Android/iOS/Desktop)

2002Bishwajeet:
## Global crash handler — native recovery on Android, iOS & Desktop

A safety net for otherwise-fatal uncaught exceptions. Every crash is captured to a readable on-disk report (independent of Crashlytics, which has been unreliable at recording these), and the user gets a native, brand-themed recovery screen instead of a silent/hard crash.

### Per platform
- **Android** — `GlobalCrashHandler` (installed early in `Application.onCreate`) writes the report, records a Crashlytics non-fatal, launches a separate `:crash`-process **Material 3 `CrashActivity`** (Homebase brand theme, light/dark), then kills the crashed process so the OS "app keeps stopping" dialog never appears. **Restart** + **Share crash report** + collapsible details. Crash-loop guard.
- **iOS** — process death is unavoidable, so recovery is **next-launch**, shown **only when the app is stuck in a startup crash loop** (≥2 consecutive startup failures, one silent retry). A one-off / mid-session crash is captured + shareable via Help with **no nag** on reopen. Native SwiftUI screen (brand colors, scrollable report); hooks armed first in `initializeApp`; heavy init deferred while recovery shows (avoids the init-crash loop).
- **Desktop** — native **Swing** recovery dialog (brand colors, rounded buttons): Restart / Reveal report / Quit.

### Shared core
`homebase-common/commonMain` `CrashReporting`: dependency-free (no Koin/coroutines/DB), safe to call mid-crash. Writes `exception + stack/cause + thread + app/device/OS/version + tail of homebase.log` to `<appData>/logs/crash/`. Unit-tested (incl. the startup-crash-loop detection).

### Preserved
Crashlytics where it works; the #737 transient-network carve-out on all platforms; the existing `homebase.log` pipeline.

### Known gap (documented)
iOS crashes **before** `application(_:didFinishLaunching)` (dyld load/link, ObjC/C++ static initializers) can't be caught by any runtime handler — a deployment-target / weak-linking matter, out of scope here.

### Verification
- JVM unit tests + Konsist, `androidApp:assembleDebug`, Desktop compile, iOS `xcodebuild` — all green.
- Runtime-verified: iPhone 17 sim (startup crash loop → recovery; mid-session → none; Continue + Share sheet), Redmi Note 5 Pro (`CrashActivity` after a real crash), Desktop Swing dialog on macOS.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -149,6 +149,7 @@ dependencies {
     implementation(libs.kotlinx.coroutines.android)
     implementation(libs.jetbrains.compose.material3)
     implementation(libs.jetbrains.compose.material.icons.extended)
+    implementation(libs.android.material)
 
     implementation(libs.koin.core)
     implementation(libs.koin.android)
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -159,6 +159,14 @@
             </intent-filter>
         </receiver>
 
+        <activity
+            android:name=".crash.CrashActivity"
+            android:process=":crash"
+            android:exported="false"
+            android:excludeFromRecents="true"
+            android:launchMode="singleTask"
+            android:theme="@style/Theme.Homebase.Crash" />
+
         <provider
             android:name="androidx.core.content.FileProvider"
             android:authorities="${applicationId}.fileprovider"
--- a/androidApp/src/main/kotlin/id/homebase/feed/MainApplication.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/MainApplication.kt
@@ -23,11 +23,9 @@ import id.homebase.api.sync.database.DatabaseManager
 import id.homebase.core.di.allModules
 import id.homebase.core.diagnostics.MainThreadWatchdog
 import id.homebase.core.location.tracking.LocationTrackingCoordinator
-import id.homebase.api.client.isTransientNetworkFailure
-import id.homebase.core.logging.CrashLogger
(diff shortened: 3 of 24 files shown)


2002Bishwajeet:
### Follow-up idea (not in this PR): out-of-band crash-report retrieval on iOS

On iOS the crash report is written to the app sandbox (`Documents/logs/crash/crash-*.txt`), but **retrieval still depends on the app launching + the in-app Share UI**. For the truly-wedged cases this handler is meant for — uncatchable pre-`didFinishLaunching` crashes, or an app that never trips the recovery threshold — the file is correctly written but unreachable by the user.

We could close that gap by exposing the reports via the **Files app** with two `Info.plist` keys:

```xml
<key>UIFileSharingEnabled</key><true/>
<key>LSSupportsOpeningDocumentsInPlace</key><true/>
```

Then `On My iPhone → Homebase → logs/crash/` is browsable/copyable even if the app can't launch at all.

Notes for when we revisit:
- The encrypted DB lives in `${NSHomeDirectory()}/databases/` (`DatabaseDriverFactory.native.kt`), **not** in Documents — so this would expose only logs + crash reports, not the vault DB.
- It exposes the *entire* Documents dir (iOS can't share a single subfolder), so audit that only `logs/` lands there first.
- Low App Review risk (standard keys), but logs can contain PII — a product/privacy call for a vault app. Keep the existing in-app Share as primary; Files-app visibility as the fallback.
- Desktop already has this for free (report sits in a user-visible app-data dir); Android would need a different mechanism (public Downloads/MediaStore).

PR #748, opened 2026-06-17
Location History: inset map in a Card so it clears day-nav and stats

Seifert69:
## Summary

Follow-up polish to the Location add-on (after #746). On the **History** screen the
trace map filled all the space between the day-navigation row and the stats row as a
full-bleed `Box(Modifier.weight(1f).fillMaxWidth())`, running flush into — and visually
covering — the controls above and the stats below (especially with map tiles on).

Wrap the canvas in an inset `Card` (`padding(horizontal = 16.dp, vertical = 8.dp)`) —
the same treatment applied to the Find-device screen in #746 — so the map is a clipped,
rounded, clearly-separated panel. `weight(1f)` is retained, so it still occupies the
main area and pan/zoom keeps working within the Card bounds. Single-file change; no
string/state changes.

## Test plan
- [x] `:homebase-core` compiles on JVM, Android, wasmJs
- [x] `homebase-core:jvmTest` green (incl. Konsist `ArchitectureTest`)
- [ ] Desktop: Location → History shows the map as an inset rounded panel with gaps above the day-nav and below the stats; day switching, date picker, and pan/zoom still work


Seifert69:
Superseded by #749 — the time-warped History playback PR re-does this same trace-canvas block and includes the identical inset Card fix (and the dwell-dot/scrubber/playback work on top). Closing in favor of #749.

PR #756, opened 2026-06-18
Location History: replay the sweep on rewind-to-start

Seifert69:
## Summary

Small follow-up to the animated History playback. Today it auto-plays once per day and
then the 24h scrubber is purely manual. This makes the **rewind-to-start gesture replay**:
drag the scrubber all the way to the left and release, and the warped sweep plays again.

- `onValueChangeFinished`: if the slider is released within 2% of the left edge
  (`REWIND_REPLAY_FRACTION`), clear the one-shot `autoPlayed` guard and bump a `replayToken`.
- `replayToken` is now part of the auto-play `LaunchedEffect` key, so the sweep re-runs
  from the first fix. Releasing anywhere else still just rests the clock at that instant
  (unchanged), and the normal once-per-day auto-play is unchanged.

## Test plan
- [x] `:homebase-core` compiles on JVM, Android, wasmJs
- [x] `homebase-core:jvmTest` green (incl. `LocationDayPlaybackTest` + Konsist)
- [ ] Desktop: History → drag scrubber fully left, release → animation replays from the
  start; release mid-track → rests there; switch days → auto-plays once as before

### DIFF
--- a/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/history/LocationHistoryScreen.kt
+++ b/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/history/LocationHistoryScreen.kt
@@ -33,6 +33,7 @@ import androidx.compose.material3.rememberDatePickerState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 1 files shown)


Seifert69:
Superseded by the reusable-playback PR — the rewind-to-start replay now lives inside the new DayPlaybackMap component (which History uses), so this standalone change is folded in there. Closing in favor of that PR.

PR #771, opened 2026-06-18
Contact subsystem: consolidate writer + model, fix "None" detail and no-op field clear

toddmitchell:
## What & why

The contact subsystem had **two parallel stacks over the same Contacts drive** (chat's `DriveContactService` and the contact-book's `ContactsProvider`/`ContactBookStream`), each with its own write path, model family, and parser. That duplication was actively causing bugs — drifting models behind the "Contact details: None" symptom, and two different write mechanics (direct-drive overwrite vs V2 server-merge) racing on the same files.

This PR consolidates the writer and the model, and fixes two user-facing contact defects. Full design + backend-confirmed semantics live in [`CONTACT_STACK_CONSOLIDATION.md`](CONTACT_STACK_CONSOLIDATION.md).

## Changes

**Phase 0 — safety net**
- `ContactBookEntryParseTest` pins `HomebaseFile.toContactBookEntry()` (incl. the synced/displayName-only "None" case, name fallbacks, image-payload detection, null/invalid returns).
- `ContactNameExtensionsTest` pins the shared name helpers.

**Phase 1 — one writer**
- Connection accept/finalize now writes contacts via the V2 controller (`ContactsProvider.syncContact`) instead of `DriveContactService`'s direct-drive write. Single writer, single (server) merge.
- Deleted the dead write path in `DriveContactService` (`saveContact` + `saveContactForOdinId`), its 3 write-only deps, and 21 unused imports.
- Backend confirmed the V2 `sync` 202 is **synchronous** (write + WS push done before it returns) and, since accept happens while connected, enriches the **full** name + fields — so this is a strict upgrade with no timing regression.

**Phase 2 — one model + one name resolution**
- Deleted the entire chat duplicate model family (`ContactServerFile`, `chat.ContactName` + `chat.{ContactPhone,ContactEmail,ContactLocation,ContactBirthday,ContactImage}`); both read paths now parse the `homebase-api` `ContactContent` family.
- Single name resolution in `homebase-api` — `ContactName.resolveDisplayName(odinId, phone, email)` + `ContactName.initials()` — used by both `toContactBookEntry()` and `DriveContactService.mapToContact()`, so they can no longer drift (the root of the "None" symptom).

**Defects (independent of the migration)**
- **§7 "Contact details: None":** the detail screen's fields section now lists the **Homebase ID**, so an identity-only contact no longer reads "None".
- **#6 no-clear merge:** the V2 merge can't blank a field (empty = leave existing). Client mitigation in `ContactSaveHelper`: the optimistic entry coalesces (keeps the old value for a blanked field, matching the server, so it no longer flashes empty then reappears), and the edit surfaces a snackbar — *"Clearing a contact field isn't supported yet…"* — in both the list and detail screens.

## Deferred

**Phase 3 (merge the two stateful read services into one backing store)** — once Phases 1–2 removed the dangerous duplication, the remaining work (one `QueryBatch` + one EventBus observer) is low-value / high-runtime-risk: it touches ~10 chat people-picker consumers and must preserve divergent semantics (optimistic updates + resurrection guard vs the odinId index + paged fetch). Left as optional later cleanup that warrants on-device validation. Rationale documented in the plan.

## Follow-ups (server-side)
- **#6 real fix:** add a server-supported way to clear a contact field (explicit sentinel or dedicated clear op) — the client can't blank a field through the current merge.

## Testing
- `homebase-api`, `homebase-chat`, `homebase-core`, `homebase-common` (incl. Konsist `ArchitectureTest`) jvmTest suites pass; all targets compile.
- Phase 1's connectionChanged/circleChanged push path was verified on-device earlier (separate branch, already merged).


toddmitchell:
Superseded by #776. That branch pivots from consolidating within the old structure to migrating the contact book onto the new `ContactRepository` (server-shaped single source of truth, now merged to main), and re-applies this PR's durable pieces on the current-main base: the §7 "Contact details: None" → Homebase ID fix, the #6 no-clear merge mitigation, and the shared `resolveDisplayName`/`initials` helpers. The stale plumbing here (it predated #775's device-import removal) is obsolete. Closing in favor of #776.

PR #778, opened 2026-06-20
Live Relay: live GPS sharing (debug-flow build)

Seifert69:
## Summary

Client side of odin-core PR #1572 (**Live Relay** — ephemeral, app-agnostic live-data streaming among connected identities). Wires the **live-GPS** use case into the existing in-conversation **"share location"** action, with send/receive logging so we can confirm the data flows end to end **before** building real UX (tag `LiveRelay`).

📄 **Implementation plan:** [`docs/live-relay-live-gps-debug-plan.md`](docs/live-relay-live-gps-debug-plan.md)

## What's here

- **homebase-api**: `LiveRelayContract` (fixed channel key + `LiveLocationPoint` codec), `LiveShareRoster` (recipient roster, see below), `LiveRelayProvider` (`POST /api/v2/live-relay`), `liveRelay` notification type + `LiveRelayReceivedNotification`, dispatch case emitting `BackendEvent.LiveRelayReceived`.
- **homebase-chat**: `LiveLocationShareService` — relays the latest GPS fix to live recipients, fired from the `LocationPointStore.onPointsBuffered` sink seam (**background / cold-wake capable, not a UI Flow**), throttled to ≥3s, roster persisted for cold-process wakes. Toggled from `ConversationContent`.
- **homebase-core**: extend `onPointsBuffered` to drive the sender; `LiveLocationDebugLogger` decodes inbound blobs and logs them; both registered + lifecycle-managed in `AppModule`.
- **Tests**: `LiveRelayContractTest` + `LiveShareRosterTest`.

## Recipient roster — `{identity, end-time-utc-ms}` pairs

Recipients are kept as `TimedRecipient(odinId, endTimeMs)` rather than a flat on/off list, so concurrent shares are correct:
- **same recipient in two requests** → one entry, latest end-time wins;
- **overlapping shares** → union of recipients (each sent to once per tick);
- **auto-expiry** → a recipient drops off once their window passes; no manual stop needed.

End-times are sender-side bookkeeping only — **no wire change**; the relay stays ephemeral/last-value-wins.

## Notes

- Wire serialization confirmed against odin-core: `ClientNotificationType` is serialized by **camelCase name**, so `LiveRelay` → `"liveRelay"` (the `6001` C# value is never on the wire).
- **All new code is `commonMain`** — no platform actuals. Verified locally: JVM + Android + wasmJs compile, `homebase-api:jvmTest` green. CI proves the **iOS** target (`linkDebugFrameworkIosArm64`) + sim tests.
- Depends on odin-core #1572 being deployed for the live manual test (two connected identities, watch `homebase.log | grep LiveRelay`).


Seifert69:
## ✅ Real-device end-to-end verification (2026-06-20) — both halves confirmed

Validated with a temporary chat trigger (since removed), **frodo (Android, real device) → sam (Desktop App)**, already connected. Logs filtered on tag `LiveRelay`.

**Send (frodo, Android)**
```
START +1 entries=1 uniqueLive=1 until=…
SEND ch=7a1e9c40-… n=1 bytes=172 -> 204    (one per captured GPS point, zero relay-failed)
```
- Background cadence is OS-throttled as designed — sends track GPS deliveries 1:1, multi-minute gaps when the backgrounded/stationary device emitted no points (the ≥3 s gate was never the limiter).
- **Auto-expiry proven:** a GPS point buffered *after* the share's `until` produced **no** `SEND` — the expired roster entry was pruned and the relay stopped.

**Receive (sam, Desktop)**
```
RECV from=frodo… ch=7a1e9c40-… bytes=172 receivedAt=…
RECV-DECODED from=frodo… lat=55.84… lon=12.57… ageMs=401 tracked=1
```
- **Wire serialization confirmed live** — the `liveRelay` case fired, so the server really sends camelCase `"liveRelay"` and our enum member matched.
- Blob decodes cleanly across Android→Desktop (no `RECV-DECODE-FAIL`); position landed in `LiveLocationReceiveStore` (`tracked=1`).
- **End-to-end latency ~400 ms** (`ageMs=401`) across all three relay hops.

### Before merge
- The temporary chat trigger was **removed** — `ConversationContent` matches `main` again, so the half-built feature isn't user-reachable. The plumbing stays fully wired in DI; activate from code via `LiveLocationShareService.start(recipients)` (see the "How to activate live sharing" note in `docs/live-relay-live-gps-debug-plan.md`).
- Still-untested paths for the UX build: share with the location-tracking switch **OFF** (exercises the coordinator's GPS re-arm incl. iOS relaunch), and flush-on-connect.

PR #781, opened 2026-06-21
Event composer: dismissible photo menu on iOS + prevent end-before-start time

2002Bishwajeet:
Fixes two sub-issues in the Event composer (M3 bottom sheet).

## 2a — Photo dropdown can't be dismissed on iOS / janky on Android
**What:** The "Add photo" dropdown's items (`Choose from Gallery` / `Take Photo`) launched the picker synchronously inside `onClick`, *before* the `DropdownMenu` had dismissed.

**Why it broke:** On iOS (CMP 1.10.2) presenting a `UIViewController` (picker/camera) while the menu is still dismissing wedges the menu open — it can't dismiss with a controller presented over it. On Android it dismissed but janky.

**Fix:** `onClick` now only closes the menu and records a `pendingPhotoAction` (`PhotoSource.Gallery`/`Camera`). A `LaunchedEffect(pendingPhotoAction)` launches the corresponding picker *after* the menu has dismissed, then clears the pending action. This mirrors the existing deferred-picker idiom in `VaultScreen` (the iOS fix from PR #760). `onDismissRequest` already closes the menu, so tapping outside dismisses it.

## 2b — End time could be set before the start time
**What:** No guard prevented `end < start`. `isValid` only checked `title.isNotBlank() && !sending`.

**Fix (calendar-style, clamping-first):**
- Changing the **START** (date or time) shifts the **END** forward by the same delta so the duration is preserved (Google-Calendar behavior) — end always stays after start.
- Picking an **END** at or before start clamps it to `start + 1h` (matching the default one-hour event length).
- Added a defensive `timesValid` `derivedStateOf` (`!hasEndTime || start < end`) folded into `isValid`, so **Send is disabled** if an invalid state is ever reached.
- Inline error `Text` (new string `chat_event_error_end_before_start`, `colorScheme.error`) under the end row as a last resort — clamping should normally prevent it from ever showing.

## Notes
- New string added to default `strings.xml` only (no `chat_event_` strings are translated in `values-da/`), imported explicitly as required.
- Compile: `:homebase-chat:compileKotlinJvm` passes; `:homebase-common:jvmTest` (Konsist) passes.
- **Needs a physical-device test** for the iOS dismiss behavior — the simulator may not reproduce the wedge.


2002Bishwajeet:
closes #786

PR #787, opened 2026-06-21
Fix iOS "no such column: fileState" launch crash (rebuild stale schema before create) + tests

Seifert69:
## What

Several iOS users (not Android, not Desktop) crash on launch with:

```
SQLiteExceptionErrorCode: error while compiling: CREATE INDEX IF NOT EXISTS idx_unread_cover
  ON DriveMainIndex(identityId, fileType, dataType, groupId, userDate, originalAuthor, fileState, archivalStatus)
no such column: fileState
```

This PR adds **regression test coverage only** — no production change.

## Root cause (proven from the code)

There are no SQLDelight `.sqm` migrations; schema upgrades are handled by a `-- Version: N` comment in the `DriveMainIndex` CREATE statement. On launch `DatabaseManager.initialize()` reads it and, if stale, calls `wipeAndRecreate()` (DROP every table, then `Schema.create()`).

The bug is **ordering**: `DatabaseManager.init {}` runs `OdinDatabase.Schema.create(driver)` *unconditionally and before* that version check (`DatabaseManager.kt:134` vs `:223-231`). `fileState` was added to `DriveMainIndex` (schema v4→5), and the **new** covering index `idx_unread_cover` over `fileState` shipped afterwards. On a device whose `DriveMainIndex` predates `fileState`:

1. `CREATE TABLE IF NOT EXISTS DriveMainIndex(...)` → no-op, the old (no-`fileState`) table is kept.
2. `CREATE INDEX IF NOT EXISTS idx_unread_cover ON DriveMainIndex(... fileState ...)` → throws `no such column: fileState`. The index is *new*, so `IF NOT EXISTS` does not short-circuit it — it evaluates its column list against the stale table.

This throws **before** the version-comment wipe net can fire — which is why a column added a while back only started crashing when a *new index over it* shipped.

## Tests

`DriveMainIndexCreateOrderTest` lives in `commonTest`, so it runs on **JVM, Android (host), and iOS** (`nativeTest` / `iosSimulatorArm64Test`) from one source. wasmJs is excluded (no SQLDelight test driver) in `build.gradle.kts`.

- **`staleSchemaThrowsOnNewFileStateIndex`** — stages the real v4 `DriveMainIndex` + its v4 indexes (so the failure lands on `idx_unread_cover`, matching the report), then asserts `OdinDatabase.Schema.create()` throws `no such column: fileState`. Reproduces the exact production crash on the iOS toolchain in CI.
- **`dropBeforeCreateRecoversCleanly`** — runs what `wipeAndRecreate()` does (DROP every `TABLE_NAMES` table, then create) and asserts `fileState` and `idx_unread_cover` come back on a clean table. Proves the recovery mechanism is sound; the defect is purely the create-before-check ordering.

Both tests **pass** today on `main` without any production fix — Test A goes green precisely because the crash reproduces (`assertFailsWith`).

Supporting plumbing: a new `createRawInMemoryDriver()` test helper (raw driver, no schema pre-applied) with per-target actuals, plus a no-op `EmptySchema` for the native driver.

## Verification

- `:homebase-api:jvmTest` — 2 tests, 0 failures (local).
- `:homebase-api:compileTestKotlinWasmJs` + `:homebase-api:compileAndroidHostTest` — green.
- iOS runs in CI (`build-check.yml` → `:homebase-api:iosSimulatorArm64Test`); this PR is partly to confirm the new `nativeTest` compiles and the reproduction holds on `sqliter`.

### DIFF
--- a/homebase-api/build.gradle.kts
+++ b/homebase-api/build.gradle.kts
@@ -159,6 +159,7 @@ kotlin {
     val wasmJsDbBackedTestClasses = listOf(
         "id.homebase.api.sync.database.ChatReadCountWrapperTest",
         "id.homebase.api.sync.database.CursorSyncTest",
+        "id.homebase.api.sync.database.DriveMainIndexCreateOrderTest",
         "id.homebase.api.sync.database.DriveLocalTagIndexTest",
         "id.homebase.api.sync.database.DriveMainIndexTest",
         "id.homebase.api.sync.database.DriveTagIndexTest",
--- a/homebase-api/src/androidHostTest/kotlin/id/homebase/api/sync/database/TestDatabaseHelper.androidHost.kt
+++ b/homebase-api/src/androidHostTest/kotlin/id/homebase/api/sync/database/TestDatabaseHelper.androidHost.kt
@@ -11,3 +11,7 @@ actual fun createInMemoryDatabase(): SqlDriver {
     OdinDatabase.Schema.create(driver)
     return driver
 }
+
+// Raw driver, no schema applied — the test stages its own DDL.
+actual fun createRawInMemoryDriver(): SqlDriver =
+    JdbcSqliteDriver(JdbcSqliteDriver.IN_MEMORY)
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/DatabaseManager.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/DatabaseManager.kt
@@ -124,6 +124,13 @@ class DatabaseManager(
     private var database: OdinDatabase
     internal var driver: SqlDriver = driverProvider()
 
+    // The `-- Version:` stamped on the on-disk DriveMainIndex CREATE statement at the
+    // moment this driver was opened, BEFORE we (re)created the schema. -1 when there
+    // was no prior database (fresh install). [initialize] reads it to surface the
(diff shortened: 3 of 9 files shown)


Seifert69:
## Update: root-caused, fixed, and confirmed on iOS

**Fix added** (commit `f84c336ae`): the create/version-check ordering in `DatabaseManager` — stale tables are now dropped *before* `Schema.create()` instead of relying on `initialize()`'s wipe *after* it. Common code; fixes all platforms.

**iOS confirmation captured.** Before the fix, the test reproduced the exact field exception on the simulator:
```
DriveMainIndexCreateOrderTest.databaseManagerOpensOverStaleSchema[iosSimulatorArm64] FAILED
  co.touchlab.sqliter.interop.SQLiteExceptionErrorCode
```
(from CI run 27918170241). With the fix, it passes on JVM (928/0 locally) and should pass on iOS.

**Note on the iOS CI job:** the `linkDebugTestIosSimulatorArm64` step is currently **flaky on `main`** with a Kotlin/Native cache-build assertion (`Lowering ReturnsInsertion: phases [...] not satisfied`) — present on `main@f0e1ba68d` independent of this branch (it re-ran green with no changes). If this PR's iOS job fails at *compile/link* (not at a test), it's that flake — re-run it. Worth a separate look at K/N cache stability for the team.

PR #836, opened 2026-06-23
chore(android): disable R8 identifier obfuscation, keep shrink/optimize (#831)

### LINKED ISSUE #831: Disable R8 identifier obfuscation (-dontobfuscate) — keep shrink/optimize — so wild crashes are traceable from raw logs (follow-up to #807)
## Problem / motivation
Follow-up to **#807** ("Keep line numbers in obfuscated release crash traces"). #807 kept obfuscation **on** and restored the line table so Crashlytics/Play traces resolve to exact lines. But in practice it's **nearly impossible to trace crashes in the wild** when identifiers are still obfuscated: a raw trace from `homebase.log`, `logcat`, a tester's screenshot, or the in-app crash-recovery screen shows `a.b.c` and needs `retrace` + the **exact** build's `mapping.txt`. That mapping dependency is the bottleneck for real-world triage.

This issue: **disable identifier obfuscation only** (`-dontobfuscate`) while keeping R8 shrinking + optimization, **if** measurement shows no major downside — and drop the now-incorrect security justification.

## The key untangling: obfuscation ≠ minification
`androidApp/build.gradle.kts` sets `isMinifyEnabled = true` for **release** (line 85) and **dev** (line 104), plus `isShrinkResources = true` (lines 89/108); **debug** is `false` (line 120). R8 folds three things into that one flag:
1. **Shrinking** — dead-code/resource elimination
2. **Optimization** — inlining, devirtualization, etc.
3. **Obfuscation** — renaming identifiers to `a.b.c`

The "keeps the AAB small" rationale in CLAUDE.md is overwhelmingly about (1) and (2) — **which we keep regardless**. `-dontobfuscate` in `androidApp/proguard-rules.pro` turns off only the **renaming** while preserving all shrink/optimize wins. So the real question is narrow: *rename identifiers or not.*

## The security argument is void for this repo
The CLAUDE.md note frames obfuscation as "defense-in-depth for a crypto/E2E app." For an **open-source** app whose sources are on GitHub, renaming `LiveLocationShareService` → `a.b.c` in the shipped DEX protects nothing that isn't already public. The security justification doesn't survive the open-source fact and should be **removed**, not used to keep renaming.

## The honest non-security ledger (renaming only)
**Keeping obfuscation buys essentially one thing:**
- **Marginal extra size.** Shorter names shrink the DEX string/method/field tables — real but small on an already-shrunk APK, typically low single-digit %, often a few hundred KB for an app this size.

**Disabling obfuscation buys:**
- **Readable traces everywhere, no mapping dependency.** `homebase.log` / logcat / screenshots / crash-recovery screen become directly readable — no `retrace`, no per-build `mapping.txt`, no Crashlytics round-trip.
- **No stale-mapping failure mode.** A wrong/old `mapping.txt` silently yields wrong names/lines; removing renaming removes that whole "the trace lied to me" class.
- **Fewer keep-rule footguns.** Reflection, kotlinx.serialization, JNI, name-based DI, Compose tooling all break under renaming unless `-keep` rules cover them. Less renaming = smaller "works in debug, crashes in release" surface.

Line numbers are already independent of this (we keep `SourceFile,LineNumberTable` per #807, `proguard-rules.pro:19`), so #807's win survives either way — obfuscation only garbles names, not lines.

## Deciding fact: measure the AAB size delta
The whole call reduces to one number. **Build the release AAB with and without `-dontobfuscate` and report the exact size difference.** If the delta is small (e.g. a few hundred KB) — and given Crashlytics already deobfuscates *and* we value debuggability from logs/screenshots (which #807 explicitly does) — ship `-dontobfuscate`.

## Proposed change (pending the measurement)
1. Add `-dontobfuscate` to `androidApp/proguard-rules.pro`; keep `isMinifyEnabled = true` + `isShrinkResources = true` (shrink + optimize stay on) for both release and dev.
2. With renaming off, `-renamesourcefileattribute SourceFile` (`proguard-rules.pro:24`) becomes moot — keep or drop, note it.
3. Update **CLAUDE.md**'s "De-obfuscating release crash stack traces" section: remove the security justification, and adjust the guidance now that raw traces are directly readable (no mandatory `retrace`/`mapping.txt` for names; mapping no longer needed for triage).
4. Sanity-check that release still builds and runs (R8 shrink/optimize on, no renaming); confirm no behavior change.

## Affected modules / targets
`androidApp` (`proguard-rules.pro`, possibly `build.gradle.kts` comments), and `CLAUDE.md`. Android release/dev build types only.

## Acceptance criteria
- [ ] Measured AAB size delta (with vs without `-dontobfuscate`) reported in the issue/PR.
- [ ] If delta is acceptable: `-dontobfuscate` enabled; release & dev AABs still shrink+optimize (minify stays on), build succeeds, app runs.
- [ ] A raw release stack trace (logcat / `homebase.log` / crash-recovery screen) shows real class/method names with exact line numbers, **without** retrace/`mapping.txt`.
- [ ] CLAUDE.md updated: security justification removed; crash-reading guidance reflects no-obfuscation reality.
- [ ] No regression from removed renaming (reflection/serialization/DI/JNI paths verified at runtime, not just compile).

## Related
Builds on **#807** (kept obfuscation + restored line numbers). This reverses the "why not just disable obfuscation?" decision in #807 specifically on the security ground, which is invalid for an open-source app.

## Scope for the plan
In scope: measure the size delta, and (if no major downside) disable identifier obfuscation via `-dontobfuscate` while keeping shrink+optimize, plus the CLAUDE.md correction. Out of scope: turning off minification/shrinking/optimization, and changing Crashlytics/Play mapping upload (harmless to keep).

2002Bishwajeet:
## What

Add `-dontobfuscate` to `androidApp/proguard-rules.pro` so R8 stops **renaming identifiers**, while keeping `isMinifyEnabled = true` and `isShrinkResources = true` for both `release` and `dev` — i.e. **shrinking + optimization stay on**, only the renaming is turned off. Plus the CLAUDE.md correction the issue asks for.

## Measured AAB size delta (the deciding fact)

**Variant measured: `release`** — built locally with `./gradlew androidApp:bundleRelease`. The release signing config in `build.gradle.kts` falls back to the bundled `debug.keystore` when `SIGNING_KEYSTORE_FILE_PATH` is unset, so the **release** AAB built and signed fine without secrets (I did *not* need the `dev` fallback). Exact `.aab` byte sizes (`wc -c`):

| Build | `.aab` bytes |
|---|---|
| Obfuscation **ON** (baseline, as-is) | `69,362,626` |
| **`-dontobfuscate`** (renaming off) | `69,405,251` |
| **Delta (OFF − ON)** | **`+42,625` bytes (~41.6 KiB, +0.061%)** |

The AAB grew by **~42 KB (+0.061%)** — an order of magnitude *smaller* than the issue's "low single-digit % / few hundred KB" expectation. Verdict: shipping `-dontobfuscate` is clearly acceptable. (The tiny delta also confirms shrink/optimize are doing the real size work, not renaming.)

I verified renaming is genuinely off by inspecting `mapping.txt`: classes map to themselves (`id.homebase.feed.MainActivity -> id.homebase.feed.MainActivity`), and there are **zero** single-letter obfuscated names. The only `source -> target` differences are `R8$$REMOVED$$CLASS$$N` entries — that's R8 **shrinking** (dead-code removal), which we keep, not renaming.

## `-renamesourcefileattribute`

`-renamesourcefileattribute SourceFile` (proguard-rules.pro) becomes a **no-op** once renaming is off. **Kept it** (with a one-line note) to minimise the diff — keeping is harmless.

## CLAUDE.md

Rewrote the "De-obfuscating release crash stack traces" → "Reading release crash stack traces" section: removed the void "defense-in-depth for a crypto/E2E app" / "keeps the AAB small" security justification, and dropped the mandatory `retrace` + per-build `mapping.txt` step for *names* (raw traces now show real names directly). Kept the Crashlytics/Play paragraph (still useful, still uploads mapping). Line numbers are unaffected — `-keepattributes SourceFile,LineNumberTable` stays (per #807).

Also updated the now-stale "code shrinking, obfuscation, and optimization" comments on both `isMinifyEnabled = true` lines in `build.gradle.kts`.

## Acceptance criteria

- [x] Measured AAB size delta (with vs without `-dontobfuscate`) reported — see table above (+42,625 bytes, +0.061%, release variant).
- [x] Delta acceptable → `-dontobfuscate` enabled; release & dev keep `isMinifyEnabled`/`isShrinkResources = true`; **release build succeeded** (BUILD SUCCESSFUL, R8 shrink/optimize ran).
- [x] A raw trace would show real class/method names + exact line numbers without retrace/`mapping.txt` — confirmed via `mapping.txt` (identity class mappings, line table kept). *(Verified from the build artifact, not from a live on-device crash.)*
- [x] CLAUDE.md updated: security justification removed; crash-reading guidance reflects no-obfuscation reality.
- [ ] **No runtime regression from removed renaming** — **NOT runtime-verified.** Removing renaming is strictly safer for reflection / kotlinx.serialization / DI / JNI (those break *under* renaming, not without it), and the build + R8 succeeded with no new warnings, but I did not install and exercise the signed release on a device to confirm those paths at runtime.

## Caveats / what I did not verify

- Measured the **release** variant (debug-keystore signing fallback), not a Play-signed production AAB; the size delta is representative regardless.
- The "no runtime regression" criterion is verified only at build/R8 level, not by running the app on a device.

Closes #831

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -178,30 +178,24 @@ adb logcat -d --pid=$(adb shell pidof id.homebase.feed.dev)
 adb logcat -c
 ```
 
-### De-obfuscating release crash stack traces
+### Reading release crash stack traces
 
-Release and `dev` builds run R8 with `isMinifyEnabled = true`, so on-device
-traces (logcat, `homebase.log`, the crash-recovery screen, a tester's
-screenshot) show obfuscated names like `a.b.c`. **Do not disable obfuscation** —
-it's defense-in-depth for a crypto/E2E app and keeps the AAB small. Instead,
-translate the trace with the `mapping.txt` R8 emits each build.
+Release and `dev` builds run R8 with `isMinifyEnabled = true` for shrinking and
+optimization, but `proguard-rules.pro` sets `-dontobfuscate`, so identifier
+renaming is **off**. On-device traces (logcat, `homebase.log`, the crash-recovery
+screen, a tester's screenshot) already show the real class/method names — e.g.
+`id.homebase.feed.MainActivity` — with exact line numbers (`-keepattributes
+SourceFile,LineNumberTable`). No `retrace` and no per-build `mapping.txt` is
+needed to read names; a raw trace is directly triageable.
 
-- **Crashlytics / Play Console** de-obfuscate automatically: the Crashlytics
-  Gradle plugin uploads `mapping.txt` on every minified build. `proguard-rules.pro`
-  keeps `SourceFile,LineNumberTable`, so these show the exact original file + line.
-- **A raw trace** (logcat / log file / screenshot) — retrace it locally against
-  the mapping for the exact build that produced it:
+- **Crashlytics / Play Console** still work as before: the Crashlytics Gradle
+  plugin uploads `mapping.txt` on every minified build, and these show the exact
+  original file + line. Harmless to keep.
 
-```bash
-# mapping.txt is written per variant by the release build:
-#   androidApp/build/outputs/mapping/release/mapping.txt   (release)
-#   androidApp/build/outputs/mapping/dev/mapping.txt       (dev)
-# Save the obfuscated trace to trace.txt, then:
-retrace androidApp/build/outputs/mapping/release/mapping.txt trace.txt
-# `retrace` ships in the Android SDK (build-tools / cmdline-tools). `r8 retrace`
-# also works. Keep the mapping.txt from each shipped build — a stale mapping
-# from a different build produces wrong line numbers.
-```
+(Renaming was disabled because the app is open-source — obfuscating names that
+are public on GitHub anyway protected nothing, while readable raw traces make
+real-world crash triage from logs/screenshots far easier. Shrinking and
+optimization, which is where R8's size win actually comes from, stay on.)
 
 ## Desktop App Logs (JVM / Android Studio `desktopApp:run`)
 
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -80,8 +80,8 @@ android {
 
     buildTypes {
         release {
-            // Enables code shrinking, obfuscation, and optimization for only
-            // your project's release build type.
+            // Enables code shrinking and optimization. Identifier obfuscation is
+            // turned off via -dontobfuscate in proguard-rules.pro.
             isMinifyEnabled = true
 
             // Enables resource shrinking, which is performed by the
@@ -99,8 +99,8 @@ android {
         create("dev") {
             applicationIdSuffix = ".dev"
 
-            // Enables code shrinking, obfuscation, and optimization for only
-            // your project's release build type.
+            // Enables code shrinking and optimization. Identifier obfuscation is
+            // turned off via -dontobfuscate in proguard-rules.pro.
             isMinifyEnabled = true
 
             // Enables resource shrinking, which is performed by the
--- a/androidApp/proguard-rules.pro
+++ b/androidApp/proguard-rules.pro
@@ -12,15 +12,19 @@
 #   public *;
 #}
 
-# Keep line numbers in release stack traces. Obfuscation stays ON — this only
-# preserves the line-number table so Crashlytics / Play Console (which have the
-# uploaded mapping.txt) and `retrace` resolve crashes to the exact original
-# file + line instead of showing "Unknown Source".
+# Disable identifier renaming only. R8 shrinking + optimization stay on (the AAB
+# size wins are theirs, not renaming's). The app is open-source, so renamed
+# identifiers protect nothing public; leaving names intact makes raw traces
+# (logcat / homebase.log / crash-recovery screen / screenshots) directly readable
+# with no retrace / per-build mapping.txt.
+-dontobfuscate
+
+# Keep line numbers in release stack traces so Crashlytics / Play Console and
+# `retrace` resolve crashes to the exact original file + line instead of
+# "Unknown Source".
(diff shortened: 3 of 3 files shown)


2002Bishwajeet:
### Scope note: iOS needs no equivalent change

This PR is intentionally Android-only, and there's nothing to mirror on iOS.

Identifier renaming (`a.b.c`) is an R8/ProGuard concept. iOS has no equivalent — Kotlin/Native builds the shared `ComposeApp` framework and Xcode builds `iosApp`, and **neither obfuscates/renames identifiers**. So the "raw release traces show `a.b.c`" problem this PR fixes doesn't exist on iOS; there is no `-dontobfuscate` analog to apply.

The iOS counterpart of Android's `mapping.txt` is **dSYM symbolication**, which is already set up:
- Release/Dev build with `DEBUG_INFORMATION_FORMAT = dwarf-with-dsym` (Debug uses plain `dwarf`) — `iosApp.xcodeproj/project.pbxproj`.
- An **"Upload Crashlytics DSYM files"** build phase uploads dSYMs to Firebase, so Crashlytics auto-symbolicates iOS crashes (the parallel to Android's `mapping.txt` upload).

Residual (inherent to native symbolication, not addressable here): a *raw, un-symbolicated* iOS crash (device console / screenshot / crash-recovery screen) still needs the matching dSYM to resolve native frames — there's no obfuscation switch to remove that the way `-dontobfuscate` does on Android. Out of scope for #831.

Issue #846, opened 2026-06-24
Review & streamline foreground GPS acquisition → submission to live sharing / history tracking


Seifert69:
## Goal
Review how & when we **acquire GPS points in the foreground**, and how each fix is then **submitted** to **live sharing** and/or **history tracking** (per the user's settings). The current flow works but spans several layers with independently-chosen intervals and an implicit join point — I want it **streamlined, intentional, and well-documented** (every cadence/threshold justified, the live-vs-history split obvious in one place).

This is the **acquisition + submission cadence** review; it pairs with **#835** (which packages the history-vs-live *routing decision*). Also related: **#823** (don't persist history when tracking off), **#824** (receive path).

## Current flow (as mapped)
**Acquire** (gated by `LocationTrackingCoordinator.wantsGps() = trackingEnabled || liveShareActive`):
- Android foreground: FusedLocation `PRIORITY_HIGH_ACCURACY`, **15s** interval, **10m** min displacement (`LocationTracker.android.kt`). Background: balanced, 60s, 25m, 10-min batch.
- iOS foreground: `kCLLocationAccuracyBest`, distanceFilter **10m**, `pausesLocationUpdatesAutomatically`. Background: 100m / 50m. Plus always-on significant-location-change.
- A **foreground ticker** (`LocationTrackingCoordinator.startTicker`) fires `onFlushDue()` immediately on entering foreground, then every **120s**.

**Submit** (`LocationPointStore.submit`):
- Dedup (drop `p.t <= last.t`) + a **stationary-noise filter**: drop if `<8m` AND `<20s` since last accepted (`MIN_DISPLACEMENT_M=8`, `MIN_INTERVAL_MS=20s`).
- History persist gated on `isTrackingEnabled()`; `_lastPoint` updated unconditionally; then `onPointsBuffered()` fired **unconditionally**.

**Route** (the `onPointsBuffered` seam, wired in `AppModule`):
- **Live:** `LiveLocationShareService.onGpsBuffered()` — relay throttle **3s** (`MIN_INTERVAL_MS`), unique-recipient dedup, last-value-wins via tryLock, prunes expired roster.
- **History:** `LocationTrackUploaderService.flushIfDue()` — flush rate-gate **60s** (`MIN_FLUSH_INTERVAL_MS`), 7-day retention.

### The intervals/thresholds, side by side
| Layer | Foreground | Background | Note |
|---|---|---|---|
[...]

PR #848, opened 2026-06-24
fix(location): link to enable history on Find Device empty state

Seifert69:
## What & why

On **Find Device** (`Route.LocationFindDevice`), locating a device renders its recorded **history**. With location history OFF, *this* device has never recorded a position, so the screen showed a dead-end "No known location for this device" + a refresh button that can never help.

Now, when the located device is **this device** and history is off, it shows a short explanation plus a tappable **"Turn on location tracking"** link to the Location dashboard (where the history toggle lives). Other devices — whose tracking we can't control — and the history-on-but-no-fix-yet case keep the plain "no location" text.

## How

- `FindDeviceUiState` gains `allowLocationHistory`, **reactively collected** so the hint clears as soon as the user enables history and returns.
- Reuses `LocationDeviceInfo.isThisDevice` and the existing `DayPlaybackMap` link pattern (`location_history_enable_tracking`, underlined primary, `clickable`); adds one explanatory string `location_find_history_off`.
- `AppNavHost` wires `onEnableHistory` to the existing `openLocation` lambda → `Route.Location`.

## Verification

- Compiles JVM/Android/Web for `homebase-core`; Konsist `ArchitectureTest` green (strings via `stringResource`).
- Manual (Android): history OFF → Find Device → this device → explanation + link → dashboard; enable history, return → hint clears reactively, then "no location" until the first fix uploads. A different device with no data still shows the plain text.

Follow-up to the location pipeline work in #847.

### DIFF
--- a/homebase-common/src/commonMain/composeResources/values/strings.xml
+++ b/homebase-common/src/commonMain/composeResources/values/strings.xml
@@ -1298,6 +1298,7 @@
     <string name="location_find_last_seen">Last seen %1$s</string>
     <string name="location_find_battery">🔋 %1$s</string>
     <string name="location_find_no_location">No known location for this device</string>
+    <string name="location_find_history_off">Location history is off, so this device hasn't recorded its position yet.</string>
     <string name="location_find_freshness">Position is as of the device's last upload and this device's last sync.</string>
     <string name="location_find_refresh">Refresh</string>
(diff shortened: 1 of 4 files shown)


Seifert69:
Closing per discussion — folding focus onto #846 instead. The empty-state hint only applied to locating *this* device with history off (other devices / the picker were unaffected, since their data comes from their own uploads), but the value is marginal enough to skip for now. Easy to revisit later.

Issue #888, opened 2026-06-29
Desktop (high priority): slow startup — UI hidden ~2s+ behind serialized DB loads (conversation list + eager Vault/Sticker/Contact/Moments streams)


Seifert69:
## Summary
The **Desktop** client has a long startup — the window/UI doesn't appear for several seconds. Analysis of a startup log shows it's **data-load-bound**: a wall of slow, largely-serialized DB reads (conversation list + Vault/Stickers/Contacts/Moments stream loads) runs **on the path to first composition**, so the window stays hidden until it's done.

## Startup timeline (from a desktop `dev` log; times redacted of personal data)
Using the existing `StartupLogger` checkpoints + `ConvListPerf` metrics (deltas from a single startup):

| t (s) | event |
|------|-------|
| 0.000 | `(StartupLogger) checkpoint: Compose application{} entered` |
| 0.27 | `ConvListPerf loadBasicConversations end-to-end=526ms (query=437 map=89) items=86` (started before this) |
| 0.31 | `checkpoint: DesktopViewModel resolved` |
| 0.26→0.64 | `enrichWithLastMessages=385ms`, `enrichWithAdmins=173ms` |
| 0.43 | `VaultStream loadAll: 14 sections, 33 entries` |
| 0.76 | `enrichAllConversationsWithUnreadCounts=314ms` |
| 0.99 | `StickerStream loadAll` |
| 1.03 | `ContactBookStream loadAll: 74 contacts` |
| **1.71** | `checkpoint: App() first composition` |
| 1.72 | `checkpoint: App() Koin injections done` |
| **2.30** | `(DesktopLifecycleOwner) onWindowVisible -> STARTED` |

So **~1.7 s elapses between "Compose application entered" and "first composition"**, filled entirely with synchronous data loading, and the **window only becomes visible ~2.3 s after that checkpoint** (true process-start is earlier — the log buffer was cleared, so total wall-clock is longer; capture a clean run to measure process-start → first-frame).

## Root cause — slow, serialized DB reads on the first-frame path
18 `SlowDbRead` warnings in that short window. Two distinct costs:

1. **Mapper (JSON-header) cost dominates** on `driveMainIndex` reads — the SQL is trivial but the row mapping is **168–403 ms**:
   - `sql=14.9ms mapper=403ms`, `sql=2.7ms mapper=268ms`, `sql=6µs mapper=233ms` … (driveMainIndex header deserialization is the bottleneck, not the query).
[...]

PR #896, opened 2026-06-29
perf(desktop): defer non-critical add-on stream loads off the first-frame path (#888)

2002Bishwajeet:
Addresses #888 (desktop slow startup) — the **safe, high-confidence slice**. The deeper DB-internals work is intentionally deferred to a measurement-driven follow-up (see below).

## Root cause
`AppModule.onPostAuthenticated` fired a wall of stream cold-loads (each a `QueryBatch` DB read + JSON-header deserialize) the instant auth resolved — racing the conversation-list cold load **and** Compose's first frame on the shared DB read-connection pool. ~1.7s elapsed between "Compose application entered" and "first composition", filled with synchronous data loading.

## Change
Move the **non-critical** add-on streams off the cold-start/first-frame path:
- `onPostAuthenticated` now starts only **ConversationStream** (the landing screen) + **ContactService** (chat name resolution the list needs).
- **Vault**, **Moments** (feed + groups + recipient-lookup + user-state), and the **location uploader** start from `AppNavHost` `LaunchedEffect`s that run **after first composition**, gated on the add-on's **icon visibility** — a user who hasn't activated an add-on never pays its cold load. `start()` is idempotent; streams reset on `SessionEnded`.
- **Contact Book** and **Sticker** streams were already loaded lazily elsewhere (`ContactBookViewModel.ensureLoaded()` / the sticker-tray mount chain) — the eager burst lines were **redundant** and removed.
- Location uploader's `start()` (retention sweep + `countUnmarked` DB read) also deferred to after first frame.

## Deliberately NOT done
- **Conv-list enrich left serial.** The three enrich passes read-modify-write the *same* `_conversations` StateFlow (mutex-guarded, single-threaded DB dispatcher) — parallelizing would cause a lost-update race for no throughput gain (the code comments say as much). Not touched.
- **DB read-pool size / `driveMainIndex` JSON-header mapper (~400ms)** — the deeper bottlenecks, but they require runtime profiling to tune safely. Left for a measurement-driven follow-up.

## Measurement harness (added)
A JVM-process-start anchor (`RuntimeMXBean.startTime`) + a window-visible checkpoint in `StartupLogger`, so you can size **process-start → first-frame → window-visible** from the log.

## ⚠️ Verification needs you (logged-in desktop)
Per #888's acceptance criteria this needs a before/after `StartupLogger` measurement, which needs your environment:
1. `./gradlew :desktopApp:run`, log in so the main screen + nav render.
2. `grep StartupLogger ~/Library/Application\ Support/HomebaseChatDev/logs/homebase.log`
3. Compare the **"Compose application entered" → "App() first composition"** gap (and process-start → window-visible) on `main` vs this branch.

## Compile/test (local)
- ✅ `:homebase-api/:homebase-chat/:homebase-core/:desktopApp:compileKotlinJvm` — BUILD SUCCESSFUL
- ✅ `:homebase-core:compileAndroidMain` + `:compileKotlinIosSimulatorArm64` (commonMain edits don't break mobile)
- ✅ `:homebase-common:jvmTest :homebase-chat:jvmTest` (incl. Konsist ArchitectureTest)

## Scope: 4 files, +75/−16 (`AppModule.kt`, `AppNavHost.kt`, `Main.kt`, `DesktopLifecycleOwner.kt`)


2002Bishwajeet:
## ⚠️ Measured before/after — this does NOT move startup on the test machine

Ran the desktop app on both `origin/main` and this branch, same machine + same logged-in session (80 conversations), back-to-back:

| Segment | Baseline (origin/main) | This branch (#888) |
|---|---|---|
| `main()` → **database initialized** | **4.03 s** | **4.12 s** |
| Compose entered → first composition | 0.52 s | 0.64 s |

**No measurable startup improvement.** The dominant cost is **~4 s of database initialization in `main()`, before Compose** — which this PR never touches (it's pre-auth). The baseline ran *second*, with the DB file already warm in disk cache, and still showed 4.0 s, so that ~4 s is **real work** (likely SQLCipher key-derivation / migrations / index build), not a cold-cache artifact. The post-login stream loads this PR defers simply weren't the bottleneck on this dataset (the issue's original log was a heavier/colder state where they measured 0.4–1.0 s).

**Recommendation:** this PR is still a correct, low-risk deferral (don't block first-frame on non-critical loads) and the StartupLogger anchors it adds are what enabled this measurement — worth keeping as hygiene. But it is **not** the startup fix. The real win is the **~4 s pre-Compose DB init** — that should be a separate, profiled follow-up (it's the actual bottleneck #888 set out to find).

2002Bishwajeet:
Closing as we found no noticeable perf improvs

PR #897, opened 2026-06-29
feat(chat): personal, synced pinned-messages bar (#887)

### LINKED ISSUE #887: Chat: personal pinned-messages bar (Signal-style) — synced to your own devices via localAppData tags; auto-pin Poll/Event/Groodle/live-share
## Problem / motivation
Add a Signal-style **pinned messages** bar to a conversation — but **personal/local**, not a shared group pin. Important messages (polls to answer, events coming up, groodles to fill in) should surface at the top of the chat so they aren't lost in scroll, and clear themselves once they're no longer actionable.

## Proposed behaviour
- **One-line pinned bar** at the top of the conversation showing a single pinned message (sender + a one-line preview), **regardless of how many are pinned** (show "1 of N" when there are several).
- **Tapping the bar advances to the next pin** (cycles through pinned messages; scrolls/highlights that message in the list).
- **A menu / affordance on the bar opens a slide-up panel** listing **all** pinned messages; from there you can **jump to** a message or **unpin** it (Signal-style).
- **Personal, but synced to *your own* devices:** pins are **not** shared with the group, but they **do sync across all of your own devices** (per-identity) via the API — not stuck on one device. (User can manually pin/unpin their own view.)

### Auto-pin defaults (the main value)
- **Poll, Event, Groodle, and live-location shares** are **auto-pinned** when they arrive/are sent.
- **Poll** and **Groodle** **auto-unpin once you fill them out** (you've voted/responded).
- **Event** stays pinned **until it expires**, then auto-unpins.
- **Live-location share** stays pinned **until the share expires** (or is stopped), then auto-unpins — same "until it expires" semantics as Event.

## Pins = a synced, owner-private tag (use the API — `localAppData.tags`)
Use the existing **owner-private, synced** local-metadata mechanism, **not** the device-local `DriveLocalTagIndex`. A file's `localAppData` (`FileMetadata.localAppData: LocalAppMetadata`) is **per-identity, encrypted to the owner, and synced to the owner's own devices but NOT shared with other recipients/group members** — exactly "personal but synced". Precedent: chat read-state already rides here (`localAppData.lastReadTime`, `ChatReadCountWrapper`), and `localAppData.tags` is already indexed (`MainIndexMeta`).
- A pin = a reserved `PIN_TAG` UUID in the message file's **`localAppData.tags`**.
- Pin = add the tag via the API (`UpdateLocalMetadataTagsOutboxRequest` → `DriveOutboxUploader.UpdateLocalMetadataTags`, enqueued in `OutboxSync`); unpin = remove it. The change **syncs to your other devices** and stays private to you. No message descriptor / protocol change; no group visibility.
- Query pinned messages locally via the indexed `localAppData.tags` (`MainIndexMeta`).

## Technical grounding (leads, not a prescription)
**Typed kinds + auto-pin** (`ChatProtocol.kt`): Event=210, DiceRoll=212, **Groodle=213**, **Poll=214**. On a new message whose `MessageContent` is Poll/Event/Groodle (parsed in `MessageContentParser`) — or a **location message with an active live share** — auto-pin it.

**Auto-unpin signals:**
- **Poll filled:** the user's vote is a reaction `_p{index}` — detect via `PollVote.optionOf(...)` / `ownVotes(message.ownReactions, optionCount)` (`poll/PollVote.kt`). When `ownReactions` contains a `_p\d+`, unpin.
- **Groodle filled:** votes are reactions `_{slot}{Y|N|M}` — `GroodleVote.myVotes(ownReactions, slots, allowMaybe)` (`groodle/GroodleVote.kt`). When non-empty (responded), unpin. *(Decide: any slot answered vs all slots — see open questions.)*
- **Event expired:** `EventDescriptor.endUtcMs` (`event/EventDescriptor.kt`); expired when `now > endUtcMs` (fallback `startUtcMs + 1h` when null).
- **Live share expired/stopped:** `LocationPreviewDescriptor.liveShareUntilMs` (`widget/LocationPreview.kt:288`); LIVE while `now < liveShareUntilMs`, ENDED when `now >= liveShareUntilMs` (or stopped → `liveShareUntilMs = now`). Unpin once ended.

**When to evaluate expiry:** the time-based unpins (Event, live share) are evaluated **on conversation open** — no periodic ticker needed. The vote-based unpins (Poll, Groodle) happen live via the reaction change (the user just voted).

**Conversation UI:** the bar sits between the `TopAppBar` and the `LazyColumn` in `ConversationContent.kt` (~before the message Box at ~1028). Jump-to-pin reuses the existing scroll-to-message pattern: find the index in `uiState.messages` and `listState.animateScrollToItem(idx)` (see the search-result scroll at ~331-339).

**Slide-up panel:** a `ModalBottomSheet` (model on `StickerOptionsSheet` / `ReactionsBottomSheet`) with a `LazyColumn` of pinned messages — each row = preview + tap (jump) + unpin.

**ViewModel seam:** `ConversationListViewModel` (observes `chatMessageStream`/messages) — evaluate auto-pin on new typed/live-share message, auto-unpin on vote/fill (reaction change) live, and **re-evaluate time-based unpins (Event/live-share) on conversation open**; expose the pin list + actions (`TogglePinMessage`/`UnpinMessage`/`ShowPinnedMessagesSheet`). Pin/unpin writes go through `OutboxSync` (`UpdateLocalMetadataTags`) so they sync to the user's devices.

## Open design questions (call out)
- **Tap semantics:** "tap advances to next pin" — does it also scroll the conversation to that message (Signal does)? Confirm the bar cycles **and** scrolls.
- **Groodle "filled out":** any slot answered, or all slots? (Lean: responded at all → unpin.)
- **Manual pin/unpin:** include a message long-press menu to pin/unpin any message personally (the panel already unpins) — confirm in scope.
- **Order/limit:** ordering of multiple pins (newest-first? pin time?) and whether to cap the "1 of N".
- **Future typed kinds:** make auto-pin rule-driven so new kinds can opt in.

## Affected modules / targets
`homebase-chat` (`widget/ConversationContent.kt` pinned bar + a new bottom-sheet panel; `conversationlist/ConversationListViewModel.kt` + `ConversationListUiAction.kt` for pin state/rules; reuse `poll/PollVote.kt`, `groodle/GroodleVote.kt`, `event/EventDescriptor.kt`, `widget/LocationPreview.kt` for `liveShareUntilMs`), `homebase-api` (`localAppData.tags` via `UpdateLocalMetadataTagsOutboxRequest`/`OutboxSync`/`DriveOutboxUploader.UpdateLocalMetadataTags`; read via the `MainIndexMeta` tag index). commonMain → all platforms. **Synced per-identity (owner-private), not group-shared.**

## Design / constraints
- Material 3; all strings via `stringResource()` (Konsist `ArchitectureTest` fails on literal `Text("…")` — incl. "1 of N"); M3 color/typography; `start`/`end` padding; `contentDescription` on the bar/menu icons; `Icons.AutoMirrored.*` for any directional icon.
- `collectAsStateWithLifecycle()`; pin state reactive — vote-based unpins reflect live; time-based (Event/live-share) unpins are recomputed on conversation open.
- Owner-private + synced per-identity via `localAppData.tags` (not group-visible); reflects on all the user's devices. Pin/unpin is an outbox write — keep it cheap (a tag toggle), and tolerate offline (enqueues, syncs later).
- One-line truncation must be surrogate-safe (`truncateToCodePoints`) for the preview.

## Acceptance criteria
- [ ] A one-line pinned bar shows at the top when ≥1 message is pinned, with "1 of N" for multiple; tapping advances to the next pin (and scrolls to it).
- [ ] A menu/affordance opens a slide-up panel listing all pins, with jump-to-message and unpin.
- [ ] Poll/Event/Groodle/live-share auto-pin on arrival/send; Poll & Groodle auto-unpin once answered (live); Event and live-share auto-unpin when expired, evaluated on conversation open.
- [ ] Pins are personal (not visible to the group) **and sync across the user's own devices** (per-identity, via `localAppData.tags`); a pin/unpin on one device shows on another after sync.
- [ ] (If in scope) manual pin/unpin of any message from a long-press menu / the panel.
- [ ] Verified on Android + iOS, including cross-device sync of a pin.

## Scope for the plan
In scope: a personal pinned-messages bar (one-line, cycle, slide-up panel) with the auto-pin/unpin rules for Poll/Event/Groodle/live-share, backed by **owner-private `localAppData.tags`** so pins sync across the user's own devices. Out of scope: **shared/group** pins (these are personal), and pinning non-message items.

2002Bishwajeet:
Closes #887.

A Signal-style **pinned-messages bar** — **personal** (never shared with the group) but **synced across the user's own devices** via the owner-private `localAppData.tags` mechanism. Pin any message into a one-line bar at the top of the conversation; typed messages (polls/events/etc.) pin themselves while they're live.

## How it works

A pin = the reserved `ChatProtocol.MessagePinnedTag` in a message file's `localAppData.tags`. Pin/unpin writes optimistically (`OptimisticWriter.updateLocalTags`, immediate UI) then enqueues an `UpdateLocalMetadataTags` outbox row that syncs to the user's own devices — the same proven owner-private lane as conversation-pinning and read-state. No group visibility, no descriptor/protocol change on the wire.

### UI
A one-line bar (sender + preview, **"N of M"** when several are pinned). Tapping it **cycles to the next pin and scrolls/highlights it**; a trailing affordance opens a **ModalBottomSheet** listing all pins (jump / unpin). Manual **Pin / Unpin** live in the message long-press menu, gated on pin state. Preview text handles typed kinds (whose `content` is empty) by surfacing the descriptor — poll question / event title / location address.

### Auto-pin / auto-unpin
- **Auto-pin** live typed messages — Poll / Event / Groodle, and live-location while its share is active — on first sighting, so the pin is already there when you open the conversation.
- **Auto-unpin** a Poll/Groodle once **answered**, and an **expired event / ended live-share** on open — both **synced** (they drop from the bar on every device and the server stops carrying a stale pin; `endUtcMs` is absolute UTC so devices agree). Neither is a *dismissal* — the message stays auto-pin-eligible if it becomes live again.
- **Manual pins are sticky.** A deliberate menu pin carries `ChatProtocol.ManualPinnedTag`, and both auto-unpin paths skip it — so a pin *you* placed survives the event ending or the poll being answered. `isPinned` (state) is unchanged; `ManualPinnedTag` is a provenance modifier read only by the auto-cleanup, so the single-tag bar query / index / menu gating stay put.

## Correctness (the parts that took real work)

Auto-pinning **your own** messages and keeping pins consistent **across devices** is where the naive version breaks. Four things make it reliable:

1. **Own-send survives the temp→server rekey.** A just-sent message exists only under a client temp fileId until the server confirms it and `rekeyCacheAfterCreate` swaps in the real one; the local `DriveMainIndex` row only rekeys on sync-back. So auto-pin **defers a still-sending message** and pins it on its confirmed sighting, and `DriveOutboxUploader.updateLocalMetadataTags` resolves the current fileId at **send time** via `getFileHeaderByUid` (a server fetch — same cost as the header lookup it replaces, and it returns the real server fileId even before the local row rekeys). Without this the tags write targeted a dead temp id and was dropped (`404`/`400 InvalidFile`), so your own sends silently never pinned.

2. **Pins no matter how the message arrives.** The auto-pin scan runs on **every** path that surfaces messages — `loadConversation` (cold open), `processIncrementalBatch` (live WebSocket), and `refreshCachedWindows` (silent DriveSync / `DriveEvent.Stopped`) — not just the live push. So a send made **offline**, or with the app **backgrounded** right after sending (confirmed later via DriveSync, not a WS echo), still pins.

3. **A dismissed pin is never resurrected.** Because the scan now runs on cold load and re-sync, an in-memory "already handled" set isn't enough — a restart or a second device would re-pin something you unpinned. A durable, synced **`ChatProtocol.AutoPinDismissedTag`** solves this: a manual unpin sets it (an auto-expiry unpin does not), a manual re-pin clears it, and the scan skips any message carrying it. Since it rides `localAppData.tags`, it holds across restarts **and** across the user's other devices.

4. **The bar reflects the pins on every open.** `refreshPinnedFor` runs on all load paths — including `loadConversationAroundMessage`, which opens a conversation to its saved scroll anchor (the common case) and previously set up the window without refreshing the bar. That gap left the bar empty on restart even though the pins were still in the data (the bubble menu still said "Unpin") until a live `BatchReceived` arrived — the "bar vs. data divergence across restart." Now the bar populates immediately on open.

## Verification

- ✅ `:homebase-api` / `:homebase-chat` / `:homebase-core` compile — JVM, Android (`compileAndroidMain`), and iOS (`compileKotlinIosSimulatorArm64`).
- ✅ `:homebase-chat:jvmTest` incl. **`MessagePinTagTest`** — pin enqueues the full tag list carrying the message `uniqueId`; a manual pin is sticky (`ManualPinnedTag`) and an auto-pin isn't; a manual unpin dismisses (`AutoPinDismissedTag`) and clears both pin tags; an auto-condition unpin (expiry / answered) syncs **without** dismissing; a re-pin clears the dismissal; idempotency; active-group newest-first query.
- ✅ `:homebase-common:jvmTest` Konsist `ArchitectureTest` (no hardcoded Composable strings, incl. "N of M").
- ✅ **Device-verified (iOS simulator + Android):** own-send live event pins immediately; after a manual unpin + **app restart** it stays unpinned (even though the scan now runs on cold open); logcat confirms the tags write completes on the first attempt with the real server fileId. On Android, logcat also confirms a manually-pinned **expired** event keeps `ManualPinnedTag` across restart and the expiry prune skips it (`manual=true`), and the bar populates on cold-open showing those sticky pins.

## Merge note

Touches `AppModule` (the auto-pin hook is wired at `ChatMessageStream` construction, resolving its dependency lazily to keep the DI cycle broken — so it survives warm/session-restore relaunches rather than no-op'ing). Additive; no overlap with other in-flight branches beyond that block.


2002Bishwajeet:
## On-device testing: personal pin bar has multi-layered sync/concurrency defects — not mergeable as-is

Ran the feature end-to-end on a physical Android device (Redmi Note 5 Pro, debug build) with live `logcat`. The bar/preview/menu render fine, but **the core promise — personal pins that sync across your own devices, auto-pinned on send — does not work on device.** Unit tests can't catch this: `MessagePinTagTest` asserts the outbox row is *enqueued*, with no real server.

Applying a couple of fixes (below) got auto-pin to actually *fire*, which then exposed several deeper problems in the pin **sync/concurrency model**. Capturing everything here for a proper rework rather than reactive patching.

### ✅ Verified working (after the fixes below)
- **#2 preview text** — bar/panel now show the poll question / event title / groodle title / location address (typed kinds have empty `content`, so the old code rendered a blank `"Sender:"`).
- **#4 a11y** — the cycle row now has an `onClickLabel` (`chat_pinned_bar_cycle`, previously an orphaned string).
- **Auto-pin fires** — the hook now runs (see fix #5).
- **Server accepts a null-versionTag first write** — after the uploader fix, a tags write **completed** server-side: `OutboxSync: completed UpdateLocalMetadataTags` (and a later write showed `hasRequestVersionTag=true`, i.e. the localAppData versionTag was created).

### Fixes made in the working tree (each is real + necessary; not yet committed)
1. **`shouldAutoPin` guard (`ChatMessageStream`)** — don't re-pin an already-answered poll/groodle or an already-expired event (extracted to a pure `internal fun shouldAutoPin(...)` + `AutoPinDecisionTest`, 16 cases). Prevents re-pin on re-sync for those kinds.
2. **Preview text helper** — `MessageUiModel.pinnedPreviewBody()` shared by the bar + sheet.
3. **a11y label** on the cycle row.
4. **Auto-pin hook wiring (`AppModule`)** — moved from inside `onPostAuthenticated` (which is deferred and frequently **never runs on a warm relaunch / session-restore** — the AuthCC `promoteToForeground` race) to **`ChatMessageStream` construction**, resolving `ChatMessageActionService` lazily inside the lambda to keep the DI cycle broken. *Without this, auto-pin was a permanent no-op on every warm relaunch* while manual pin still worked.
5. **`DriveOutboxUploader.updateLocalMetadataTags`** — mirror the working `updateLocalMetadataContent` sibling: **drop only when the file header is genuinely gone from the server; otherwise send with the (possibly null) versionTag.** The old code threw `NotFound` on a null versionTag, conflating "missing versionTag" with "missing file." A fresh **message** has never had `localAppData` written, so its `localAppData.versionTag` is null on first write — conversation-pin never hit this because conversations already carry a localAppData versionTag from read-state. This was silently dropping *every* first-time message pin (`— no version tag in request and local file gone`, then `404 NotFound`).

### ❌ Still broken (the actual blockers — pre-existing in the sync design)

**A. Auto-pin fights manual pin/unpin → duplicate writes → one always dropped.**
A live typed message (e.g. an event) is an auto-pin candidate on every conversation load / `BatchReceived`. So immediately after you manually **unpin** it, auto-pin **re-pins** it. One user action produces two competing `UpdateLocalMetadataTags` rows to the same message, which then collide on optimistic concurrency:
```
15:32:14.608  sending UpdateLocalMetadataTags  hasRequestVersionTag=true
15:32:14.619  sending UpdateLocalMetadataTags  hasRequestVersionTag=true
15:32:14.804  completed
15:32:14.854  DROPPING  VersionTagMismatch (47ccf319-…)
```
The `shouldAutoPin` guard (fix #1) only blocks *answered polls / expired events* — it does **not** stop a still-live event from being re-pinned after the user dismisses it. Root: auto-pin has no durable "user dismissed this" signal; the in-memory `autoPinHandled` set is per-device/per-session.

**B. Concurrent/successive tags writes to the same message drop on `VersionTagMismatch`.**
`ChatMessageActionService.updateMessageTags` and `ConversationService.updateConversationTags` capture `localAppData.versionTag` **at enqueue time**. The first write bumps the server versionTag; any write already enqueued with the old tag is rejected. The uploader should resolve the **current** versionTag at **send time** (and/or the outbox should serialize same-file tag writes and refresh between them). Combined with (A), one of every racing pair is lost.

**C. Header (`localAppData.tags`) vs bar (`DriveLocalTagIndex`) diverge.**
The bubble long-press menu reads `MessageUiModel.isPinned` (from the header's `localAppData.tags` via `MessageMapper`) and shows **"Unpin"**, while the bar reads `DriveLocalTagIndex.selectFilesByLocalTagInGroup` (`ChatMessageStream.getPinnedMessages`) and shows **empty** — on the same message, surviving a restart. The two read paths disagree, so the menu and the bar contradict each other. (Likely fallout of the optimistic-write timestamp guard vs. server re-sync interacting with A/B.)

**D. Own-send auto-pin clobber + stale temp fileId.**
Auto-pinning a message you *just sent* (`isPendingSend`) races the create: the optimistic pin is clobbered when the create confirmation re-upserts the file, and the tags row carries the temp fileId that gets rekeyed temp→server (`rekeyCacheAfterCreate`), so `getFileHeader(tempFileId)` 404s. Prior art for the intended fix (resolve by stable `uniqueId` via `getFileHeaderByUid`, update the request fileId) was documented during the original PR work but is **not present** in the current branch. `UpdateLocalMetadataTagsOutboxRequest` currently carries no `uniqueId`.

### Recommended rework (one coherent piece, not more point patches)
1. **Serialize message-pin writes on the localAppData versionTag** — resolve the current versionTag at send time (uploader), or coalesce/queue same-file tag writes and refresh the base tag between them. Fixes B (and makes A's collisions harmless).
2. **Auto-pin must coexist with manual state** — a durable, synced "auto-pin already evaluated / user-dismissed" marker (e.g. a second reserved tag) so `isPinned=false` is distinguishable from "never seen." Fixes A and the re-pin-on-resync/own-device resurrection.
3. **Single source of truth for pin state** — make the bubble menu and the bar read the *same* path (or guarantee the header→`DriveLocalTagIndex` rebuild is consistent under optimistic write + re-sync). Fixes C.
4. **Own-send**: either defer auto-pin to the *confirmed* sighting (real fileId, on server, no clobber — needs the confirmed message to re-arrive via `BatchReceived`, which it appears to), or add `uniqueId` to the request and resolve via `getFileHeaderByUid` at send time. Fixes D.

Net: fixes #1–#5 above are worth keeping (the uploader + hook-wiring ones are essential), but the feature needs items 1–4 before it can ship. Verified on Android; iOS pass still pending.

*(Investigated via device + logcat; the on-device `homebase.log` file-writer buffers and doesn't flush live — use `adb logcat` for real-time app logs.)*

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
```diff
@@ -102,6 +102,12 @@ data class MessageListUiState(
     val isSendingMessage: Boolean = false,
     val pendingOutgoing: ImmutableList<PendingOutgoingMessage> = persistentListOf(),
     val highlightedMessageId: Uuid? = null,
+    /** Pinned-messages bar for the open conversation, newest-first. Empty hides the bar.
+     *  Populated reactively from [id.homebase.chat.services.ChatMessageStream.observePinnedMessages]. */
```
2002Bishwajeet:
```suggestion

```
Delete this

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
```diff
@@ -1440,6 +1467,34 @@ class ConversationListViewModel(
...
+     * #887: one-shot prune of time-expired auto-pins for [conversationId] on open.
+     * Ended events (now past endUtcMs, or startUtcMs + 1h when open-ended) and
+     * stale live-location shares (now ≥ liveShareUntilMs) leave the pinned bar.
+     * LOCAL-ONLY ([unpinMessage] with localOnly = true): each device recomputes
+     * expiry on its own clock, so we don't fan a clock-relative decision out to
+     * the user's other devices via the outbox.
+     */
+    private fun unpinExpiredPins(conversationId: Uuid) {
+        viewModelScope.launch {
+            val now = Clock.System.now().toEpochMilliseconds()
```
2002Bishwajeet, line 1545: Is this thread safe?

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
```diff
@@ -1470,6 +1525,9 @@ class ConversationListViewModel(
             if (convo != null && convo.unreadCount > 0) {
                 frozenUnreadBoundary[conversationId] = convo.lastRead
             }
+            // #887: prune time-expired auto-pins (ended events, stale live shares).
+            // LOCAL-ONLY — each device recomputes on open; not synced via the outbox.
```
2002Bishwajeet: if they are expired we should update it in the server too. unless they are repinned manually again

homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
```diff
@@ -1535,6 +1597,22 @@ class ConversationListViewModel(
                     }
                 }
             }
+            // Pinned-messages bar (#887) — collected in parallel with the message
```
2002Bishwajeet: remove these 3 comment

homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/ConversationContent.kt
```diff
@@ -904,6 +914,13 @@ fun ConversationContent(
                     }
                     .background(MaterialTheme.colorScheme.surfaceContainerLowest)
             ) {
+                // Pinned-messages bar (#887) — first child, hidden when empty.
```
2002Bishwajeet:
```suggestion

```

homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/MessageItem.kt
```diff
@@ -96,6 +96,9 @@ fun MessageItem(
         if (policy.allowShare) remember(message.id) { { onUiAction(ConversationListUiAction.ShareMessage(message)) } } else null
     val onDelete =
         remember(message.id) { { onUiAction(ConversationListUiAction.DeleteMessage(message.id)) } }
+    // Pin/unpin (#887) — delete-style: always available, independent of ActionPolicy.
```
2002Bishwajeet:
```suggestion

```

2002Bishwajeet (commented):
@claude  look at these comments please

PR #955, opened 2026-07-02
Cache bucket 3/3 (PR 2): dedicated HLS playback-chunk cache

Seifert69:
Part 2 of 2 for #845 — stacked on #954. Retarget to `main` after #954 merges; CI runs then. Closes #845 on merge.

## What

HLS playback range-reads were cached per `(chunkStart, chunkLength)` in the shared **200 MB payload LRU** — watching one long video evicted every cached image/attachment, and an image burst evicted warm video chunks mid-playback (S2 in the issue). Chunks now live in a dedicated **`homebase-hls-chunks-v1` cache (100 MB, tracked)**.

**The load-bearing design point:** routing is one request-shaped decision inside `DriveFileProviderCached.getPayloadBytesRaw` (`chunkStart != null` → chunk cache), *not* per-caller — so the **seg-0 preloader**, ExoPlayer's decrypted-range reads, iOS `LocalVideoServer`'s encrypted-range reads, and desktop VLC range reads all converge on the same cache. A per-caller split would make every prefetch a guaranteed playback miss *plus* LRU pollution. All platforms cache the same encrypted range bytes (decryption is post-read), so one cache serves everything. PR 1's 50 MB admission fence also covers open-ended AVPlayer probes (`bytes=0-`) producing file-sized "chunks".

Lifecycle rides existing machinery: `KNOWN_CACHE_DIRS` inclusion gives KEEP-on-startup-sweep / DELETE-on-logout automatically; `clearCaches`/`getCacheStats`/Storage screen gain the third row ("Video stream cache", en+da). `rekeyCachedFile` intentionally ignores chunk entries (they only exist post-sync under the server fileId — documented in code).

## Tests

- Range reads land in the chunk cache; **zero range entries in the payload LRU** (the invariant).
- **Owner-required convergence test**: prefetch seg-0 via `prefetchPayloadChunk`, then read the same range via `getPayloadBytesEncryptedChunk` (the iOS playback shape) → **one network fetch total**.
- Full reads still land in the payload LRU (no leak the other way); warm-chunk re-read = no network; clear/stats (100 MB cap, sentinel isolation)/rekey behavior; `CacheAudit` counts the dir as known; sweeper KEEP/DELETE mode tests.

Verified locally: `jvmTest` + `testDebugUnitTest` (all modules), `:homebase-api:wasmJsTest`, android/wasm compiles.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #845, opened 2026-06-24
Cache bucket 3/3: downloaded payloads stay disposable in cacheDir — stream/size-guard large reads + isolate HLS chunk cap


Seifert69:
## Context
Part of a 3-bucket restructure of payload cache handling (splitting the too-broad #842/#843 by **durability need**):
1. Raw pre-encryption upload temps — disposable.
2. Encrypted-for-outbox staging — durable (the #842 bug).
3. **Downloaded payloads (incl. HLS) — THIS issue — disposable, re-fetchable.**

Guiding rule: a downloaded payload is **re-fetchable from the server, therefore disposable**. Durability is NOT needed here. The improvements below are **orthogonal to durability** — they're about memory and cache-eviction fairness.

## Today
In `cacheDir`:
- encrypted bytes in `homebase-payloads-v2/` (200 MB **LRU**; also where HLS **range-chunks** land, keyed by `(chunkStart, chunkLength)`)
- decrypted playback temps: `hbvid_*/video.mp4`, `hlsdl_*.mp4`, audio temps

**Target: stays in `cacheDir`.** Re-fetchable = disposable; no durable dir.

### 1. Stream / size-guard large payload reads past the LRU (download-side OOM)
Decrypt has two paths: `DriveFileProviderCached.streamPayloadDecryptedToPath()` streams (good), but `getPayloadBytesDecrypted()` / `decryptBytes()` **loads the full payload into memory** — an OOM risk when something reads a large payload as bytes (the mirror of the upload-side OOM in Bucket 2). Route large reads through the streaming-to-file path and/or size-guard the in-memory path so a big download can't blow the heap.

### 2. Isolate the HLS playback-chunk cap (eviction fairness)
HLS range-chunks share the single 200 MB `homebase-payloads-v2/` LRU with everything else. A **download burst can evict an in-flight video's chunks**, stalling playback. Consider giving HLS playback chunks their **own cap/budget** (separate LRU or reservation) so concurrent downloads can't evict the video you're currently watching.

## Cross-platform (required)
- Streaming-read + size-guard behavior consistent on Android/iOS/Desktop/Web actuals (each has its own file/stream + cache dir).
- HLS chunk-cap strategy validated per platform player path.

## Well-tested (required)
- Test: reading a large payload uses bounded memory (streamed), no full-file load / no OOM.
- Test: under a simulated download burst, an in-flight HLS video's chunks are not evicted (if the separate cap is implemented).

## Acceptance criteria
[...]

PR #958, opened 2026-07-02
Fix video progress spinner: complete on first of READY / first-frame / error

Seifert69:
Found while device-testing the #845 builds — but pre-existing on `main` (reproduced identically on build 1947), diagnosed from the device log:

```
13:10:52.992  mp4 first-frame: 284ms
13:10:53.099  player state → READY        ← 100ms later; listener already removed
```

## The race

The per-play `Player.Listener` removed itself in `onRenderedFirstFrame`, but progress completion (`progressJob.cancel()` + `onProgress(1f)`) lived only in the `STATE_READY` handler. On device+codec combos where the first frame renders **before** READY, nobody was left listening:

- **MP4**: bar frozen at exactly **80%** over a playing video.
- **HLS**: spinner frozen at **0%** (warm cache → no download ticks below 1f) over a playing video.
- **Decode error** (e.g. 10-bit AVC High-10 HLG → `NO_EXCEEDS_CAPABILITIES`): READY never comes → spinner forever over a dead player.

## Fix

Both listeners route `STATE_READY` / `onRenderedFirstFrame` / `onPlayerError` through one idempotent `completeProgress(source)` (which logs which signal won, for future triage). Android-only: desktop switches UI state unconditionally; iOS already guards with `AVPlayerItemStatusFailed` + 10s timeout and emits 1f unconditionally.

The **10-bit send-side encode policy** (why that video can't decode at all on this device) is a separate issue to file.

Stacked on #955; retarget after it merges.

https://claude.ai/code/session_01HWmvC3HQ7E1rHaD2ZXFkpC

### DIFF
--- a/homebase-chat/src/androidMain/kotlin/id/homebase/chat/widget/video/VideoPlayerSurface.android.kt
+++ b/homebase-chat/src/androidMain/kotlin/id/homebase/chat/widget/video/VideoPlayerSurface.android.kt
@@ -300,14 +300,28 @@ actual fun VideoPlayerSurface(
                             player.setMediaSource(mediaSource)
                             val prepareStart = TimeSource.Monotonic.markNow()
                             val hlsFirstFrameListener = object : Player.Listener {
-                                private var stateReadyLogged = false
+                                private var progressCompleted = false
+                                // Progress completion fires on WHICHEVER of STATE_READY /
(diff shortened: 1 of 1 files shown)


Seifert69:
Folding this into #955 directly (same file family, avoids a 3-deep stack) — commit cherry-picked onto hls-chunk-cache-845.

Issue #966, opened 2026-07-03
Live location: let recipients share their live location back (reciprocity) + optional dedicated share screen (Signal/WhatsApp-style)


Seifert69:
## Problem / motivation
When a sender shares **live location**, the recipient currently gets a **read-only** view ("X is sharing live location / time left") with **no way to reciprocate**. It should be easy for the recipient to **share their own live location back**. Two directions to decide between (see options), plus a broader UX idea (dedicated share screen).

## Current state (grounded)
- The **"Share live location" offer is gated to the sender's own bubble**: it only renders for `controls.sentByYou && canStart` (`LocationPreview.kt:498`), where `canStart` also requires the pin be <15 min old (`SHARE_OFFER_WINDOW_MS`, `:550`). The recipient never sees an offer.
- The **recipient of a live share** sees only a muted caption — `LiveShareActionArea`, `isLive && !sentByYou` branch is caption-only (`LocationPreview.kt:481-485`); tapping the card opens the map (`onOpenMap`). The **receiver+static** case has **no action area at all** (`// Receiver + STATIC: no action area`, `:544`) — a natural spot for a CTA.
- **Live is bootstrapped off an already-sent static pin**, not chosen at compose time: the attachment sheet only sends a static `MessageContent.Location` (`AttachmentOptions.kt:445-452`, no `liveShareUntilMs`); live is started later via a duration menu on the sender's own bubble (`DURATION_OPTIONS` 15m/30m/1h/2h/4h/24h, `LocationPreview.kt:552-562`).
- **The live map screen** (`LiveLocationScreen.kt`) has only a back button — no reciprocate action.

## Proposed — reciprocity (the core ask), two options to choose from
- **Option A — CTA only when the sender is live-sharing (contextual reciprocity):** on the recipient's view of a LIVE share, show a **"Share your live location"** action — add an actionable branch for `isLive && !controls.sentByYou` in `LiveShareActionArea` (`LocationPreview.kt:481-485`) and/or a button on `LiveLocationScreen`. Reads as "they're sharing with me — here's a one-tap way to share back."
- **Option B — always available (regardless of sender):** everyone can start a live share anytime — surface the duration menu on **received** static bubbles too (fill the empty `// Receiver + STATIC` branch, `:544`) and/or add a live-share entry to the attachment sheet (`AttachmentOptions.kt:445`) that sends a pin already marked live.

[...]

PR #974, opened 2026-07-03
fix(location): share-screen keyboard overlap + blank first-open map (#966)

Seifert69:
Production-test fixes for the #966 share-location screen (per test feedback; all further fixes from this test round will be folded into this PR).

## Fixes

1. **Keyboard overlapped the comment row** — the bottom controls surface (live-share banner + comment/send) now applies `imePadding()`, so tapping the comment field lifts the whole bar above the keyboard.

2. **Map blank on first open** — two combined causes:
   - With no last-known GPS fix, `initialBbox` stayed `null` → `TiledMapView` had no viewport → **no tiles were ever fetched**. Now a whole-world bbox is seeded as the no-fix fallback, so the map always renders and the fresh fix re-fits it when it lands.
   - On a first-ever open the init-time `requestLatestGps` returns `PermissionDenied` because the permission prompt is still up, and nothing retried after the grant. `onPermissionGranted()` now re-requests a fix and seeds the map (unless the user has already panned), so granting the prompt centers the map in place instead of requiring an exit-and-reopen.

## Verification

- `:homebase-core` compiles on JVM/Android/wasmJs.
- Manual: first-install flow (permission prompt up) → map shows world, snaps to fix on grant; airplane-mode cold open → world view, pan + static send still work; comment field stays visible above the keyboard while typing.

### DIFF
--- a/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/share/ShareLocationScreen.kt
+++ b/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/share/ShareLocationScreen.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.consumeWindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.imePadding
 import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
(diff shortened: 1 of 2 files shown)


Seifert69:
Pushed the cold-start fixes from the second test round (screenshots):

- **New centralized cached-coordinate API** — `LocationService.lastKnownCoordinates()`: memory → persisted history → OS last-known fix (survives app kills on Android), newest wins, ANY age, never powers the radio. The share screen seeds from it instantly, then refines with a fresh fix — kill-and-reopen should now land on your last position immediately instead of 0,0-until-GPS-warms.
- **"Half a world map" fixed** — the shared 24-tile budget truncated the world grid top-down, drawing only the top band with the pin floating below it. Truncation is now center-out (tiles nearest the viewport center win), which benefits every map surface.
- **No more bogus 0,0** — on the world fallback with no real position the pin stays unset: send disabled, no geocode, no "0.0, 0.0" chip. Pan manually and everything arms as usual.

New tests: `lastKnownCoordinates` matrix in `LocationServiceTest` (warm-memory short-circuit, cold-start OS-fix hydration, newest-wins, no-permission, no-source) + `MapProjectionTruncationTest` (center-out cut, untouched within-budget grids).

Issue #816, opened 2026-06-22
Show a Homebase-purple location-pin indicator in the top toolbar while actively sharing GPS


Seifert69:
## Problem / motivation
When the client is actively sending GPS / live-location to anyone, there's no always-visible, app-level signal that "you are sharing your location right now." Users can forget a share is running. We want a persistent indicator in the app's top toolbar — a **Homebase-purple location pin** — shown whenever any outgoing live-location share is active, tappable to manage it.

(The OS status bar already shows a system location indicator while GPS is in use, but apps can't add a custom icon there — it's reserved — so this is an in-app indicator.)

## Proposed behaviour
- Whenever ≥1 **outgoing** live-location share is active, show a Homebase-purple location-pin indicator in the top toolbar (app chrome), app-wide.
- When no outgoing share is active, the indicator is hidden.
- It updates live as shares start/stop **and as they expire** (a share ending should clear the pin without a manual refresh).
- **Tapping it** opens the location dashboard (`Route.Location`) — which, per #812, lists who you're sharing with and lets you stop sharing — or the live map (`Route.LocationLive`). Pick the dashboard so the user lands somewhere they can act.

## Placement (lead — confirm with design)
Top bars are currently **per-screen**, not a single shared bar, so there are two viable homes:
1. **Top-bar action (preferred per the request):** add the pin to the action region of the main top bar(s) — e.g. `ConversationListPane` TopAppBar actions (`homebase-chat/.../widget/ConversationListPane.kt` ~222-270, next to Search/Menu). Most faithful to "top toolbar," but would need to be added to each top-level screen's bar (or a shared actions slot) to be truly app-wide.
2. **Global persistent strip (alternative):** the app-level banner column in `AppNavHost.kt` (~599-694, where `UpdateAvailableBanner` / `ConnectionRequestHeaderBanner` already render on top-level screens) — genuinely app-wide with one insertion point, but it's a banner row rather than the toolbar.

Recommend option 1 for the visual the user described; fall back to 2 if a single global insertion is preferred.

## Technical grounding (leads, not a prescription)
[...]

PR #975, opened 2026-07-03
fix(location): remember position across restarts + center-out tile cap (#966)

Seifert69:
Continuation of the #966 production-test fixes — this commit was pushed to the `fix-location-share-screen-966` branch a few minutes **after** #974 was merged, so it needs its own PR (a merged PR doesn't reopen on new pushes). This PR is now the collection point for further fixes from this test round.

## What (second test round: kill-and-reopen opened on a half-drawn world map at 0,0)

- **New centralized cached-coordinate API** — `LocationService.lastKnownCoordinates()`: memory → persisted history row → OS last-known fix (survives app kills on Android), newest wins, ANY age, never powers the radio; hydrates the store for all consumers. Root cause was position amnesia: the in-memory last point is null after a process kill (its DB re-seed helper was never called by anyone) and nothing consulted the OS cache beyond the 30 s freshness window — so the screen waited on the GPS radio. The share screen now seeds from the cached coordinate instantly, then refines with a fresh fix.
- **"Half a world map" fixed** — the shared `MAX_TILES_PER_VIEW` (24) truncated an over-budget grid top-down, drawing only the top band of a portrait world view with the pin floating below it. Truncation is now center-out; benefits every map surface (live map, History, Find Device).
- **No bogus 0,0** — on the world fallback with no real position the pin stays unset: send disabled, no geocode, no "0.0, 0.0" chip; panning arms everything as usual.

## Tests

`LocationServiceTest` matrix for `lastKnownCoordinates` (warm-memory short-circuit, cold-start OS-fix hydration + store publish, newest-of-persisted-vs-OS, no-permission → history only, no-source → null) and `MapProjectionTruncationTest` (center-out cut keeps the four center tiles and drops corners; within-budget grids untouched). Compile green on JVM/Android/wasmJs.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiState.kt
@@ -102,6 +102,13 @@ data class MessageListUiState(
      *  trim discards the newer slice; drives the bottom spinner, the FAB's
      *  ScrollToLatest branch, and gates auto-follow on incoming messages. */
     val hasNewerMessages: Boolean = false,
+    /**
+     * When MY live share fully covers this conversation's recipients: the absolute end of that
+     * coverage, else null. While set (and in the future), location bubbles hide their "share
+     * live location" offers — no invitations to start a share that's already running (#966
+     * follow-up; the app-wide sharing indicator is #816).
+     */
+    val ownLiveShareUntilMs: Long? = null,
     /** A loadOlder fetch is in flight; suppresses re-entry. */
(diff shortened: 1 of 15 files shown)


Seifert69:
Added the "already sharing" fix (option 1 per discussion): while your live share covers the open conversation, all share-live invitations are hidden — the share-back link on received LIVE bubbles, the duration offers on fresh static bubbles (both sides), and the share screen's live banner. Your own live bubble keeps Stop; static send stays available. Coverage requires every recipient of the conversation to have an active relay-roster entry (`liveShareCoverageUntilMs`, unit-tested); the app-wide purple-pin indicator with tap-to-manage remains #816.

Issue #988, opened 2026-07-04
Add per-hop logging to the push→location-capture→upload chain (Android + iOS) so stale-locate failures are diagnosable


Seifert69:
## Problem
When a push-triggered location capture fails to produce a fresh upload (breaking emergency-locate freshness), the logs currently **can't tell us where it broke**. In the incident that prompted this, the recipient's 3.5h log simply had **zero `PushReceived` captures** — which could mean any of: FCM/APNs not delivered, the worker/handler was deferred, `onPushArrived` ran but `isCaptureWanted` was false, the capture ran but the upload never enqueued, or the upload enqueued but never sent. We couldn't distinguish them.

## Ask — per-hop instrumentation on the push→capture→upload chain
Add a log line at **every hop** so a future occurrence is unambiguous, on **both platforms**:
1. **Push received** — Android `DriveFcmService.onMessageReceived` (+ message priority if available); iOS `didReceiveRemoteNotification` / the Notification Service Extension.
2. **Handler started** — Android `DriveSyncWorker.doWork` (already logs start/attempt — extend with "location-capture path"); iOS `onPushArrivedAsync`.
3. **`onPushArrived` entered** (`NotificationEntry`) → about to `forceCaptureIfTracking(PushReceived)`.
4. **`forceCaptureIfTracking`** — log the `isCaptureWanted()` result (why it did/didn't capture) in `LocationService.kt:208-213`.
5. **GPS fix acquired** (or timeout) — `requestLatestGps(PushReceived)` result + source.
6. **Upload enqueued** — `LocationTrackUploaderService.flushIfDue`/`flush` (did it defer on battery? was it rate-gated?).
7. **Upload confirmed** — the hour-file POST success.

Each line should carry the reason tag (`PushReceived`) and enough context (captured? uploaded? deferred-why?) that reading `homebase.log` alone answers "did the push produce a fresh point, and if not, which hop dropped it."

## iOS — must be considered too
Instrument the **iOS** push path with the same hops (`didReceiveRemoteNotification → onPushArrivedAsync → onPushArrived → forceCaptureIfTracking → upload`). iOS Kermit logs land in `homebase.log` (exportable from the device), so this is what lets us diagnose the iOS background-capture gaps (#953) from a real device without a debug build. Include whether the silent push arrived and whether the background upload completed before suspension.

## Affected modules / targets
[...]

Issue #986, opened 2026-07-04
Verify chat message FCM push is sent priority=high (Dozing Android device must wake to capture location)


Seifert69:
## Investigation task (add the finding as a comment when closing)
**Confirm whether the chat message FCM push is sent with `priority: high`.** A Dozing / suspended Android device only *receives* an FCM promptly if the message was sent high-priority; a normal-priority FCM is delayed by Doze, so the device doesn't wake to run the message handler — and the push-triggered location capture never fires.

## Why it matters (evidence)
Cross-referencing two device logs for the emergency-locate "Who you can locate":
- **Recipient (Shelly, Android)** had her app **suspended from ~09:25 to 12:37** (WebSocket/`processInbox` stopped for 3h), and her whole log has **zero `requestLatestGps(PushReceived)` captures**. When a message was sent to her at **12:32**, her device did **not** wake to capture/upload a location point; it only came back at 12:37 when she manually opened the app.
- **Sender (Michael)** correctly queried her drive (`verifyTemporalAccess … hasAccess=true`) and kept reading `newestFileModified=1783156545681` (**09:15:45 UTC**) — so "Who you can locate" showed a stale **~3h** point. The display and query were correct; the recipient just never uploaded a fresh point on the push.

If the FCM isn't high-priority, a suspended device won't wake for the message → no `onPushArrived` → no location capture. (Note: a later test showed the push *did* arrive, so this may be intermittent / Doze-dependent — which is exactly the high-priority question.)

## What to check
- Where the push is sent (likely **odin-core** / the push backend): is the FCM message priority set to **`high`** for chat messages? (Android `priority: "high"`; APNs `apns-priority: 10` for the iOS side.)
- Whether it's conditionally downgraded (e.g. normal priority when the recipient looks "connected").

## Deliverable
Per request: **post the finding as a comment on this issue and close it** — e.g. "FCM priority is/ isn't high; set here: <file/line or odin-core location>", so we know whether the fix is a backend change or whether the delivery gap is elsewhere (see the client-side + logging companions).

## Related
- Client-side companion **#987** (push→capture rides a downgradeable worker; also covers iOS headless upload) and per-hop logging companion **#988** (Android + iOS) are filed for Seifert. Cross-ref #953 (iOS background location).

2002Bishwajeet:
## Finding: FCM priority **is** high — unconditionally, on both platforms

[...]

Issue #953, opened 2026-07-02
iOS background location barely uploads (breaks emergency locate): add headless upload + more background wakes


Seifert69:
## Summary
On **Android**, background location points upload occasionally (OS-batched Fused deliveries cold-wake a dead process; a push enqueues a `DriveSyncWorker`). On **iOS**, in practice almost nothing lands unless the app is open — which starves the **emergency-location** feature (an emergency contact sees only the last point that reached the victim's location drive; on iOS that's "whenever they last opened the app" = stale). Investigate and improve iOS background capture + upload **without** a persistent foreground notification (same constraint as #878).

## What's already wired on iOS (so we build on it, not redo it)
- `allowsBackgroundLocationUpdates = true` + `activityType` (`LocationTracker.apple.kt:62-69`); `UIBackgroundModes: location` + `remote-notification` (`iosApp/iosApp/Info.plist:43-48`).
- **Significant Location Change** monitoring as the relaunch-after-termination vector (`LocationTracker.apple.kt:75`), re-armed on every process start incl. the SLC relaunch (`MainViewController.kt:148-156` → `LocationTrackingCoordinator.onProcessStart()`).
- **Silent-push → one-shot fetch** — the iOS equivalent of Android's `requestLatestGps(PushReceived)` **already exists**: `iOSApp.swift:167-188 didReceiveRemoteNotification → NotificationEntry.kt:66 forceCaptureIfTracking(PushReceived)`. (Depends on the server sending `content-available` silent pushes + iOS honoring them.)

So the CLLocationManager config and the push-capture path are reasonable. The gaps are **(a) reliable background *wake* diversity** and **(b) reliable *headless upload***.

### Gap A — a captured background fix often can't finish uploading (highest impact)
The location POST uses the Ktor **Darwin** client with the **default foreground `URLSession`** (`PlatformHttpClient.native.kt:12-14`) — there is **no** `URLSessionConfiguration.background(withIdentifier:)` anywhere, and **no `beginBackgroundTask`/expiration handler** protecting an in-flight upload. So even when a fix *is* captured (SLC delivery or the ~30 s push window), if iOS suspends the app mid-POST the hour-file upload is cut (rows re-buffer for next cycle, `LocationTrackUploaderService.kt:451-463`). Android gets WorkManager's OS-budgeted, network-constrained, **retryable** upload guarantee (`DriveSyncWorker`); iOS has a bare handler with nothing to keep it alive.

[...]

PR #992, opened 2026-07-04
fix(location): make push/background-wake capture+upload actually land (#987)

### LINKED ISSUE #987: Push→location-capture is unreliable: Android rides a downgradeable worker; iOS needs headless upload — make capture-on-push land on both
## Problem
The push-triggered location capture (`forceCaptureIfTracking(PushReceived)`) that the emergency-locate feature depends on can be **deferred or skipped even when the push is delivered**, because it rides a downgradeable background task. Result: a message to a suspended recipient produces no fresh location upload, and the sender's "Who you can locate" stays stale.

## Evidence
Recipient (Shelly, **Android**) was suspended ~09:25→12:37; a message sent at 12:32 produced **no** `PushReceived` capture (0 across a 3.5h log), and the sender kept reading a **3h-old** `newestFileModified` (09:15:45). The query + display were correct — only the recipient's push→capture→upload failed to run.

## Root cause — Android: the capture rides a non-expedited-fallback worker
`androidApp/.../DriveFcmService.kt:41`:
```kotlin
OneTimeWorkRequestBuilder<DriveSyncWorker>()
    .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
```
When the FCM arrives, the actual work (`DriveSyncWorker.doWork` → `entry.onPushArrived` → `forceCaptureIfTracking(PushReceived)`) is dispatched as **expedited-with-fallback-to-normal**. On a Dozing / low-battery device the expedited quota is quickly exhausted → the work **falls back to normal deferrable work** → the OS can delay it for minutes/hours in Doze → the capture runs late or not at all. The capture gate itself is light (`LocationService.kt:208-213` — just `isCaptureWanted()`), so the failure is *getting the handler to run promptly*, not the capture logic.

**Fix direction (Android):** run the (quick, time-critical) location capture on a more reliable path — e.g. do the capture directly in `DriveFcmService.onMessageReceived` (it's a fast one-shot GPS + enqueue), or use a guaranteed-expedited path for the capture specifically, instead of a worker that can silently downgrade to deferrable. Pair with the backend FCM `priority: high` check (#986) so the push is actually delivered to a Dozing device in the first place.

## iOS side — must be considered too
On iOS the push→capture is **inline** in the ~30s handler (`didReceiveRemoteNotification → onPushArrivedAsync → onPushArrived → forceCaptureIfTracking(PushReceived)`, `iosApp/iOSApp.swift:167-188`) — no WorkManager downgrade — **but** it has its own reliability gaps that produce the same stale outcome:
- The silent push must actually arrive (`content-available`, APNs `apns-priority`), and iOS throttles/suppresses these after force-quit / in Low Power Mode.
- Even when the capture runs, the **upload must complete headless** within the background window — today the location POST uses the default foreground `URLSession` with no `beginBackgroundTask`, so a suspend mid-upload cuts it (tracked in #953).
So for iOS this issue means: ensure the push handler captures **and** the fresh point actually uploads before suspension (background `URLSession` / `beginBackgroundTask`), and that the wake push is delivered. Coordinate with #953.

## Affected modules / targets
`androidApp` (`DriveFcmService.kt` / `DriveSyncWorker.kt` — the capture's execution path), `homebase-common`/`homebase-core` (`LocationService.forceCaptureIfTracking`, the shared `onPushArrived` in `NotificationEntry`), **iOS** (`iosApp/iOSApp.swift` push handler + the headless-upload work from #953). Android + iOS.

## Acceptance criteria
- [ ] A message delivered to a suspended/Dozing **Android** recipient reliably triggers a location capture + upload (not silently deferred by a downgraded worker).
- [ ] On **iOS**, the push handler captures and the fresh point uploads headless within the background window (coordinated with #953).
- [ ] Verified end-to-end on both platforms: after a message, the recipient's newest uploaded point is fresh and the sender's "Who you can locate" updates.
- [ ] No unnecessary battery drain (the capture is a bounded one-shot; respects the existing `isCaptureWanted`/battery gates).

## Scope for the plan
In scope: make the push-triggered location capture+upload reliable on Android (don't ride a downgradeable worker) and on iOS (capture + headless upload in the background window). Out of scope: the FCM/APNs message priority itself (#986 backend) and the broader iOS background-wake work (#953) — coordinate with both. Cross-ref: #986, #953.

Seifert69:
Closes #987.

## The three failure legs (all verified in code, all fixed)

**1. The outbox never drained during a background wake — both platforms, and deeper than the issue's own diagnosis.** `syncIfAuthenticated` never touches the outbox, and `OutboxSync.send()` hard-returns while `!isOnline` — a flag only the *websocket* connect flips, which never happens in a background wake. So even a prompt capture's hour-file just sat in the queue until the next foreground connect.
→ `send(force = true)`: skips **only** the offline early-return, for one drain. Same worker loop, same checkout ordering/dependency gates, same backoff, same events. `isOnline` untouched (WS coordinator owns it); internal re-kicks stay unforced so the bypass never outlives the wake. **Nothing bypasses the outbox — the fix wakes the existing drain at a moment connectivity provably exists.** Other queued items (chat messages stranded offline, older hour-files) get their attempt in the same drain — a bonus, not a side effect.

**2. Universal drain trigger — all wake paths, not just push** (added during review discussion): `LocationTrackUploaderService` gains a `drainNow` seam called after any flush that actually enqueued. Because *every* background point producer funnels through `flush()` — push wakes, Android `LocationUpdatesReceiver` PendingIntent batches into a cold process, iOS SLC relaunch — they're all covered with zero per-path wiring. Online: equivalent to today's kick. Genuinely offline: one fast-failing POST per flush, bounded by the existing 60s rate-gate + normal backoff.

**3. Push path awaits the upload:** new `PushLocationCapture` seam (homebase-common) + `PushCaptureUploader` (homebase-core): gated capture → find the **actual pending hour-file row** (deterministic uid — also catches previously stranded rows; handles "capture served from cache / flush rate-gated ⇒ nothing to await" with a fast return) → forced bounded drain → **row-verified** confirmation. Row-verified because EventBus replay=1 can hand a stale `ItemCompleted` for the same deterministic uid; the row delete precedes the emit, so row-gone is authoritative.
- **iOS**: `NotificationEntry.onPushArrived` awaits it (15s budget) → the fetch `completionHandler` now fires *after* the upload attempt. The #988 log ordering check (`handler: complete` vs `Hour-file upload confirmed`) flips to the good order.
- **Android**: the capture also runs **inline in `DriveFcmService.onMessageReceived`** (8s budget inside the ~10s high-priority FCM wake) — no longer hostage to the expedited worker's silent Doze downgrade. The worker remains the free retry leg (the 60s acquisition/flush gates make its duplicate call a cheap coalesce; its drain only re-attempts if the row is still pending).

## Priority (user decision)
Location hour-files now enqueue at **outbox priority 0** (chat/moments = 1; checkout is `ORDER BY priority ASC`) — an emergency-locate point never queues behind a big media upload during a short wake. Pinned by test.

## Gates respected — no battery bypasses
`isCaptureWanted`, the 60s acquisition guard, the 60s flush rate-gate, and the battery-saver deferral all apply unchanged; the orchestrator returns fast when they gate.

## Tests
- `OutboxSyncTest` +4: forced drain works offline; force never flips `isOnline`; forced failure backs off and does NOT self-retry offline; priority-0 row drains before priority-1.
- New `PushCaptureUploaderTest` (8): gate-closed/no-pending fast exits, happy path, **stale-replay guard**, event-lost-row-gone timeout fallback, ItemFailed early exit, hour-boundary uids, wrong-drive/uid noise ignored.
- Full `homebase-api`/`common`/`core` jvmTest green; Android + wasmJs + `:androidApp:compileDebugKotlin` green.
- ⚠️ iOS Kotlin framework needs a macOS/CI compile (Linux host here). No Swift changes in this PR.

## Manual test (Android, the incident's repro)
1. Recipient backgrounded/Dozing (`adb shell dumpsys deviceidle force-idle`), send them a message → recipient log shows `received:` → `inline-capture: starting` → `capture(PushReceived): fix …` → `drain: confirmed uid=…` within seconds; sender's "Who you can locate" age refreshes on the next verify.
2. Kill network mid-wake → `drain: NOT confirmed … stillPending=true`; the row survives and the worker/next wake drains it.
3. Background tracking without any push (PendingIntent batches) → hour-file uploads now happen in the wake (`send(force=true)` line) instead of stranding until app open.

## Out of scope (coordinated)
FCM/APNs message priority (#986 — the `(DOWNGRADED)` marker from #988 gives the evidence), iOS background URLSession/`beginBackgroundTask` (#953 — the structural fix for uploads outliving the fetch window).

### DIFF
--- a/androidApp/src/main/kotlin/id/homebase/feed/DriveFcmService.kt
+++ b/androidApp/src/main/kotlin/id/homebase/feed/DriveFcmService.kt
@@ -10,12 +10,16 @@ import androidx.work.WorkManager
 import co.touchlab.kermit.Logger
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
+import id.homebase.core.location.PushLocationCapture
 import id.homebase.core.notifications.NotificationService
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeoutOrNull
 import org.koin.android.ext.android.inject
 
 class DriveFcmService : FirebaseMessagingService() {
 
     private val notificationService: NotificationService by inject()
+    private val pushLocationCapture: PushLocationCapture by inject()
 
     override fun onNewToken(token: String) {
         super.onNewToken(token)
@@ -65,6 +69,25 @@ class DriveFcmService : FirebaseMessagingService() {
         Logger.i(tag = "PushCapture") {
             "enqueue: uniqueWork=$WORK_TAG policy=KEEP expedited=downgradable-to-regular"
         }
+
+        // #987: the expedited work request above silently downgrades to DEFERRABLE when the
+        // Doze quota is exhausted — the worker (and its capture) can then run hours late.
+        // onMessageReceived itself executes inside the high-priority FCM wake (~10s budget,
+        // network up), so run the bounded capture + forced outbox drain inline, AFTER the
+        // worker enqueue so chat sync/notification display is never delayed. The worker's
+        // own captureAndUpload call later is a cheap coalesce (60s acquisition guard + 60s
+        // flush rate-gate + replace-enqueue) and the free retry leg if this inline drain
+        // failed. runBlocking is the accepted bridge here — FirebaseMessagingService has no
+        // goAsync, and returning early would drop the process's wake guarantee; a queued
+        // next FCM message waits ≤8s on this thread, which back-to-back pushes absorb via
+        // the last-known/nothing-pending fast paths.
+        Logger.i(tag = "PushCapture") { "inline-capture: starting budgetMs=$INLINE_CAPTURE_BUDGET_MS" }
+        runBlocking {
+            withTimeoutOrNull(INLINE_CAPTURE_BUDGET_MS + BUDGET_SLACK_MS) {
+                runCatching { pushLocationCapture.captureAndUpload(INLINE_CAPTURE_BUDGET_MS) }
+                    .onFailure { Logger.w(tag = "PushCapture", throwable = it) { "inline-capture: failed" } }
+            } ?: Logger.w(tag = "PushCapture") { "inline-capture: hard timeout at ${INLINE_CAPTURE_BUDGET_MS + BUDGET_SLACK_MS}ms" }
+        }
     }
 
     /** FCM priority ints per RemoteMessage: 1=high, 2=normal, 0=unknown. */
@@ -80,5 +103,12 @@ class DriveFcmService : FirebaseMessagingService() {
         const val KEY_BODY = "fcm_body"
         const val KEY_DATA_PREFIX = "fcm_data_"
         const val KEY_ENQUEUED_AT_MS = "fcm_enqueued_at_ms"
+
+        /** High-priority FCM grants ~10s of execution: 5s GPS cap + ~3s drain POST inside
+         *  an 8s budget, 2s process slack (#987). */
+        const val INLINE_CAPTURE_BUDGET_MS = 8_000L
+
+        /** Outer hard-stop margin over the cooperative budget (belt and braces). */
+        const val BUDGET_SLACK_MS = 1_000L
     }
 }
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/OutboxSync.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/OutboxSync.kt
@@ -2,6 +2,7 @@ package id.homebase.api.sync.database
 
 import co.touchlab.kermit.Logger
 import id.homebase.api.client.drives.files.DeleteFilesByGroupIdOutboxRequest
+import id.homebase.api.client.isTransientNetworkFailure
 import id.homebase.api.client.drives.files.DeleteLocalFilesByFileIdRequest
 import id.homebase.api.client.drives.files.DriveOutboxUploader
 import id.homebase.api.client.drives.files.SendReadReceiptByFileIdsOutboxRequest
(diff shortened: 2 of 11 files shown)


Seifert69:
Added per review discussion: **connectivity circuit-breaker** (6fb74822a).

Concern: with the offline gate bypassed, a genuinely offline wake (airplane mode — reachable only via the GPS-batch `drainNow` path, since a push arriving implies network) would have raced through all queued rows, and each hopeless attempt would have charged `checkOutCount` toward the MAX_RETRIES (~48h) permanent-drop budget that offline rows never spent before.

Now: a transport-level failure (`isTransientNetworkFailure` — the same classifier the crash handlers use; HTTP-response errors never match) checks the probed row back in via new `checkInUncharged` SQL (**attempt count untouched** — the '-1 the pop count' semantics, implemented as never-charge; nextRunTime still advances 30s for rate-bounding) and **aborts the rest of the pass**: if row 1 can't reach the network, rows 2..50 can't either. Poison rows (serialization bugs, server rejections) still take the charged path and eventually drop, so no immortal head-of-line blocker.

Pinned by test: two queued rows + `NetworkException` → only the first probed, uncharged, backoff-stamped; second untouched. Full sweep green.

Seifert69:
Third commit (c20f7b963): **parallel backlog drain** — the wake no longer serializes GPS-then-send.

Field evidence from last night's log: 3.8s of `getCurrentFix: acquiring fresh` with the network idle while the outbox waited. GPS radio and POSTs don't contend, so the drain now kicks (`CoroutineStart.UNDISPATCHED`) at wake start: stranded rows upload **during** the GPS wait, and the fresh fix's priority-0 row jumps the running drain's queue when it lands (~4s in).

Also fixes two send-nothing paths: gate-closed (tracking off) and rate-gated-with-no-pending-row previously returned **without draining anything** — stranded chat messages got a push wake that sent nothing. Now every wake drains the backlog unconditionally (offline cost still bounded to one uncharged probe by the circuit breaker).

Capture path byte-identical; staleness thresholds stay fixed per reason (explicitly decided — live sharing already keeps the cache fresh via the continuous tracker's live profile, so a dynamic threshold would rarely change behavior).

New log line to watch for: `drain: backlog kick at wake start` BEFORE `forceCaptureIfTracking`, with `OutboxSync: sending` lines landing inside the `acquiring fresh` window. Tests updated (+ ordering pin `backlogKick_firesBeforeCapture`); full suite green.

Issue #998, opened 2026-07-06
Location "Who you can locate": while offline, keep rows spinning up to 15s waiting for the app to come online before showing the broken cloud


Seifert69:
## Problem / motivation
Expanding "Who you can locate" right after opening the app — **before the app is online** — shows a **broken cloud** per person, because the temporal-verify call fails with no connection. That's technically correct but reads as an error for a transient "not online yet" state. Better: keep the row **spinning for up to ~15 seconds while waiting for the app to come online**, then verify; only show the broken cloud if it's still offline after that (or on a genuine no-access result).

## Clarifying the states (grounded)
The "broken cloud" is specifically the **`Unreachable`** state, not `Broken`:
- **Offline / network error:** `verifyTemporalAccess` **throws** → caught to `null` → `LocateVerifyStatus.Unreachable` → **`Icons.Default.CloudOff`** (the broken cloud) — `LocationViewModel.kt:360-373`, `LocationDashboardContent.kt:691-696`.
- **Genuine no-access** (server answered `!hasAccess`): `LocateVerifyStatus.Broken` → **`Icons.Default.LinkOff`** (broken link) — a *different* icon.
- **Loading:** `CircularProgressIndicator` (the spinner).

So the ask reduces to: **while offline, keep the row in `Loading` (spinner) instead of resolving straight to `Unreachable`.** A genuine `Broken` (no-access) must still show immediately — it's only reached when the server actually responded.

## Current behavior
`verifyLocatablePass` (`LocationViewModel.kt:360-385`) runs the verify per row; an offline throw immediately becomes `Unreachable` (broken cloud). There is **no wait-for-online and no online-triggered retry** — the only retry is the 60s TTL loop (`needsReverify` returns true for error states, #985). So a just-opened, not-yet-connected app flashes broken clouds and only re-checks up to 60s later.

## Proposed change
1. **Inject the online signal.** Add `authConnectionCoordinator: AuthConnectionCoordinator` to `LocationViewModel` (ctor ~52-70) + the DI block (`AppModule.kt` ~876-894, `authConnectionCoordinator = get()`). Use its canonical **`isOnline: StateFlow<Boolean>`** (`AuthConnectionCoordinator.kt:143-153` — "True when the WebSocket is connected AND the handshake completed"). Precedent: `MomentsFeedViewModel`/`ContactBookViewModel` already inject it.
[...]

PR #1041, opened 2026-07-08
fix(location): re-flush on drive mount so iOS background captures upload (#1018)

### LINKED ISSUE #1018: iOS 1.4.1767: background location uploads strand 16h+ — #987 force-drain never fires, outbox stays offline-gated until foreground
**Priority: HIGH.**

## Summary

On iOS **1.4.1767**, a device stopped uploading its location hour-files to its own
identity server for **~16h51m** (2026-07-06 20:28 → 2026-07-07 13:19) even though it was
awake and actively capturing GPS the whole time. The outbox stayed offline-gated in the
background and only drained when the app was brought to the **foreground**. The #987
"drain even without the websocket" force-bypass — which exists specifically to prevent
this — **never fired once** in the entire 10-day log.

**The same background-upload flow is confirmed working on Android in production**, so the
#987 force-drain logic itself is sound. That narrows this to **iOS**: either the shipped
1767 iOS binary doesn't contain the #987 code, or there's an **iOS-specific bug** in the
force-drain path (background/SLC wake). **Bishwa has an iPhone and can reproduce/test this
directly.**

## Environment

- Platform: **iOS** (`AppleLocationTracker`, profile `HistoryBackground`)
- Build: **1.4.1767 (1767)** — confirmed running during the stall (updated 1765→1767 at
  2026-07-06 11:02; the 2026-07-07 09:27 background relaunch logged 1767).
- Feature: live/background location history upload to the user's own drive
  `2e191a14-8640-4ebc-b0c8-aaac913f6fa8`.

## Steps to reproduce (observed timeline)

1. App goes to background in the evening. Last hour-file upload confirmed **2026-07-06
   20:28**. Log then goes silent (iOS suspended the process).
2. Overnight/next morning iOS relaunches the tracker in the background
   (`2026-07-07T09:27:12 (AppleLocationTracker) Started (profile=HistoryBackground)`).
3. GPS keeps flowing and points buffer normally
   (`LocationPointStore Persisted`, `LocationFixRouter Routed 1/1 fixes (last src=gps)`),
   but every outbox attempt logs **`OutboxSync: send() skipped — offline`**, and hour
   flushes log **`Flush NOT enqueued … rows stay buffered, will retry`**.
4. No upload lands until **2026-07-07 13:19**, when the app is foregrounded and the outbox
   finally drains in one catch-up burst.

This is **recurring**, not a one-off — the same signature appears earlier in the log:
2026-07-03 08:30 → 07-05 18:14, and 2026-07-05 22:02 → 07-06 11:01 (~13h).

## Expected vs actual

- **Expected:** background location captures upload to the user's server on a rolling
  basis (minutes), even when the WebSocket isn't connected — this is exactly what #987
  (`drainNow()` → `OutboxSync.send(force = true)`) was added to guarantee.
- **Actual:** uploads strand for 16h+ until the next manual foreground. The user's live
  location on the server is stale for the entire background window.

## Evidence (from the attached `homebase.log`)

**Root-cause signature — the outbox offline gate never lifts in the background:**

<details><summary>homebase.log — during the 16h stall (07-07)</summary>

```text
2026-07-07T09:27:12.569145Z Info: (StartupLogger) Version: 1.4.1767 (1767)
2026-07-07T09:27:12.588608Z Info: (AppleLocationTracker) Started (profile=HistoryBackground)
2026-07-07T11:37:23.075510Z Warn: (LocationTrackUploader) Flush NOT enqueued hour=1783422000000 … rows stay buffered, will retry
2026-07-07T11:37:23.077355Z Debug: OutboxSync: send() skipped — offline
2026-07-07T11:37:23.078699Z Info: (LocationTrackUploader) Flushed hour=1783368000000 uid=09d3c76d-… points=2 stored=2 overflowPayload=false mode=update
2026-07-07T11:37:23.080766Z Warn: (LocationTrackUploader) Flush NOT enqueued hour=1783414800000 … rows stay buffered, will retry
…repeats every few minutes until…
2026-07-07T13:19:xx  ← app foregrounded, outbox drains: many “Hour-file upload confirmed” at once
```
</details>

**The key contradiction:** at `2026-07-07T11:37:23` a flush **did** enqueue
(`Flushed hour=1783368000000 … stored=2`). Per
`LocationTrackUploaderService.kt:216-222`, a successful enqueue (`anyEnqueued == true`)
**must** call `drainNow()` → `OutboxSync.send(force = true)`, which logs
`OutboxSync: send(force=true) — bypassing offline gate (push wake)`
(`OutboxSync.kt:169`). That line — and any `PushCaptureUploader` activity — appears
**0 times** in the entire log. So the force-drain that should have rescued this device is
not executing on the 1767 binary.

Upload-confirmation timeline (gaps = stalls):
```
… 2026-07-06T20:04, 20:28   ← last confirms before the stall
   2026-07-07T13:19         ← next confirm (16h51m later, on foreground)
```

## Suspected area / mechanism (hypothesis — needs the build's SHA to confirm)

The outbox offline gate is `OutboxSync.isOnline` (`OutboxSync.kt:126`, set only by the
connection coordinator, tied to the WebSocket). In an iOS background wake the WS never
connects, so `isOnline` stays `false` and `send()` early-returns "skipped — offline"
(`OutboxSync.kt:164`). #987 added `send(force = true)` to bypass this and wired
`LocationTrackUploaderService.drainNow` to it, called after any successful flush.

The log proves the bypass **isn't running** on 1.4.1767 despite a qualifying enqueue.
Since **Android works in production**, the common `OutboxSync.send(force=true)` /
`LocationTrackUploaderService` logic is not itself broken — this is iOS-specific. Two
candidate explanations, either of which fits:

1. **The 1767 iOS binary does not actually contain #987** (commit `96e7da15b`, landed on
   `main` 2026-07-04 20:26). Build 1765 first launched 2026-07-05 08:45 — only ~16 min
   after #994 merged (07-05 08:29), tighter than an iOS build takes — so 1765 was likely
   cut *before* the merge; 1767 (07-06) *should* include it by date, but iOS
   release/TestFlight builds can lag or be cut from a different branch. **Confirm by
   checking the exact commit SHA the 1767 build was cut from** (is `96e7da15b` an
   ancestor?).
2. If #987 *is* in 1767, then `drainNow` is effectively a no-op on this build/platform
   (e.g. DI wired to the default `{}` , or `send(force=true)` still declining) — a real
   bug in the force-drain path for background/SLC wakes (as opposed to silent-push wakes).

Per CLAUDE.md's root-cause rule: this is a lead backed by the log, not a proven fix. The
first diagnostic step is confirming whether `96e7da15b` is in the 1767 binary.

## Reproduce on-device (Bishwa)

Bishwa has an iPhone — quickest confirmation path:
1. Install 1767 (or a current iOS build), enable location history sharing.
2. Background the app, leave the phone untouched (screen off) for a while so the WS drops
   and iOS suspends/relaunches via SLC.
3. Tail `homebase.log` (`run-as`-equivalent iOS log export) for the background window and
   grep for `send(force=true)` / `bypassing offline gate` vs `send() skipped — offline`.
   - If you see only `send() skipped — offline` after a `Flushed hour=…` line → same bug.
   - If `bypassing offline gate` fires → the shipped 1767 simply lacked the code.

## Scope for the plan

- First: confirm whether the shipped **1767** iOS build contains #987 (`96e7da15b`). If
  **not**, the fix is a build/release-branch issue — ship a build that includes it and
  re-verify against this exact scenario. (Android already ships and works, so this is
  purely about the iOS binary / iOS-specific behaviour.)
- If it **is** in 1767: instrument and fix the iOS `drainNow` → `send(force = true)` path
  so a successful background/SLC-wake enqueue actually attempts an upload while `isOnline
  == false` (expect a visible `send(force=true) — bypassing offline gate` line per flush).
  Since Android is fine, look at iOS-specific differences: background execution
  time/suspension, the DI wiring of `drainNow` on the Darwin target, and whether the
  Ktor/Darwin POST is allowed to run in the background wake.
- Keep the offline gate itself (`isOnline`) owned by the connection coordinator — don't
  flip it; the correct fix is the force-bypass drain, not faking online.
- Add/confirm a regression signal: after a background flush with WS down, exactly one
  forced POST per flush (bounded by the 60s rate-gate), not a silent strand until
  foreground.

_Note: log excerpts are trimmed to timestamps + messages; drive/device UUIDs are the
user's own and contain no names or addresses._

---

## Diagnostic update for @2002Bishwajeet — what the production iOS 1767 log actually shows

Presuming 1767 **does** contain #987 (treat as given for this analysis), the log narrows the
problem in a useful way. Over the full headless window **09:27–13:19** on iOS 1.4.1767:

| Signal | Count |
|---|---|
| `Flushed hour=…` (points enqueued while headless) | **8** |
| `send(force=true)` / `bypassing offline gate` (forced drain fired) | **0** |
| `Hour-file upload confirmed` (anything uploaded) | **0** |
| `drainNow failed` warnings | **0** |

Everything drains only at **13:19**, the instant `promoteToForeground()` runs the deferred
`[onPostAuthenticated, connect, driveRegistry.start, loadProfile]` and the WS comes up.

### The key inference — it's *not* the network layer (yet)

In `LocationTrackUploaderService.flush()`, a successful enqueue sets `anyEnqueued = true`,
and the post-loop block `if (anyEnqueued) drainNow()` (**`LocationTrackUploaderService.kt:216-223`**)
calls `OutboxSync.send(force = true)`, which logs `bypassing offline gate` **synchronously —
before the coroutine launches** (`OutboxSync.kt:169`). We have **8 enqueues and 0 of those
lines.** A POST that started and then froze in the background would *still* have logged the
attempt — so this is **the forced drain not executing at all on the iOS headless flush**,
which is upstream of any background-URLSession concern. Chase "why doesn't the forced drain
fire on the iOS background wake," not "why doesn't the POST complete."

### Two corroborating signals

1. **The pileup is a stranded CREATE.** Repeated
   `OutboxSync: refusing to replace a pending UploadNewFile with an UpdateFile … would strand
   the un-sent create` — the initial hour-file **create** is stuck un-sent in the outbox, so
   every later update flush is refused (`Flush NOT enqueued`). The forced drain is exactly
   what's meant to push that create through.
2. **iOS defers the connection while headless.** `promoteToForeground → deferred=[…, connect,
   …]` — `connect` (what raises `isOnline`) is deferred in the background, so `isOnline`
   never becomes true headless. The only mechanism designed to bypass that offline gate is
   #987's forced drain, which is absent. Also note `observeOutbox()` / `start()` hang off
   `onPostAuthenticated` (deferred headless), so the buffer-drain observer may not even be
   live during a headless wake.

(All of the above are observations from the log + inferences, not a proven root cause. The
build-SHA caveat above still applies — but even presuming #987 is present, the drain isn't
firing.)

## Reproduce & compare on your iPhone (Frodo test identity)

Cheapest way to confirm which of the two failure modes it is — capture your **own** log and
diff it against the production findings above:

1. Sign in on your iPhone as your **Frodo** test identity (`frodo.baggins.demo.rocks`),
   enable location history.
2. Background the app and leave the phone untouched (screen off) long enough for the WS to
   drop and iOS to suspend / SLC-relaunch the process.
3. Export your Frodo `homebase.log` for that background window and grep for:
   - `Flushed hour=` — did points enqueue while headless?
   - `bypassing offline gate` / `send(force=true)` — did the forced drain fire?
   - `Hour-file upload confirmed` — did anything upload?
   - `promoteToForeground` — when did foreground finally run the deferred `connect`?
4. **Compare to the table above:**
   - **Frodo shows `Flushed` > 0 but 0 `bypassing offline gate`** (same as production) →
     confirmed: the forced drain isn't firing on the iOS headless flush. Look at whether
     `flush()`'s `drainNow()` is reached headless and whether the SLC/push wake even routes
     through this path (the `onPostAuthenticated` deferral). **Not** a URLSession bug.
   - **Frodo shows `bypassing offline gate` but no `Hour-file upload confirmed`** → different
     mode: the drain fires but the POST can't complete in the background → background
     URLSession / `beginBackgroundTask` territory (a separate fix).

Either way, your Frodo-vs-production diff is the discriminator that tells us which fix to
build — please attach that log excerpt to this issue.

2002Bishwajeet:
## Problem (part of #1018)

On iOS **headless wakes** (SLC relaunch / emergency-locate push), background location
captures buffer fine but never upload — they strand until the app is next foregrounded.
For an emergency-location grant (Frodo case) this means the recipient's view of the
sharer's position is stale for the entire background window.

## Root cause (upstream of #987's force-drain)

`LocationTrackUploaderService.flush()` starts with a drive-mount gate
(`LocationTrackUploaderService.kt:186`, present since `8d00896d6`, 2026-06-15):

```kotlin
if (!optionalDriveActivation.isActivated(locationLabeledDrive)) { logSkip(...); return }
```

On a cold background wake the GPS fix routes **~300ms before** `OptionalDriveActivation`
finishes mounting the Location drive, so this gate is closed and `flush()` returns
**before it ever enqueues**. And the only background retry — the flush ticker — is
deliberately **foreground-only** (`LocationTrackingCoordinator.kt:137`,
`if (isForeground) startTicker() else stopTicker()`). So the one flush the wake gets is
thrown away and nothing retries → strand until foreground.

This sits **upstream of #987**: the `if (anyEnqueued) drainNow()` force-drain
(`:229`) can't fire because `flush()` returns at `:186` and never enqueues. #987 is
correctly in `main` — it just never gets a turn on this path.

Android is unaffected: its process stays warm between wakes, so the drive is already
mounted and `isActivated` is already true when `flush()` runs.

## Fix

Arm one bounded waiter from the guaranteed-reached skip path that re-flushes the instant
the drive mounts:

```kotlin
armFlushOnMount(reason)   // at the drive-not-mounted return
...
private fun armFlushOnMount(reason: GpsRequestReason?) {
    if (awaitingMount) return                 // one waiter at a time
    awaitingMount = true
    scope.launch {
        try {
            val mounted = withTimeoutOrNull(MOUNT_WAIT_MS) {           // 15s ceiling
                optionalDriveActivation.isActivatedFlow(locationLabeledDrive).first { it }
            }
            if (mounted == true) { logger.i { "Location drive mounted — re-flushing…" }; flush(reason) }
        } finally { awaitingMount = false }
    }
}
```

Once it re-flushes, the existing `#987` `drainNow()` fires normally and the POST goes out
in the background runtime iOS grants the wake (~10s SLC / ~30s push).

- **No-op on Android/foreground** — `isActivated` is already true, so `armFlushOnMount`
  never arms.
- **Idempotent** — `awaitingMount` keeps a single in-flight waiter.
- **Safe** — the re-flush re-checks activation *and* credentials, so a spurious late mount
  can't upload for a logged-out identity; the waiter is a suspended coroutine, free while
  idle and cancelled with the scope if iOS suspends the process first.
- Covers **both** the history-tracker path and the emergency-locate **push** path
  (`PushCaptureUploader`) — they share this `flush()`.

42-line addition, one file, no existing logic modified.

## Scope / honesty

- This closes the **drive-mount-race** half of #1018 (the signature captured on a current
  Frodo build: `flush()` skipping at the drive gate). It does **not** attempt to answer the
  separate build-provenance question in the issue (whether the shipped 1767 binary actually
  contained #987) or any residual background-`URLSession` completion concern — those need
  the on-device SHA / log.
- Not using a `Fixes #1018` auto-close keyword for that reason — leaving the issue open
  until the device capture confirms.

## Verification

- ✅ `:homebase-core:compileKotlinJvm` — clean off `main`
- ✅ `:homebase-core:compileKotlinIosSimulatorArm64` — clean off `main` (pre-existing
  deprecation warnings only)
- ⏳ **Device (pending, the real proof):** build iOS dev → sign in as Frodo, enable
  location history → background, leave untouched so iOS suspends/SLC-relaunches → export
  `homebase.log` and confirm `Location drive mounted — re-flushing buffered points`
  followed by `Flushed hour=…` and a draining `pending` count **in the background**, not
  at the next foreground.

### DIFF
--- a/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/LocationTrackUploaderService.kt
+++ b/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/LocationTrackUploaderService.kt
@@ -220,17 +220,18 @@ class LocationTrackUploaderService(
                     .onFailure { logger.e(it) { "ensureDeviceProfile failed" } }
                 logger.d { "Flush: ${hours.size} pending hour(s), ${finalizeHours.size} closed to finalize reason=${reason ?: "periodic"}" }
             }
-            var anyEnqueued = false
+            val outcomes = ArrayList<HourFlushOutcome>(hours.size + finalizeHours.size)
             for (hourBucket in hours + finalizeHours) {
                 runCatching { flushHour(hourBucket * HOUR_MS) }
-                    .onSuccess { enqueued -> anyEnqueued = anyEnqueued || enqueued }
+                    .onSuccess { outcomes.add(it) }
                     .onFailure { logger.e(it) { "flushHour($hourBucket) failed" } }
             }
-            if (anyEnqueued) {
+            if (shouldKickDrain(outcomes)) {
                 // Drain even without the websocket (#987): background wakes (push,
                 // PendingIntent batch, SLC relaunch) never connect the WS, so the normal
                 // enqueue kick declines offline and the row would strand until the next
                 // foreground connect. Equivalent to the normal kick when online.
+                // Refused hours kick too — see [shouldKickDrain].
                 runCatching { drainNow() }
                     .onFailure { logger.w(it) { "drainNow failed after flush" } }
             }
@@ -263,10 +264,9 @@ class LocationTrackUploaderService(
         }
     }
 
-    /** @return true when the hour was handed to the outbox (created or update-coalesced). */
-    private suspend fun flushHour(hourStartMs: Long): Boolean {
+    private suspend fun flushHour(hourStartMs: Long): HourFlushOutcome {
         val points = buffer.selectByTimeRange(hourStartMs, hourStartMs + HOUR_MS)
-        if (points.isEmpty()) return false
+        if (points.isEmpty()) return HourFlushOutcome.NoRows
 
         val uid = locationHourFileUid(deviceId.value, hourStartMs)
         val (headerJson, stored) = LocationTrackCodec.encodeHeader(deviceId.value, hourStartMs, points)
@@ -295,7 +295,7 @@ class LocationTrackUploaderService(
                     "mode=${if (existing == null) "create" else "update"} — rows stay buffered, will retry"
             }
         }
-        return enqueued
+        return if (enqueued) HourFlushOutcome.Enqueued else HourFlushOutcome.Refused
     }
 
     /**
@@ -587,6 +587,33 @@ internal fun shouldDeferBackgroundFlush(
     appForeground: Boolean,
     hasStaleUnuploaded: Boolean,
 ): Boolean = powerSaveMode && !appForeground && !hasStaleUnuploaded
+
+/** Outcome of flushing one hour bucket — the input to [shouldKickDrain]. */
+internal enum class HourFlushOutcome {
+    /** Handed to the outbox (created or update-coalesced). */
+    Enqueued,
+
+    /** The outbox declined the enqueue (e.g. replaceEnqueue's WouldStrandCreate guard) —
+     *  a pending row already in the outbox blocks this hour; rows stay buffered. */
+    Refused,
+
+    /** No rows for the hour (raced away between the hour scan and the flush). */
+    NoRows,
+}
+
+/**
+ * Whether [LocationTrackUploaderService.flush] should kick the forced outbox drain (#987).
+ * Enqueued is the obvious case. Refused kicks too: a refusal means a pending row is ALREADY
+ * sitting in the outbox blocking this hour (e.g. an un-sent create that an update may not
+ * replace) — and that blocking row is exactly what a forced drain sends. On a background wake
+ * with no push (iOS SLC relaunch, Android PendingIntent cold start) nothing else kicks the
+ * outbox, so gating the drain on Enqueued alone strands the blocking row until the next
+ * foreground connect — the #1018 production pileup ("refusing to replace a pending
+ * UploadNewFile" every few minutes for 16h). Cost of the extra kick is the same bounded
+ * fast-failing attempt as the Enqueued case. Pure for unit testing.
+ */
+internal fun shouldKickDrain(outcomes: List<HourFlushOutcome>): Boolean =
+    outcomes.any { it == HourFlushOutcome.Enqueued || it == HourFlushOutcome.Refused }
 
 /** What an outbox event means for the location point buffer. */
 internal enum class BufferAction {
--- a/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/LocationTrackUploaderService.kt
+++ b/homebase-core/src/commonMain/kotlin/id/homebase/core/ui/screens/location/LocationTrackUploaderService.kt
@@ -39,9 +39,11 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withTimeoutOrNull
 import kotlin.time.Clock
 import kotlin.uuid.Uuid
 
@@ -101,6 +103,10 @@ class LocationTrackUploaderService(
     private var lastFlushAttemptMs = 0L
     private var observerStarted = false
     private var deviceProfileEnsured = false
+    // One in-flight "wait for the drive to mount, then re-flush" waiter at a time. Set under
+    // flushMutex, cleared by the waiter itself. A benign re-arm race (the waiter clears it just as
+    // a fresh skip sets it) at worst launches a second idempotent waiter — not worth a lock.
+    private var awaitingMount = false
 
     private val _lastFlushTime = MutableStateFlow<Long?>(null)
     val lastFlushTime: StateFlow<Long?> = _lastFlushTime.asStateFlow()
@@ -179,6 +185,13 @@ class LocationTrackUploaderService(
             lastFlushAttemptMs = now
             if (!optionalDriveActivation.isActivated(locationLabeledDrive)) {
                 logSkip(reason) { "Flush skipped: Location add-on not activated (drive not mounted) reason=${reason ?: "periodic"}" }
+                // iOS-only race: a cold background wake (SLC relaunch / emergency-locate push) routes
+                // its GPS fix ~300ms BEFORE the drive-mount pipeline finishes, so this flush loses to
+                // the mount and — with the foreground-only ticker stopped — nothing retries and the
+                // point strands until the next foreground session. (Android's process stays warm, so
+                // its drive is already mounted and it never hits this.) Re-flush the instant the drive
+                // mounts. Harmless on Android/foreground: isActivated is already true, so we never arm.
+                armFlushOnMount(reason)
                 return
             }
             if (credentialsManager.getActiveCredentials() == null) {
@@ -222,6 +235,31 @@ class LocationTrackUploaderService(
                     .onFailure { logger.w(it) { "drainNow failed after flush" } }
             }
             refreshPendingCount()
+        }
+    }
+
+    /**
+     * Re-flush once the Location drive mounts, bounded by [MOUNT_WAIT_MS]. Closes the cold-wake
+     * race where a fix's flush ran before the mount pipeline finished (see the call site). The wait
+     * is a suspended coroutine on the app scope — it costs nothing while idle and is cancelled with
+     * the scope if the wake ends first (iOS suspends the process); the flush re-checks activation and
+     * credentials, so a spurious late mount can't upload for a logged-out identity.
+     */
+    private fun armFlushOnMount(reason: GpsRequestReason?) {
+        if (awaitingMount) return
+        awaitingMount = true
+        scope.launch {
+            try {
+                val mounted = withTimeoutOrNull(MOUNT_WAIT_MS) {
+                    optionalDriveActivation.isActivatedFlow(locationLabeledDrive).first { it }
+                }
+                if (mounted == true) {
+                    logger.i { "Location drive mounted — re-flushing buffered points reason=${reason ?: "periodic"}" }
+                    flush(reason)
+                }
+            } finally {
+                awaitingMount = false
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Seifert69:
## #1018's open provenance question is now resolved — the production 1767 binary never contained #987

iOS build number = `version.code.base` (1505) + the `build-mobile-release-prod.yml` run number, so **build 1767 = prod run 262 = `b8f400b05`** (merge of #979), built **2026-07-04 07:37** — ~13h *before* #987 merged (`96e7da15b`, 2026-07-04 20:26). `git merge-base --is-ancestor` confirms neither 1765 nor 1767 contains #987. That fully explains the production log's "8 enqueues, 0 `bypassing offline gate` lines": the force-drain code simply wasn't in the binary. Nobody needs to keep debugging why #987 "didn't fire" on 1767 — it couldn't.

That makes this PR's mount-race fix the right scope for **current** builds: it closes a real hole that would have partially defeated #987 on cold background wakes even in a build that has it. The mechanism looks sound — it waits for the mount that `AuthConnectionCoordinator`'s Authenticated branch already performs unconditionally *before* the headless early-return (`AuthConnectionCoordinator.kt:194-236`), rather than faking any state; bounded wait, single waiter, re-checks activation + credentials on re-flush.

One narrative nit: **"Android is unaffected: its process stays warm" is slightly overstated** — #987 explicitly covers Android `PendingIntent` batches into a *cold* process, where the same mount race exists. No code change needed (the fix is commonMain and covers it), but the "iOS-only race" comment reads narrower than the fix actually is.

Remaining before closing #1018: the on-device Frodo capture this PR already lists as pending — with provenance resolved it's now the only proof of what bites on a current build — plus shipping a build with #987 + #998 + this to the reporting device and re-checking the original signature.

Seifert69:
## Companion commit pushed to this branch: `a12b77a60` — drain also kicks on *refused* enqueues

Separate change from the mount-race fix above, closing the **other** strand visible in the #1018 production log.

**The gap:** `flush()` gated the #987 forced drain on `anyEnqueued` — but a flush whose hours are all **refused** (`replaceEnqueue`'s `WouldStrandCreate` guard: a pending un-sent create blocks the hour's update → "Flush NOT enqueued") left `anyEnqueued = false` and never kicked the drain. A refusal means a pending row is *already sitting in the outbox*, and that blocking row is exactly what the forced drain sends. On a pushless background wake (iOS SLC relaunch, Android `PendingIntent` cold start) nothing else kicks the outbox — `PushCaptureUploader`'s backlog kick only runs on push wakes — so the stranded create sat until the next foreground connect. The production log shows this exact signature: "refusing to replace a pending UploadNewFile" every few minutes across the 16h stall.

**The change** (`LocationTrackUploaderService.kt`):
- `flushHour` now returns a tri-state `HourFlushOutcome` — `Enqueued` / `Refused` / `NoRows` — instead of a boolean.
- The drain decision is the pure `shouldKickDrain(outcomes)`: kick on any `Enqueued` **or** `Refused`; stay quiet when there was nothing to send. Cost of the extra kick is the same bounded fast-failing attempt as the existing one; `isOnline` ownership untouched.

**Tests:** `LocationDrainKickTest` (jvmTest, 5 cases) following the same pure-function pattern as `LocationFlushDeferralTest` / `LocationBufferActionTest`. `:homebase-core:compileKotlinJvm` + `compileAndroidMain` clean, location suite green locally.

PR #1082, opened 2026-07-13
fix(chat): sending clock (not Sent check) for stuck-outgoing last message in list (#1076)

### LINKED ISSUE #1076: Conversation list shows single-check (Sent) for a message still stuck outgoing in the outbox
## Symptom

On a very spotty network, I sent a 1:1 message that is **stuck outgoing** (still in the outbox, not yet sent). In the **conversation overview (list)**, that conversation's row shows the last message with a **single check (Sent)** — which is wrong; it hasn't been sent.

## Root cause

The conversation-list row renders the delivery indicator with a **hardcoded `isPendingSend = false`** and only the stored status Int:

```
// ConversationItem.kt:223-229
} else if (lastMessageIsFromActiveUser && lastMessageDeliveryStatus != null) {
    DeliveryStatus(
        isPendingSend = false,                       // <-- always false
        deliveryStatus = lastMessageDeliveryStatus,  // <-- MessageAppData.deliveryStatus
        ...
    )
}
```

`DeliveryStatus` only shows the "sending" (clock/alarm) state when `isPendingSend == true` (`MessageBubble.kt:876`); otherwise it maps the `deliveryStatus` Int to a tick. And **`MessageAppData.deliveryStatus` defaults to Sent** (documented at `MessageInfoUiState.kt:164`: "`MessageAppData.deliveryStatus` defaults to Sent"). So a message still sitting in the outbox — never actually sent — is rendered with the Sent single-check in the list.

The message **bubble** and message-info screen get this right because they compute `isPendingSend` from whether a live outbox row still exists (see the `isPendingSend`/`ServerPresence` logic in `MessageInfoUiState.kt`). The **list row has no such signal** — `ConversationUiModel` doesn't carry a "last message is pending send" flag, so `ConversationItem` can only pass `false`.

## Fix

Give the conversation list the same pending-send awareness the bubble has:

1. Add a `lastMessageIsPendingSend: Boolean` to `ConversationUiModel`, derived from whether the last message still has a live outbox row (pending send) — mirroring the bubble/message-info source of truth.
2. Pass it through in `ConversationItem.kt:226` as `isPendingSend = enrichedData.conversation.lastMessageIsPendingSend`.

Then a stuck-outgoing last message shows the sending/clock indicator in the overview, and only flips to the single check once it actually settles on the server.

## Files
- `homebase-chat/.../widget/ConversationItem.kt:223-229` — hardcoded `isPendingSend = false`
- `homebase-chat/.../data/ConversationUiModel.kt` — add `lastMessageIsPendingSend` (currently only `lastMessageDeliveryStatus`, which defaults to Sent)
- `homebase-chat/.../messageinfo/MessageInfoUiState.kt` — existing pending-send / ServerPresence logic to mirror
- `homebase-chat/.../widget/MessageBubble.kt:868-883` — `DeliveryStatus(isPendingSend=…)` "sending" branch

_Repro: send while offline/very slow so the message stays in the outbox, then look at the conversation list row. Identities redacted; Android build under airplane wifi._

2002Bishwajeet:
Closes #1076.

## Problem

On a spotty network, a 1:1 message stuck in the outbox (never sent) showed a **single check (Sent)** in the conversation-list row. The row hardcoded `isPendingSend = false` and rendered the tick purely from `lastMessageDeliveryStatus` — which **defaults to Sent** (`MessageAppData.deliveryStatus`). The bubble and message-info screen get this right because they read `MessageUiModel.isPendingSend`; the list model carried no such flag.

## Fix

Give the list the same pending-send signal the bubble uses:

- Add `lastMessageIsPendingSend: Boolean` to `ConversationUiModel`, set from `MessageUiModel.isPendingSend` at all three preview-write sites — cold-load enrichment (`updateWithLatestMessage`), live arrival (`applyIncomingMessageBump.withPreviewOf`), and the orphan placeholder in `ConversationStream` — right next to `lastMessageDeliveryStatus`.
- Pass it into `DeliveryStatus(isPendingSend = …)` in `ConversationItem`, replacing the hardcoded `false`.

The flag is written in the same statement as `lastMessageDeliveryStatus` everywhere, so the two stay coupled. The `sqlUserDate == existing.latestMessageTimestamp` re-emit branch in `applyIncomingMessageBump` already refreshes the preview on a delivery-status change, so **the clock self-clears to a check once the send settles** — no stale-clock risk.

## Verification

- `:homebase-chat:compileKotlinJvm` + `:homebase-chat:compileAndroidMain` — BUILD SUCCESSFUL.
- `:homebase-chat:jvmTest` green, including a new regression case `propagates_isPendingSend_forStuckOutgoingLastMessage` in `ConversationUiModelUpdateWithLatestMessageTest`.
- Device repro (send under airplane/slow wifi so the message stays in the outbox, then view the list) recommended before merge.

---

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
@@ -34,12 +34,6 @@ data class ConversationUiModel(
     val dirty: Boolean = false,
     val avatarModel: ConversationAvatarModel,
     val lastMessageDeliveryStatus: Int? = null,
-    /**
-     * True when the last message is still sitting in the outbox (pending send) —
-     * the same [MessageUiModel.isPendingSend] signal the bubble uses. Without it
-     * the list would map [lastMessageDeliveryStatus] (which defaults to Sent) to a
-     * single check for a message that never actually sent. See issue #1076.
-     */
     val lastMessageIsPendingSend: Boolean = false,
     val lastMessageIsDeleted: Boolean = false,
     val lastMessageFirstPayload: PayloadDescriptor? = null,
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
@@ -34,6 +34,13 @@ data class ConversationUiModel(
     val dirty: Boolean = false,
     val avatarModel: ConversationAvatarModel,
     val lastMessageDeliveryStatus: Int? = null,
+    /**
+     * True when the last message is still sitting in the outbox (pending send) —
+     * the same [MessageUiModel.isPendingSend] signal the bubble uses. Without it
+     * the list would map [lastMessageDeliveryStatus] (which defaults to Sent) to a
+     * single check for a message that never actually sent. See issue #1076.
+     */
+    val lastMessageIsPendingSend: Boolean = false,
     val lastMessageIsDeleted: Boolean = false,
     val lastMessageFirstPayload: PayloadDescriptor? = null,
     val lastMessageHasMultiplePayloads: Boolean = false,
@@ -175,6 +182,7 @@ data class ConversationUiModel(
                     lastMessage = msg.content.truncateToCodePoints(40),
                     latestMessageTimestamp = candidate,
                     lastMessageDeliveryStatus = msg.messageAppData.deliveryStatus,
+                    lastMessageIsPendingSend = msg.isPendingSend,
                     lastMessageIsDeleted = msg.isDeleted,
                     lastMessageFirstPayload = msg.payloads?.firstOrNull(),
                     lastMessageHasMultiplePayloads = (msg.payloads?.size ?: 0) > 1,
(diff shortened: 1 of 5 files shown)


homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
```diff
@@ -34,6 +34,13 @@ data class ConversationUiModel(
     val dirty: Boolean = false,
     val avatarModel: ConversationAvatarModel,
     val lastMessageDeliveryStatus: Int? = null,
+    /**
+     * True when the last message is still sitting in the outbox (pending send) —
```
2002Bishwajeet: remove this comments. they are not needed

homebase-chat/src/jvmTest/kotlin/id/homebase/chat/data/ConversationUiModelUpdateWithLatestMessageTest.kt
```diff
@@ -120,4 +120,16 @@ class ConversationUiModelUpdateWithLatestMessageTest {
         // Equality not required; the contract is "no regression".
         assertEquals(ahead, result.latestMessageTimestamp.toEpochMilliseconds())
     }
+
+    /** #1076: a last message still stuck in the outbox must carry its pending-send
```
2002Bishwajeet: same here

PR #1100, opened 2026-07-15
feat(api): add DriveQueryProvider.queryBatchCollection

toddmitchell:
## Summary
- Adds a client for the V2 `POST /drives/query-batch-collection` endpoint, letting callers run multiple named query-batch sections (each with its own drive/filters/paging options) in a single round trip.
- Refactors `DriveQueryProvider`'s per-file salvage-decode (bad/corrupt `fileMetadata` degrades to a placeholder instead of failing the batch) into a shared helper so both `queryBatch` and the new `queryBatchCollection` use identical decode behavior.
- New request/response types: `QueryBatchCollectionRequest`/`CollectionQueryParamSection` and `QueryBatchCollectionResponse`, reusing the existing `FileQueryParams`, `QueryBatchResultOptionsRequest`, and `QueryBatchResponse` types.

No call sites are wired up yet — this only adds the provider method.

## Test plan
- [x] `./gradlew :homebase-api:compileKotlinJvm`
- [x] `./gradlew :homebase-api:jvmTest --tests "id.homebase.api.client.drives.query.DriveQueryProviderQueryBatchCollectionTest"` — pins the request shape (no `driveId` path segment, per-section fields round-trip) and per-section response decode (healthy section reuses the salvage-decode path; `invalidDrive` section comes back empty without failing the collection).

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/QueryBatchCollectionRequest.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/QueryBatchCollectionRequest.kt
@@ -1,0 +1,29 @@
+package id.homebase.api.client.drives
+
+import id.homebase.api.client.drives.query.FileQueryParams
+import id.homebase.api.serialization.UuidSerializer
+import kotlinx.serialization.Serializable
+import kotlin.uuid.Uuid
+
+/**
+ * Request body for POST /drives/query-batch-collection — runs multiple named
+ * query-batch sections against one or more drives in a single round trip.
+ */
+@Serializable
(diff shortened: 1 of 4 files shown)


Seifert69:
Note for the wire-up (tracked in #1102): when one drive paginates but the other 6 finish in the first round, **shrink the collection each round** — drop finished sections, carry only the unfinished section(s) (with advanced cursors) into the next `queryBatchCollection`, until all are exhausted; don't re-query finished drives. On **killroy** (a change lands mid-sync), **re-run a full collection over all 7** — finished drives must be treated as potentially-dirty again. Detail in #1102.

Issue #921, opened 2026-07-01
Incoming connection request: hide meaningless pre-connection sections, show public profile (+ optional circle-picker on Accept)


Seifert69:
## Problem / motivation
Opening an **incoming connection request** ("Wants to connect with you", Accept/Reject) reuses the shared **`ContactDetailScreen`**, which then shows sections that are **meaningless before you're connected**:
- **Contact details: None**
- **Groups in common:** "Connect with this person to see groups you have in common."
- **Circles:** "Connect with this person to add them to your circles."
- Tabs **Details / Activity / About** — all effectively empty (Activity needs an existing conversation; About needs synced data — neither exists pre-connection).

So the whole detail body is placeholder text. For a stranger asking to connect, what's useful is **their public profile** — and, optionally, choosing **which of your circles** to put them in when you accept.

*(Screenshot not attached — it shows a profile photo/face.)*

## Root cause (grounding)
- The request state is `uiState.requestDirection == RequestDirection.INCOMING` (`ContactDetailViewModel.kt:168-176`), which drives "Wants to connect with you" + Accept/Reject (`ContactDetailScreen.kt:619-637`).
- **Sections aren't gated on connection state.** DETAILS tab (`ContactDetailScreen.kt:280-300`) renders `ContactFieldsSection` (ungated → "None") + `GroupsInCommonSection` + `CirclesSection` gated only on `hasOdinId` — which is true for a requester. Their not-connected branches show the "Connect with this person to…" hints (`contactbook_detail_groups_connect`, `contactbook_detail_circles_connect`).
- **All three tabs always show** (`ContactDetailScreen.kt:407-412, 230-236`); About = synced `ext_data` (empty pre-connection), Activity = 1:1 conversation overview (no conversation yet → empty).
- **Public profile IS fetchable pre-connection:** `PublicIdentityRepository.resolve(odinId)` (name, given/surname, status) + `fetchSiteData(odinId)`/photo section from `sitedata.json`, no connection required — the same call Add Contact uses (`AddContactViewModel.kt:190`). `ContactDetailViewModel` doesn't inject it today.

## Part A — hide the meaningless sections, show the public profile (primary)
When `requestDirection == INCOMING` and `!isConnected`:
[...]

PR #1134, opened 2026-07-22
feat(contactbook): pending-request public profile + circle picker on Accept (#921)

toddmitchell:
Part A of #921. Part B (circle-picker on Accept) intentionally left as a fast-follow, per the issue author's recommendation to ship A first.

## Problem
Opening an incoming connection request reuses `ContactDetailScreen`, which before connecting shows only placeholders: "Contact details: None", the "connect to see groups/circles" hints, and empty Activity/About tabs. None of it helps you decide whether to accept.

## Fix (Part A)
For a pending incoming request (`requestDirection == INCOMING && !isConnected`):
- **Collapse to the single Details tab** — Activity needs a 1:1 conversation and About needs synced `ext_data`, neither of which exists pre-connection, so both tabs were always empty.
- **Hide the connection-scoped sections** in Details (contact fields, groups-in-common, circles), leaving the header's public profile to stand alone.
- **Resolve and show the requester's public profile**: `ContactDetailViewModel` now injects `PublicIdentityRepository` and reads the requester's `sitedata.json` (the same pre-connection lookup Add Contact uses), overlaying name/status onto the otherwise-synthetic entry. Best-effort — wrapped in `runCatching`, threaded through the existing state combine so it survives re-emissions, and it never gates Accept/Reject.

Once connected, the full detail (fields, groups-in-common, circles, all three tabs) renders exactly as before.

## Acceptance criteria
- [x] Pending incoming request no longer shows "Contact details: None", groups-in-common, circles, or the empty Activity/About tabs
- [x] Pending request shows the requester's public profile (name, status) to inform Accept/Reject
- [x] Once connected, full contact detail renders as before
- [ ] Verified on Android + iOS (UI-only change; not device-tested here)

## Notes / follow-ups
- **Photo:** this surfaces name + status via the existing avatar (initials from the resolved name), not the actual profile *image*. The image needs a separate profile-image drive read (`photo` sitedata section + payload fetch) — larger than Part A warranted; happy to add if wanted.
- **Part B** (choose which of your own circles to add the contact to, on Accept; wire the unused `ConnectionNetworkProvider.addToCircle`) stays open on #921.

## Testing
- [x] `:homebase-core:compileKotlinJvm`, `:homebase-core:compileKotlinIosSimulatorArm64`, `:homebase-core:jvmTest` pass

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/connections/ConnectionRequestProvider.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/connections/ConnectionRequestProvider.kt
@@ -177,7 +177,8 @@ class ConnectionRequestProvider(
     // ------------------------------------------------------------
 
     suspend fun acceptIncomingRequest(
-        senderId: OdinId
+        senderId: OdinId,
+        request: AcceptConnectionRequestV2
     ) {
 
         val creds = requireCreds()
@@ -188,7 +189,7 @@ class ConnectionRequestProvider(
         val response = encryptedPutJson(
             url = apiUrl(creds.domain, endpoint),
             token = creds.accessToken,
-            jsonBody = "{}",   // API expects no body
+            jsonBody = OdinSystemSerializer.serialize(request),
             secret = creds.secret
         )
 
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/connections/ProviderTypes.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/connections/ProviderTypes.kt
@@ -24,6 +24,17 @@ data class AddCircleMembershipRequest(
     val circleId: Uuid
 )
 
+/**
+ * Body for the V2 accept-incoming-request endpoint: the circles the accepting user places the
+ * new connection into, atomically with the accept. Serialized via [id.homebase.api.serialization.OdinSystemSerializer]
+ * (camelCase `circleIds`); the backend's `AcceptConnectionRequestV2.CircleIds` binds it
+ * case-insensitively. Empty list = accept without adding to any circle.
+ */
+@Serializable
+data class AcceptConnectionRequestV2(
+    val circleIds: List<Uuid> = emptyList()
+)
+
 @Serializable
 data class RevokeCircleMembershipRequest(
     val odinId: OdinId,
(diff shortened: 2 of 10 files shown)


toddmitchell:
Refactored per review: the pending state now lives in a single self-contained **`PendingRequestProfile`** component, branched once at the top-level `when()` in `ContactDetailScreen`, rather than scattering `isPendingIncoming` conditionals through the connected view's DETAILS `when`-block and tab list (those are back to exactly what they were).

Also now shows the **profile photo** (public `/pub/image` via `ContactBookAvatar` → `PublicAvatar` — resolves pre-connection) and a **short bio summary** (new: parsed from the `short-bio-summary` sitedata section, key `short_bio`, added to `PublicIdentity`). Top-bar Edit/management menu are hidden for a pending request. Accept flips this same screen to the full connected detail in place.

Still: dedicated component, not a dedicated screen/route — the Accept action reuses the shared screen and flips in place.

Compiles JVM + iOS; `homebase-core:jvmTest` + `homebase-api:jvmTest` pass. Still needs on-device Android/iOS verification.

toddmitchell:
Part B (circle picker on Accept) is now included on this branch, so this PR closes **all of #921**.

**Data layer** — accept now carries the circles atomically via the new backend `AcceptConnectionRequestV2 { CircleIds }`:
- `ConnectionRequestProvider.acceptIncomingRequest(senderId, AcceptConnectionRequestV2)` PUTs `{circleIds:[...]}` instead of an empty body (serialized via `OdinSystemSerializer`; `UuidSerializer` emits standard GUIDs the backend binds case-insensitively).
- `ConnectionRequestService.acceptIncomingRequest` gained a `circleIds: List<Uuid> = emptyList()` param — the default keeps the other caller (`AddContactViewModel`) and the no-selection path unchanged.

**UI / VM**:
- `ContactDetailViewModel` exposes `assignableCircles` (all user-defined circles, same system-circle exclusion as the contact's own chips — Confirmed/Auto dropped, Emergency kept) and a new `AcceptRequestWithCircles` action that parses the 32-char N-format ids to `Uuid` and forwards them.
- `PendingRequestProfile` shows the circles as inline `FilterChip`s above Accept/Reject (per the chosen design); the ticked set rides the Accept. Ids that fail to parse are dropped rather than aborting the accept.

**Endpoint note:** kept `PUT /connections/requests/incoming/{senderId}` (now with a body). If the V2 contract lives at a different route, say so and I'll update the URL.

Compiles JVM + iOS; `homebase-core`/`homebase-chat`/`homebase-api`/`homebase-common` JVM tests pass. Still needs on-device Android/iOS verification.

Issue #959, opened 2026-07-02
10-bit HLG in-app capture bypasses the 8-bit pin via probe fail-open (toggle OFF) — ships AVC High-10 most devices can't decode


Seifert69:
## Summary
A video recorded with the in-app camera (from the full-screen attachment editor / content bubble) on a Pixel 8 shipped as **AVC High-10 10-bit HLG** (`avc1.6E0028`) — a profile most Android hardware decoders reject — **with the dev-menu "allow 10-bit video" toggle OFF on the sending device**. Receivers get `ERROR_CODE_DECODING_FAILED / NO_EXCEEDS_CAPABILITIES` and a dead player with no visible error.

The 8-bit-pin *policy* is correct and was not overridden — it was **bypassed by a probe hole**: the already-optimal pass-through never learned the source was 10-bit HDR.

## Root cause (verified in code; the pass-through leg is the lead to confirm on-device)
`FfmpegCompressPlanner.isAlreadyOptimal` correctly disqualifies pass-through for `bitDepth > 8` and for `isHdr` — but both inputs come from `FFmpegUtils.android.kt`'s `MediaExtractor` probe, which **fails open**:

- `readBitDepth` (~L315): infers bit depth from `MediaFormat.KEY_PROFILE` and **returns 8 when the key is missing**. `KEY_PROFILE` is optional in extractor track formats and frequently absent.
- `readIsHdr`: depends on color-transfer / HDR-metadata keys — likewise optional in extractor output.

With both keys absent, a 10-bit HLG H.264 capture that is within the resolution/bitrate budget is classified 8-bit SDR "already optimal" → **passes through un-re-encoded** → the original High-10 bytes ship. (The HLS segmenter is stream-copy, so the profile survives to the wire.) The re-encode leg is safe — it pins `yuv420p` unconditionally when the toggle is off — which is why only in-budget captures escape.

Note the descriptor said `codec=video/hevc` while the wire samples are `video/avc avc1.6E0028` — the probe layer is inconsistent with itself, which supports the mis-probe diagnosis and is worth fixing alongside.

## Environment
- Sender: Pixel 8, dev build, `allow_ten_bit_video` **confirmed OFF**; video recorded via the in-app camera (FSAE / content bubble), not a gallery pick
- Receiver: Android, version-independent (reproduced on builds 1947 and 1949)

## Steps to reproduce
1. On a Pixel 8 (10-bit HLG camera default), record a video with the **in-app camera** from the attachment editor.
[...]

PR #1137, opened 2026-07-23
fix(video): 10-bit capture fails closed + decode-failure message (#959)

toddmitchell:
Refs #959. Covers parts 1 (the actual fix) and 2 (receiver error UI). Part 3 (descriptor hevc-vs-avc mismatch) left as a follow-up.

## The bug
A 10-bit HLG in-app capture on a Pixel 8 shipped as AVC High-10 (`avc1.6E0028`) **with the 8-bit pin ON** — undecodable on most Android receivers, which showed a silent black player. The 8-bit policy was never overridden; it was *bypassed* by a probe that fails open.

## Part 1 — fail closed + authoritative probe (the fix)
`FfmpegCompressPlanner.isAlreadyOptimal` correctly disqualifies pass-through for 10-bit/HDR, but its inputs came from `MediaExtractor`, which returned **8-bit/SDR when `KEY_PROFILE` / color-transfer were absent** (frequently, for camera captures). So a 10-bit source was misclassified "already optimal" and passed through un-re-encoded.

- `VideoTrackInfo.bitDepth`/`isHdr` are now **nullable** — null = "probe couldn't determine" (matches the type's own docstring). Every platform probe reports null instead of fabricating 8-bit SDR.
- `isAlreadyOptimal` **fails closed**: pass-through requires a *positive* 8-bit-SDR probe; null → re-encode (the encode leg already pins `yuv420p`). Defaults changed to null so a caller that forgets to thread the probe can't accidentally pass a 10-bit source through.
- **Authoritative fallback (avoids needless transcodes):** when Android's `MediaExtractor` leaves bit depth/HDR undetermined, resolve it with an `FFprobeKit` read of `pix_fmt`/`color_transfer` — so a genuinely-8-bit in-budget clip still passes through, and only actually-10-bit ones re-encode. Parsing shared with the JVM path via `VideoProbeTags` (also fail-closed).

## Part 2 — decode-failure message (no more dead player)
The Android player's `onPlayerError` only logged, leaving a cleared spinner over black. It now flips to the existing `VpsState.Error` and shows **"Unable to play 10-bit video on this device"** when the failing renderer format is 10-bit (ColorInfo 10-bit luma/chroma, or a High-10/Main-10 codec string), else a generic message. Strings via `stringResource`.

## Tests
- [x] Planner regression cases: null-bit-depth → re-encode, null-HDR → re-encode, defaults-fail-closed, toggle-on behavior
- [x] Unit tests for the shared `pix_fmt` / colour-tag parsing
- [x] Compiles JVM / Android / iOS / wasm; full `homebase-api` suite + Konsist pass
- [ ] **Needs a Pixel 8 (10-bit HLG camera) to verify end-to-end:** capture with the toggle OFF → received payload is 8-bit `yuv420p` (ffprobe) and plays on a non-Pixel device; and confirm the decode-failure message renders

## Not in this PR
Part 3: reconcile the descriptor `codec` field (`hevc`) with the wire codec (`avc`). Follow-up.

### DIFF
--- a/homebase-api/src/androidMain/kotlin/id/homebase/api/video/FFmpegUtils.android.kt
+++ b/homebase-api/src/androidMain/kotlin/id/homebase/api/video/FFmpegUtils.android.kt
@@ -5,6 +5,7 @@ import android.util.Log
 import androidx.core.net.toUri
 import com.arthenica.ffmpegkit.FFmpegKit
 import com.arthenica.ffmpegkit.FFmpegKitConfig
+import com.arthenica.ffmpegkit.FFprobeKit
 import com.arthenica.ffmpegkit.FFmpegSession
 import com.arthenica.ffmpegkit.FFmpegSessionCompleteCallback
 import com.arthenica.ffmpegkit.ReturnCode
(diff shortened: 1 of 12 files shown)


toddmitchell:
**macOS CI fix pushed** (`e888ba4f1`).

The "Build iOS on macos-latest" job was failing at `compileTestKotlinIosSimulatorArm64`:
```
FFmpegProbeFallbackNativeTest.kt:59: Argument type mismatch: actual type is 'Boolean?', but 'Boolean' was expected.
```
Making `VideoTrackInfo.bitDepth`/`isHdr` nullable broke that iOS-only test (it asserted the AVFoundation fallback returns `8`/`false`). My earlier local check only compiled iOS *main* sources, not the test sources — so I missed it. Fixed now.

Rather than just patch the test, I made the native path fail-closed like the others (it had the same fail-open bug in two spots):
- `probeVideoNative`'s ffprobe branch now uses the shared `VideoProbeTags` parsing (null on absent tags) instead of a local `bitDepthFromPixFmt` that defaulted to 8; the native copy is removed.
- `avProbeVideoTrack` reports `bitDepth`/`isHdr = null` (it can't read pix_fmt) — harmless, since its null codec already forces a re-encode.
- The test asserts null, with an updated comment.

Verified locally with `compileTestKotlinIosSimulatorArm64` (the exact step CI runs).

PR #1152, opened 2026-07-26
fix(chat): resolve attachment MIME from the picked URI, not the sandbox copy (#1149)

### LINKED ISSUE #1149: Photo-picker images send as application/octet-stream and render as generic file chips — MIME resolved from the extension-less sandbox copy, not the picked URI
## Symptom

A user picked **two images** from the gallery and sent them. The message arrives (for both sender and recipients) as **two generic file chips** — a blank document icon plus a download arrow — instead of image bubbles. No thumbnail, no inline preview, no full-screen viewer; the only affordance is "download".

Android; the resolution code is in `commonMain`, so any platform whose picker returns an extension-less display name hits it.

## Root cause

The payload's `contentType` ends up `application/octet-stream`, because the MIME type is resolved **from the sandbox copy** — whose filename has no extension — instead of from the originally picked URI.

Chain:

1. `AttachmentHandler.handleAttachPlatformFile` copies the picked file into the cache **first**, then reads the MIME from the copy:
   ```kotlin
   val it = picked.materializeForUpload(fileOperationsProvider)   // AttachmentHandler.kt:133
   val ct = it.mimeType()?.toString() ?: detectContentTypeFromExtensionOrHint(it.name)
   ```
2. The copy is named `chat_attach_<uuid>_<original name>` (`AttachmentUploadResolve.android.kt:20`). Android's system photo picker hands back display names like `photopicker-100002260X` — **no extension**.
3. Send time resolves the type from that name again — `ImageAttachmentResolve.kt:20` for the image path, `MessageActionsHandler.kt:664/701/731` for the others:
   ```kotlin
   resolveContentType(fileName = name, platformMimeType = mimeType()?.toString())
   ```
   The original `content://` URI (where `ContentResolver.getType()` would have returned `image/jpeg`) is gone by then, and the plain cache `File` resolves its MIME from the extension it doesn't have.
4. `detectContentTypeFromExtensionOrHint` (`ContentTypeUtil.kt:5-9`) does `substringAfterLast('.')` — on a dot-less name that returns the whole string, the map lookup misses, and it falls through to `application/octet-stream`.
5. `MessageAttachmentBuilder.build` only generates thumbnails under `contentType.startsWith("image/")` (`MessageAttachmentBuilder.kt:33`) → the payload ships with **no preview thumbnail at all**.
6. Every renderer keys off the same prefix, so the bubble draws the file chip: `MediaItem.kt:245`, `MessageBubble.kt:980`, `MediaMessage.kt:69`, and the list preview `MessageContentLabel.kt:116/128`.

Note `isImage = true` (set by the gallery launcher, `ConversationContent.kt:596`) only steers the *pending* classification into `AttachmentPendingFile.FileImage`; it never overrides the resolved `contentType`, so the payload still goes out as octet-stream.

## Evidence (sender's log)

- `addMessageWithFiles: message=<id> conversation=<id> files=2`
- `sendMessageInternal: ... payloads=2`
- `DriveUploadProvider.uploadFile ... payload key=chat_web0 size=1911696` / `key=chat_web1 size=1751552`
- `CacheAudit`: `chat_attach_<uuid>_photopicker-1000022602 — 1911689 bytes` and `chat_attach_<uuid>_photopicker-1000022601 — 1751539 bytes`

Two extension-less photo-picker source files, sizes matching the two uploaded payloads. The upload itself succeeded; only the type is wrong.

## Why it's intermittent

Most gallery items come back with a display name that *does* carry an extension (`IMG_2026….jpg`), so the extension fallback saves it. The system photo picker's `photopicker-NNNNNNNNNN` names have none — so the same "send two photos" flow works or fails depending on which picker/provider serves the item.

## Fix direction

1. Read `mimeType()` from the **original** `PlatformFile` (the `content://` URI) *before* `materializeForUpload`, and carry it through to `AttachmentInput.contentType`. This is the actual defect.
2. Also preserve the extension when naming the sandbox copy (derive it via `extensionForMimeType(...)`) so downstream extension-based resolution stops being a coin flip.
3. Belt and braces: when resolution yields `application/octet-stream`, sniff the magic bytes — `ImageFormatDetector.detectFormat` (`homebase-api/.../image/ImageFormatDetector.kt`) already exists and is not consulted here.

Test: extend `ContentTypeUtilTest` / add an attachment-resolve test asserting a dot-less picker name plus a `content://`-sourced `image/jpeg` MIME still yields `image/jpeg` at `AttachmentInput` time.

## Not covered by this issue

Already-sent messages keep their octet-stream payloads; they will keep rendering as file chips unless a separate migration/heuristic (sniff on render when the type is octet-stream) is added.

2002Bishwajeet:
## Symptom

Two images picked from the gallery and sent arrive — for sender and recipients — as **two generic file chips**: blank document icon plus a download arrow. No thumbnail, no inline preview, no full-screen viewer. The upload itself succeeds; only the type is wrong.

Intermittent because most gallery items come back with a display name that *does* carry an extension (`IMG_2026….jpg`), so the extension fallback saves them. Android's **system photo picker** vends `photopicker-NNNNNNNNNN` — no extension — and those break.

## Root cause

The payload's `contentType` ends up `application/octet-stream` because the MIME is resolved from the **pick-time sandbox copy**, whose filename has no extension, instead of from the originally picked URI.

1. `AttachmentHandler.handleAttachPlatformFile` copied the picked file into the cache **first**, then read the MIME from the copy.
2. The copy was named `chat_attach_<uuid>_<original name>` — a plain file. Verified against FileKit 0.13.0 bytecode: Android's `PlatformFile.mimeType()` uses `ContentResolver.getType(uri)` for an `AndroidFile.UriWrapper` but **extension-only** (`MimeTypeMap.getMimeTypeFromExtension`) for a `FileWrapper`. So the copy returns `null` and the original `content://` URI — the only handle that knew `image/jpeg` — is already gone.
3. Send time re-resolved from that same copy (`ImageAttachmentResolve.kt`, plus the File / FileVideo / Audio arms of `MessageActionsHandler.addMessageWithFiles`).
4. `detectContentTypeFromExtensionOrHint` does `substringAfterLast('.')`; on a dot-less name that returns the whole string, the map lookup misses, and it falls through to `application/octet-stream`.
5. `MessageAttachmentBuilder.build` only generates thumbnails under `contentType.startsWith("image/")` → the payload shipped with **no preview thumbnail at all**.
6. Every renderer keys off the same prefix (`MediaItem`, `MessageBubble`, `MediaMessage`, `MessageContentLabel`), so the bubble drew the file chip.

`isImage = true` from the gallery launcher only steers the *pending* classification into `AttachmentPendingFile.FileImage`; it never overrode the resolved `contentType`.

## The fix

**Read the type off the picked handle, before materializing, and carry it through.**

- `AttachmentUploadResolve.kt:61` — new `internal fun PlatformFile.pickedContentType()`: `resolveContentType(fileName = name, platformMimeType = mimeType()?.toString())`. Using `resolveContentType` (not the raw MIME) matters: a document-picker file whose content URI type *is* `application/octet-stream` but whose name says `.pdf` must still resolve to `application/pdf`, so this can never regress relative to the old copy-based read.
- `AttachmentHandler.kt:126` — call it on `picked` **before** `materializeForUpload`, and store the result as `sourceContentType` on the resulting `FileImage` / `FileVideo` / `File`.
- `ConversationListUiState.kt` — `sourceContentType: String? = null` on those three `AttachmentPendingFile` variants (same shape as the existing `FileVideo.sourceFileName` precedent). Null keeps today's behaviour for share-received / already-local files.
- `MessageActionsHandler.kt:692,704,713` + `ImageAttachmentResolve.kt:25` — consult `sourceContentType` at every send site, falling back to the old resolution when it is null. `toImageAttachmentInput` gained an optional `sourceContentType` param.
- `EventComposerSheet.kt:182,188` — the Event cover-photo pickers had the identical `materializeForUpload(...).toImageAttachmentInput(...)` shape; fixed the same way.

**Plus: stop the coin flip at the source.** `sandboxCopyName(name, mimeType)` (`AttachmentUploadResolve.kt:69`) appends an extension derived from the picked MIME when the picker's name has none — `extensionForMimeType` already existed in `ContentTypeUtil.kt`. The android / apple / jvm `materializeForUpload` actuals now use it (one line each). Web's actual is a no-op that keeps the original `PlatformFile`, so it was never affected.

### Deliberately not done

- **Magic-byte sniffing** (`ImageFormatDetector.detectFormat`) as a third fallback — belt-and-braces on a fix that already works; the extra path isn't wanted.
- **No back-migration.** Already-sent messages keep their octet-stream payloads and will keep rendering as file chips. No render-time heuristic was added.
- `detectContentTypeFromExtensionOrHint`'s missing `substringAfterLast('.', "")` default is left alone — it makes dot-less names resolve to octet-stream by accident rather than by intent, but it is not the cause and changing it would alter behaviour for files named exactly like an extension.

## Test

`homebase-chat/src/jvmTest/.../ImageAttachmentContentTypeTest.kt` — a dot-less picker name plus a `content://`-sourced `image/jpeg` must still yield `image/jpeg` at `AttachmentInput` time. Written first against the pre-fix tree:

```
ImageAttachmentContentTypeTest[jvm] > dotlessPickerName_keepsContentUriMime[jvm] FAILED
    org.junit.ComparisonFailure at ImageAttachmentContentTypeTest.kt:14
    expected:<image/jpeg> but was:<application/octet-stream>
1 test completed, 1 failed
```

After the fix, 4/4 pass (`tests="4" skipped="0" failures="0" errors="0"`). A sibling case pins the octet-stream fallback so the reason for the carry-through stays documented, and two cover `sandboxCopyName` (appends `.jpg` for a dot-less name; leaves an existing extension and an unmapped MIME alone).

## Verification

| Command | Result |
|---|---|
| `:homebase-chat:jvmTest :homebase-api:jvmTest --rerun-tasks` | BUILD SUCCESSFUL — 1018 + 1094 tests, 0 failures |
| `:homebase-chat:compileKotlinJvm` / `compileAndroidMain` / `compileKotlinWasmJs` / `compileKotlinIosSimulatorArm64` | BUILD SUCCESSFUL (all four targets) |
| `:homebase-core:` same four targets | BUILD SUCCESSFUL |
| `:androidApp:compileDebugKotlin` | BUILD SUCCESSFUL |

`homebase-common` was not touched, so its Konsist architecture tests are unaffected.

Still needs on-device confirmation on Android: pick two photos through the **system photo picker** (the gallery/paperclip route, not the in-app gallery grid) and check they render as image bubbles with thumbnails on both sender and recipient.

Fixes #1149

### DIFF
--- a/homebase-chat/src/androidMain/kotlin/id/homebase/chat/conversationlist/AttachmentUploadResolve.android.kt
+++ b/homebase-chat/src/androidMain/kotlin/id/homebase/chat/conversationlist/AttachmentUploadResolve.android.kt
@@ -5,8 +5,8 @@ import io.github.vinceglb.filekit.FileKit
 import io.github.vinceglb.filekit.PlatformFile
 import io.github.vinceglb.filekit.cacheDir
 import io.github.vinceglb.filekit.copyTo
+import io.github.vinceglb.filekit.mimeType
 import io.github.vinceglb.filekit.name
-import kotlin.uuid.Uuid
 
 // Native PlatformFile carries a real path / content:// URI; the existing path-based send
 // pipeline (and resolveToFilePath for content URIs) already handles it. No copy needed at send time.
@@ -17,7 +17,7 @@ actual suspend fun PlatformFile.toUploadPath(fileOps: FileOperationsProvider): S
 // copying through FileKit's copyTo (which reads the content URI) yields a stable plain file the
 // send path can always read — equivalent to the downstream resolveToFilePath copy, just earlier.
 actual suspend fun PlatformFile.materializeForUpload(fileOps: FileOperationsProvider): PlatformFile {
-    val dest = PlatformFile(FileKit.cacheDir, "chat_attach_${Uuid.random()}_$name")
+    val dest = PlatformFile(FileKit.cacheDir, sandboxCopyName(name, mimeType()?.toString()))
     copyTo(dest)
     return dest
 }
--- a/homebase-chat/src/appleMain/kotlin/id/homebase/chat/conversationlist/AttachmentUploadResolve.apple.kt
+++ b/homebase-chat/src/appleMain/kotlin/id/homebase/chat/conversationlist/AttachmentUploadResolve.apple.kt
@@ -5,8 +5,8 @@ import io.github.vinceglb.filekit.FileKit
 import io.github.vinceglb.filekit.PlatformFile
 import io.github.vinceglb.filekit.cacheDir
 import io.github.vinceglb.filekit.copyTo
+import io.github.vinceglb.filekit.mimeType
 import io.github.vinceglb.filekit.name
-import kotlin.uuid.Uuid
 
 // iOS PlatformFile carries a real file path; the path-based send pipeline reads it directly.
 // No copy needed at send time — the scoped copy happened at pick time (materializeForUpload).
@@ -16,10 +16,11 @@ actual suspend fun PlatformFile.toUploadPath(fileOps: FileOperationsProvider): S
 // (this runs in the picker-callback launch before any further suspension). FileKit's copyTo opens
 // the source through its NSURL — which still holds the scope grant here — and writes a plain
 // sandbox file, so the later scope-less path read in the send path always succeeds.
-// Preserve the original name so the send path's content-type / display-name resolution is unchanged;
-// prefix with a uuid to keep concurrent picks from colliding.
+// Keep the original name (plus a MIME-derived extension when it has none) so the send path's
+// content-type / display-name resolution still works; prefix with a uuid so concurrent picks
+// can't collide. See sandboxCopyName.
 actual suspend fun PlatformFile.materializeForUpload(fileOps: FileOperationsProvider): PlatformFile {
-    val dest = PlatformFile(FileKit.cacheDir, "chat_attach_${Uuid.random()}_$name")
+    val dest = PlatformFile(FileKit.cacheDir, sandboxCopyName(name, mimeType()?.toString()))
     copyTo(dest)
     return dest
 }
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/AttachmentHandler.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/AttachmentHandler.kt
@@ -11,7 +11,6 @@ import id.homebase.core.audio.AudioRecorder
 import id.homebase.core.audio.AudioWaveFormGenerator
 import id.homebase.core.clipboard.platformFileFromPath
 import id.homebase.core.localization.TranslationUtil
-import id.homebase.core.util.detectContentTypeFromExtensionOrHint
 import id.homebase.chat.services.image.StickerImageProcessor
 import id.homebase.chat.services.image.removeBackground
 import id.homebase.chat.services.image.warmUpBackgroundRemoval
@@ -22,8 +21,6 @@ import id.homebase.resources.chat_attach_file_failed
 import id.homebase.resources.chat_message_audio_recording_help
 import id.homebase.resources.chat_remove_background_failed
 import io.github.vinceglb.filekit.PlatformFile
-import io.github.vinceglb.filekit.mimeType
-import io.github.vinceglb.filekit.name
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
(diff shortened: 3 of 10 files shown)


Seifert69:
I'm wondering if the code would be cleaner if PlatformFile contains a filetype() function which can infer the filetype so that we don't need to scatter logic like checking for an empty mime string and then calling a function to detect.

2002Bishwajeet:
Followed up on this in #1197 — you were right, and the scatter was wider than this PR touched.

`resolveContentType(f.name, f.mimeType()?.toString())` was written out at **14** sites, and **7** more read `mimeType()` raw with no extension fallback at all. `PlatformFile` is FileKit's type so it can't take a member function, but the extension you're describing already existed here as `pickedContentType()` — just chat-local and named for one use case. #1197 promotes it to `homebase-common` as `PlatformFile.contentType()` and sweeps every site onto it.

The 7 raw-MIME sites turned out to be the same defect this PR fixed for chat, still live elsewhere: with an extension-less picker name, `MomentComposeScreen` classified a picked **video** as an image (twice), and `ShareReceiverActivity` did the same on its gallery route.

One thing the refactor deliberately doesn't touch: #1149's cause is temporal, not structural — the picked `content://` handle has to be read before `materializeForUpload` discards it. A tidier `contentType()` at send time still reads the sandbox copy. The `sourceContentType` carry-through stays.

Issue #1153, opened 2026-07-27
Conversation-list preview disagrees between live arrival and cold load when the newest item is a status message


2002Bishwajeet:
## Summary

After #1148 (PR #1151), the conversation-list preview for a thread whose newest item is a **status message** (`dataType = 202` — group rename, photo update, member add/remove) differs depending on how the row was populated:

| path | preview shown |
|---|---|
| **live arrival** (`applyIncomingMessageBump`) | the status message's own text — *"You updated the conversation photo"* |
| **cold start / post-`DriveSync.Stopped` enrichment** (`ConversationMapper.applyLastMessage`) | the newest **non-status** message |

Same conversation, same data — the preview text changes after a restart or a background sync.

## Why

The two paths own the preview independently and disagree about whether a status message *is* the last message:

- `applyIncomingMessageBump` calls `withPreviewOf()` on its strictly-newer branch for **every** arrival; only the unread `increment` is gated on `!m.isStatusMessage` (`homebase-chat/.../services/convo/ConversationMessageBump.kt:82`, `:107-131`).
- `selectAllConversationPlusLastMessage` **excludes** `dataType = 202` (`homebase-api/.../database/ChatReadCount.sq:51`), so enrichment can only ever return a real message.

Before #1148 this was masked: the enrichment result lost the `candidate >= latestMessageTimestamp` guard and the row kept `mapToBasic`'s `" "` placeholder, so the cold-load case rendered *"No messages yet"*. #1148 correctly fixed the blank, which makes the underlying disagreement visible instead.

Note this is **pre-existing** and was not introduced by #1151 — that PR strictly improved the cold-load case (blank → real text).

## Options

1. **Include status messages in the last-message query.** Drop the `dataType != 202` exclusion (or add a status-aware variant) so enrichment can return the status message and both paths agree on showing *"You updated the conversation photo"*. Matches WhatsApp/Signal behaviour. Touches the SQL + anything relying on that projection excluding status rows — needs care around the unread-count queries that share the file.
2. **Stop status messages setting the preview on the live path.** Gate `withPreviewOf()` on `!m.isStatusMessage` in `applyIncomingMessageBump` so live matches cold-load and the preview is always the newest real message. Smaller and lower-risk, but group-admin activity then bumps the row's sort order with no visible reason for the move.

Option 1 is the better product behaviour; option 2 is the cheaper correctness fix. Either way the two paths should be made to agree.

## Evidence

[...]

PR #1204, opened 2026-08-03
fix(chat): a message that lands while a conversation is archived unarchives it (#1145)

### LINKED ISSUE #1145: Archived conversation stays archived when it receives a new message
### Problem

Archiving a conversation is permanent until the user manually unarchives it. If an
archived conversation receives a new incoming message, it stays buried in
**Archived** — it does not return to the main conversation list. The user only
finds out there was a new message if they happen to open the Archived screen.

Every other chat app (WhatsApp, Signal, Telegram, iMessage) treats archive as
"hide until there's new activity", not "hide forever".

### Expected

A new **incoming** message in an archived conversation unarchives it: the
conversation reappears in the main list, sorted by the new message's timestamp,
with its unread badge.

### Current behaviour

`ConversationState.Archived` is derived from the `ConversationArchivedTag`
(`a569e5cd-…`) local tag on the conversation file
(`ConversationMapper.kt:186`). The tag is only ever removed by
`ConversationService.unarchiveConversation()`, and every caller of that is a
user-initiated UI action (`ConversationListPane`, `ArchivedConversationsScreen`,
`ConversationContent` menu). Nothing in the message-receive path touches it.

### Sketch

In the path that applies a newly-received inbound chat message to a conversation
(sync/notification handling, the same place that bumps the conversation's last
message + unread count), if the conversation carries `ConversationArchivedTag`,
drop it — i.e. call the existing `unarchiveConversation()`. The tag lives in
`localAppData`, so the unarchive syncs to the user's own devices like the manual
one does.

### Open questions

- **Own-sent messages:** sending into an archived conversation should probably
  also unarchive it (you're clearly re-engaging), but that's a separate trigger
  from "receives a message". Suggest: yes, unarchive on send too.
- **Muted conversations:** if a conversation is both archived and muted, should
  it stay archived? WhatsApp's "Keep chats archived" setting exists exactly for
  this. Simplest v1: always unarchive, no setting. Add the setting only if it's
  actually asked for.
- **System/status messages** (joins, leaves, delivery receipts) should not count
  as "a message" for this purpose.

2002Bishwajeet:
Closes #1145.

## Correcting the premise

The issue says nothing in the message-receive path touches the archive flag. That was not true when it was written. `ConversationStream.processMessageBatchIncrementally` has flipped an archived conversation back to Active on an incoming message since #299 (9d47235ef, 16 Apr 2026), wired in `AppModule` to `ConversationService.unarchiveConversation`. On the WebSocket push path, auto-unarchive already works on `main`.

So this PR is not "add auto-unarchive". It is two narrower things: cover the sync path, and stop a status message from counting as re-engagement.

## The window that is genuinely broken

That live branch runs on `DataEvent.BatchReceived`, which only the WebSocket push path emits. DriveSync is a silent batched DB write by design (`DriveSync.kt`: "DriveSync is a silent batched DB write") — it upserts headers into `DriveMainIndex` and signals `DriveEvent.Stopped`, with no per-row events. `Stopped` reloads the list from the DB, the `ConversationArchivedTag` is still on the file, and the thread renders Archived again.

The broken set is therefore: **rows that reach `DriveMainIndex` via QueryBatch without a corresponding WS push.** In practice that is a first sync after install or re-login, a cold-wake sync that drains before a socket is up, and any push the socket misses or drops while the sync cursor still picks the row up.

It is *not* "everything delivered by FCM cold-wake, reconnect catch-up, or background sync", which is what an earlier revision of this description claimed. A captured July 2026 log shows the opposite for reconnect: `syncAll` returns 0 records and the rows arrive moments later over WSPush — i.e. much of what looks like offline backlog comes down the path that already works. The fix is still worth making; the blast radius is smaller than the issue implies.

## The fix

**Cold half (new).** `archiveConversation` stamps `archivedAt` into the conversation's owner-synced `localAppData.content`, alongside the existing `lastReadTime` / `lastExitedAt` / `draft`. `enrichWithLastMessages` — which both `start()` and the `DriveEvent.Stopped` handler already run — unarchives any archived thread whose last message is newer than that stamp and isn't ours. Because it reads only persisted state, it also works on a cold start, where an in-memory before/after diff has nothing to compare against.

**Live half (gate only).** The existing WS-push branch now also requires `!m.isStatusMessage`. A group rename or a join/leave is not re-engagement — the issue calls this out explicitly. Nothing else about that branch changes.

**Re-fire gate (new).** The cold half runs on *every* sync `Stopped` and every cold start, re-reading state from the DB each pass. If the unarchive write never reaches the local row — the conversation is missing from `DriveMainIndex`, or `updateConversationTags` throws — the row still reads `Archived` next pass, and an ungated cold half would fire again and enqueue another pair of outbox rows, indefinitely. `AutoUnarchiveGate` dedups per session, the same shape as `ConversationStream.recoveryAttemptedIds`. It keys on the `archivedAt` baseline rather than the conversation id alone, so archiving a thread *again* stamps a fresh baseline and re-opens the gate — a second archive/unarchive cycle in one session still works.

## Why not reuse `fileMetadata.updated` as the baseline

It looks like a free archive timestamp and it isn't. `OptimisticWriter.updateLocalTags` stamps `updated = updated + 1ms` — a version counter, not a wall clock. On a thread untouched for weeks it reads weeks old, so every message already in the DB looks post-archive and **archiving would never stick**. Hence the explicit stamp.

`archivedAt` is cleared on unarchive, so a later re-archive whose stamp write fails degrades to "never auto-unarchive" rather than to a stale timestamp that unarchives on the next sync.

## Known limitation: no backfill for existing archives

There is deliberately no migration that stamps `archivedAt` onto conversations that are already archived when this ships.

`shouldAutoUnarchive` requires a non-null `archivedAt` — with no baseline every message in the DB looks newer than the archive, so a backfill-free `null` is the only safe reading. The consequence: **for a thread archived before this upgrade, the new cold-sync half never fires.** It is inert on those threads, not merely conservative, and it stays inert until the user archives that thread again, which writes the stamp.

Those threads are not left worse off than today: the live WS branch does not consult `archivedAt` at all, so they keep exactly the `main` behaviour — a push-delivered message still unarchives them. They just don't gain the sync-path coverage until they're re-archived.

This is a deliberate scope decision, not an oversight. Recording it here so it isn't later read as a bug.

## Status messages, both paths

Cold path: the last message comes from `selectAllConversationPlusLastMessage`, which excludes `dataType = 202`. Live path: the new `!isStatusMessage` gate. This matches the dual-gate invariant established in #1153.

## Deliberately not done

Sending into an archived conversation still does not unarchive it. The issue lists that as a separate trigger under Open questions.

No "keep chats archived" setting — the issue's own v1 recommendation.

## Verified

- Rebased onto `main` (`1fc3ebee4`), no conflicts. The only new main commit touching a file here is #1397 (`412bad7e9`, optimistic rollback), which rewrote `rollbackWrite` and factored out `optimisticStamp()`; the new `stampConversationArchivedAt` sits in a different region and replayed cleanly.
- `./gradlew homebase-chat:jvmTest homebase-common:jvmTest homebase-api:jvmTest` — green.
- `AutoUnarchiveTest` (6 cases) pins the decision: unarchives after the stamp; **stays archived** for a message from before it, for a message at exactly the stamp instant, for our own message, and with no baseline; never touches a non-Archived state.
- `AutoUnarchiveGateTest` (3 cases) pins the re-fire gate: five cold passes with the write never landing produce one enqueue; a re-archive re-opens the gate; conversations are gated independently.
- Both gate assertions were mutation-checked. Making `markFired` always return true fails `coldPassRepeatedWhileTheWriteNeverLands_enqueuesOnce` and `reArchivingStampsANewBaseline_soTheGateReopens`; keying it on the conversation id alone fails `reArchivingStampsANewBaseline_soTheGateReopens` and nothing else.
- `ConversationServiceLifecycleTest` archive/unarchive assertions updated 1 → 2 outbox rows (the content stamp rides alongside the tag update).

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
@@ -57,6 +57,8 @@ data class ConversationUiModel(
     val isGroup: Boolean = false,
     val isLegacyGroup: Boolean = false,
     val exitedAt: Instant? = null,
+    /** Null unless archived, and on threads archived before the stamp existed (#1145). */
+    val archivedAt: Instant? = null,
     /** Server-stamped last-modified time of the conversation file (fileMetadata.updated). */
     val fileUpdated: Instant = Instant.fromEpochMilliseconds(0),
     /**
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/data/ConversationUiModel.kt
@@ -57,7 +57,7 @@ data class ConversationUiModel(
     val isGroup: Boolean = false,
     val isLegacyGroup: Boolean = false,
     val exitedAt: Instant? = null,
-    /** Null unless archived, and on threads archived before the stamp existed (#1145). */
+    /** Null unless archived, and on threads archived before the stamp existed. */
     val archivedAt: Instant? = null,
     /** Server-stamped last-modified time of the conversation file (fileMetadata.updated). */
     val fileUpdated: Instant = Instant.fromEpochMilliseconds(0),
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
@@ -2,6 +2,7 @@ package id.homebase.chat.services.convo
 
 import id.homebase.chat.data.ConversationState
 import kotlin.time.Instant
+import kotlin.uuid.Uuid
 
 /**
  * Cold half of auto-unarchive: decides from persisted state alone, so it still
@@ -25,3 +26,17 @@ internal fun shouldAutoUnarchive(
         archivedAt != null &&
         !lastMessageIsFromActiveUser &&
         lastMessageUserDate > archivedAt
+
+/**
+ * Dedup for the cold half. It runs on every sync `Stopped` and on every cold
+ * start, so a tag removal that never reaches the local row would otherwise
+ * re-fire — and re-enqueue — on every pass. Keyed on the baseline rather than
+ * the id alone, so a later re-archive stamps a fresh [Instant] and re-opens the
+ * gate. Session-scoped, like `ConversationStream.recoveryAttemptedIds`.
+ */
+internal class AutoUnarchiveGate {
+    private val fired = mutableMapOf<Uuid, Instant>()
+
+    fun markFired(conversationId: Uuid, archivedAt: Instant): Boolean =
+        fired.put(conversationId, archivedAt) != archivedAt
+}
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
@@ -4,9 +4,9 @@ import id.homebase.chat.data.ConversationState
 import kotlin.time.Instant
 
 /**
- * Cold half of auto-unarchive (#1145): decides from persisted state alone, so it
- * still fires after a restart or a silent DriveSync catch-up — neither of which
- * emits the `BatchReceived` the live half in
+ * Cold half of auto-unarchive: decides from persisted state alone, so it still
+ * fires after a restart or a silent DriveSync catch-up — neither of which emits
+ * the `BatchReceived` the live half in
  * `ConversationStream.processMessageBatchIncrementally` listens for.
  *
  * Callers must pass the last message from `selectAllConversationPlusLastMessage`,
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/convo/AutoUnarchive.kt
@@ -1,0 +1,27 @@
+package id.homebase.chat.services.convo
+
+import id.homebase.chat.data.ConversationState
+import kotlin.time.Instant
+
+/**
+ * Cold half of auto-unarchive (#1145): decides from persisted state alone, so it
+ * still fires after a restart or a silent DriveSync catch-up — neither of which
+ * emits the `BatchReceived` the live half in
+ * `ConversationStream.processMessageBatchIncrementally` listens for.
+ *
+ * Callers must pass the last message from `selectAllConversationPlusLastMessage`,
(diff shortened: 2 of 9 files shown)


2002Bishwajeet:
## Review caveats

Three things a future reader needs that the diff does not show on its own. Line refs are against `68cc0d446`.

### 1. The gate keys on the `archivedAt` baseline, not the conversation id — deliberately

`AutoUnarchive.kt:40-41` — `fired.put(conversationId, archivedAt) != archivedAt`.

The in-file precedent, `recoveryAttemptedIds` (`ConversationStream.kt:159`), is a plain `mutableSetOf<Uuid>()`. Copying it literally here buys a real false negative: archive → auto-unarchive → re-archive → sync-delivered message, all inside one process lifetime. The second unarchive stays suppressed for the rest of the session.

It also fails loudly rather than quietly. The gate at `ConversationStream.kt:1183` guards only the `onUnarchiveConversation` call; the in-memory flip at `ConversationStream.kt:1174` runs unconditionally. An id-keyed gate would render the thread Active for that pass, never persist the tag removal, and revert on the next `Stopped` reload — a flicker, not a no-op.

`AutoUnarchiveTest.kt:83` `reArchivingStampsANewBaseline_soTheGateReopens` is the test that fails under an id-only key, and the only one — `coldPassRepeatedWhileTheWriteNeverLands_enqueuesOnce` and `conversationsAreGatedIndependently` both still pass. Anyone "simplifying" the key back to an id is caught by exactly that assertion.

### 2. Known test gap: the call-site wiring is unverified

The mutation checks cover the `AutoUnarchiveGate` class, not its wiring into `enrichWithLastMessages` (`ConversationStream.kt:1158-1193`).

Nothing instantiates `ConversationStream`. `grep 'ConversationStream('` across the repo hits the class declaration at `ConversationStream.kt:63` and nothing else; construction is `singleOf(::ConversationStream)` at `AppModule.kt:750`, which needs the full DB + DI graph. The repo's convention for this is spelled out in `ConversationMessageBumpTest.kt:427-433` — extract the decision to a pure seam, test the seam, keep the loop coupled by hand.

Stated plainly: **inverting the `!` at `ConversationStream.kt:1164` would be caught by no test.** That line was verified by reading only.

### 3. Why the "known limitation" paragraph in the body is not a bug

The cold half is genuinely inert — not merely conservative — on threads archived before this ships. `shouldAutoUnarchive` requires a non-null `archivedAt` (`AutoUnarchive.kt:26`), and `ConversationMapper.kt:239-241` populates that field solely from `localAppData?.archivedAt`. No stamp, no baseline, no fire.

The live WS branch, though, never consults `archivedAt`. Its gate is `!m.isStatusMessage && !m.isAuthoredBy(...)` (`ConversationStream.kt:509`); the field appears in that branch only in the `copy(archivedAt = null)` it writes. Pre-existing archived threads therefore keep exactly today's `main` behaviour — a push-delivered message still unarchives them. They lose nothing; they gain sync-path coverage only once re-archived, which writes the stamp.

Issue #1102, opened 2026-07-16
Batch drive-sync pulls into one query-batch-collection call instead of ~7 per-drive query-batch round trips (wire up #1100)


Seifert69:
## Enhancement

Drive sync currently pulls each mounted drive with its **own** round trip. A sync cycle (cold start / foreground) fires ~**7 separate `POST /drives/<id>/files/query-batch`** calls (chat, contacts, moments, location, stickers, vault, feed). Lump the own-host drives into a **single** call using the new **`query-batch-collection`** API.

The client method already exists — **`DriveQueryProvider.queryBatchCollection`**, added in **#1100** (merged), which runs multiple named sections (each with its own drive/filters/paging) over `POST /drives/query-batch-collection` in one round trip. That PR explicitly wired **no call sites** — this is the wire-up.

## Current shape
- `DriveSyncManager` holds `driveSyncs: Map<Uuid, DriveSync>` — one `DriveSync` per drive; `syncAll()` iterates them.
- Each `DriveSync` builds `FileQueryParams` + `QueryBatchResultOptionsRequest` (with its own cursor) and calls `driveQueryProvider.queryBatch(driveId, request)` (`DriveSync.kt:337`).
- ⇒ N drives = N round trips per cycle.

## Proposed
In the sync cycle (`DriveSyncManager.syncAll()` / the initial per-drive pull), gather each **own-host** drive's query into a `CollectionQueryParamSection` (keyed by driveId), issue **one** `queryBatchCollection`, then route each section's `QueryBatchResponse` back to its `DriveSync` for the existing decode → DB-write → cursor-advance path.

## Benefits
- **1 round trip instead of ~7** → faster cold-start/foreground sync, fewer connections/handshakes, less per-request overhead.
- Big win on **high-latency / flaky networks**: 7 sequential/parallel requests each risk a DNS/connect failure; one call collapses that exposure (relevant to the airplane/hostile-DNS cases).

## Caveats / must-preserve
1. **Own-host drives only.** Remote/owner-hosted (peer) drives route over peer with a separate retry cadence (`isRemote` in `DriveSyncManager`) — keep them out of the collection; batch only the logged-in user's own-host drives.
[...]

PR #1241, opened 2026-08-05
Adapt the collection client to per-section status and a request-level budget

Seifert69:
Client-side adaptation to the server changes in [odin-core#1630](https://github.com/homebase-id/odin-core/pull/1630) (odin-core issue #1629). API layer only — **no call sites are wired**; that's #1102.

## What changed on the server

`POST /api/v2/drives/query-batch-collection` (V2 only; V1 is untouched):

- A section-level fault — unknown drive, archived drive, no read grant — no longer fails the whole collection. It comes back as a section with an explicit status.
- `maxRecords` is a budget for the **whole call**, filled greedily in request order, not a per-section limit. Sections the budget doesn't reach return `budgetExhausted` with the submitted cursor echoed verbatim.
- A section can specify its own `fileSystemType`, so one collection can mix Standard and Comment.

## Request

- **`maxRecords` moves onto `QueryBatchCollectionRequest`** as a required constructor parameter. Required rather than defaulted because the server 400s on anything below 1 — a caller who forgets it should fail to compile, not at runtime. Also checked client-side, which saves a wasted round trip mid-sync.
- **`CollectionSectionResultOptions` replaces `QueryBatchResultOptionsRequest`** for sections: identical minus `maxRecords`, so the type can't express a value the server would silently ignore. `QueryBatchResultOptionsRequest` is unchanged and still used by single `query-batch`.
- **Sections take an optional `fileSystemType`**, null meaning "use the request-level one".

## Response

- **`QueryBatchCollectionSection` replaces `QueryBatchResponse`** as the collection element type, adding `status`, `errorMessage`, `errorCode`. `QueryBatchResponse` is untouched — it stays the shape of the single-query endpoints, which don't have these fields.
- **`status` is nullable.** An unknown value from a newer server, or none at all from an older one, decodes to `null` under `coerceInputValues` rather than throwing. There's a test pinning this: one unrecognised enum value quietly failing an entire sync would be miserable to diagnose in the field.
- **`isFailure`** separates "empty because something broke" from "empty because nothing changed" — `budgetExhausted` is explicitly *not* a failure, it just wasn't reached. **`describeFailure()`** produces the greppable per-section line #1102 asks for.
- **`invalidDrive` stays populated** and `isFailure` falls back to it, so this still behaves correctly against a server predating the change.

## For #1102

Its stated open risk — *"server per-section failure semantics are UNVERIFIED (must test against a real tenant)"* — is now closed. odin-core#1630 covers non-existent, archived and ungranted drives against a real tenant across SQLite and Postgres, not a MockEngine. The fail-soft premise behind caveat #2 holds.

Two things that will matter when wiring the sync engine:

- The budget is now **global to the call**, so section N's results depend on what sections 1..N-1 consumed. Section order is caller-controlled priority and the server never reorders — a drive with a large backlog will consume consecutive budgets, and rotating the order is how you'd get fairness.
- `budgetExhausted` sections must be re-sent **with the cursor exactly as returned**. It's the cursor you submitted, echoed back.

## Tests

8 tests in `DriveQueryProviderQueryBatchCollectionTest`, all MockEngine:

- per-section independent decode, with statuses
- budget sent once per request; section options provably carry no `maxRecords` on the wire
- per-section `fileSystemType` serialized only when set
- `budgetExhausted` echoes its cursor, carries no rows, is not a failure
- unknown status → `null`, not a throw
- missing status → falls back to `invalidDrive` (old server)
- `describeFailure()` output
- `maxRecords < 1` rejected before any request is sent

`./gradlew :homebase-api:jvmTest` — 1126 tests, 0 failures. Android and WasmJs targets compile.

**Verified against a live tenant.** A temporary developer-menu probe exercised this API against a deployed identity — fault isolation with the bad section leading, the global greedy budget, verbatim cursor echo on `budgetExhausted`, and a multi-round drain with no duplicate rows. **5/5 passed**; full output in the results comment below. The probe is deliberately not part of this PR — it is parked on branch `qb-collection-probe-archive` to be resurrected as a regression check once #1102 wires the sync engine.

https://claude.ai/code/session_01SLmHNQ1cSDFP4eRHvfUpGC


Seifert69:
## Probe run against a live tenant — 5/5

Desktop client, logged-in identity, deployed odin-core. **The tenant has odin-core#1630** (statuses populated, `BudgetExhausted` present).

### Fault isolation — confirmed, and it survives a leading fault

```
section=ghost  status=DriveNotFound rows=0 hasMoreRows=false invalidDrive=true
               code=invalidDrive message=Invalid drive id 428580df-…
section=chat   status=Ok            rows=100
→ PASS — collection survived; 100 rows still served behind it

section=wallet status=NoReadGrant   rows=0 hasMoreRows=false invalidDrive=true
               message=Unauthorized to read to drive [a6f991e2-…]
section=chat   status=Ok            rows=100
→ PASS — 100 rows served behind it
```

Both faults were injected at the **head** of the section list, so this also pins **a failed section consumes no budget and does not stop the fill** — the sections behind it were served in full. `walletDrive` exists on this tenant and the app has no read grant on it, so `noReadGrant` is a real result, not a `driveNotFound` in disguise.

Worth noting for the sync wire-up: failed sections come back `hasMoreRows: false`, so a drain loop keyed on `needsAnotherRound` will not spin forever on a broken drive.

### Budget — global and greedy, exactly as specified

Budget 1000 → chat returned exactly 1000 and everyone behind it was skipped. Budget 1 → exactly 1 row total. Skipped sections echoed their submitted cursor verbatim.

### Drain loop — the shrinking collection emerges from the budget

```
round 1: head=chat     → chat=50,  contacts=0,  profile=0   still paging=3
round 2: head=contacts → contacts=50, profile=0, chat=0     still paging=3
round 3: head=profile  → profile=41, chat=9,  contacts=0    still paging=2
round 4: head=contacts → contacts=29, chat=21               still paging=1
round 5: head=chat     → chat=50                            still paging=1

rows by drive: {chat=130, contacts=79, profile=41}
250 rows, no duplicate fileId
```

Three things fall out of this:

1. **The collection shrinks on its own** — 3 sections → 1 in four rounds, as drives drain and stop reporting `needsAnotherRound`. #1102's caveat #3 asked for orchestration to carry unfinished sections forward; the budget contract produces it for free.
2. **Rotation is mandatory, not a refinement.** The first run of this probe used a fixed order: chat took 50/50 for five straight rounds and contacts and profile received **zero**. Rotating the head each round is the only reason every drive is served above.
3. **A round can serve several drives.** Round 3 drained profile with 41 and spilled the remaining 9 into chat — the fill is "serve in order until the budget is out", not one drive per round.

### Cost of pessimistic `hasMoreRows`, measured

A `budgetExhausted` section returns `hasMoreRows: true` because the server never looked. Round 3 shows the price: profile carried "still paging" through two rounds before revealing it had 41 rows left and was done. One wasted tail round per drive. We declined the `LIMIT 1` lookahead in odin-core#1629 as an optional optimisation — this quantifies it, and it is not worth revisiting.

### Open question this leaves for #1102

Strict rotation gives every drive an equal share of throughput. For a chat app that may be wrong — a chat backlog and a sticker backlog are not equally urgent. Fixed priority starves (proven above), strict rotation is fair, weighted rotation (chat leads 2 rounds in 3) is the middle. Section order is caller-controlled and the server never reorders, so this is purely a client policy decision, and it belongs with the sync-engine design rather than here.

PR #1242, opened 2026-08-05
perf(chat): seek the pinned-bar tag index instead of probing it per message

2002Bishwajeet:
## Problem

The pinned-messages bar refreshes on every conversation open, and its query joins in the wrong direction.

`DriveLocalTagIndex`'s only index is `UNIQUE(identityId,driveId,fileId,tagId)` — it leads with `fileId`, so a `tagId` lookup can't seek it. With no usable index and no `ANALYZE` statistics, SQLite drives `selectFilesByLocalTagInGroup` from `DriveMainIndex` instead: seek the conversation by `groupId`, then probe the tag table once per message. Open cost grows with the conversation's length.

```
SEARCH m USING INDEX idx_chatmessage_convid_userDate (groupId=? AND fileType=?)
SEARCH t USING COVERING INDEX sqlite_autoindex_DriveLocalTagIndex_1 (...)
```

## Fix

Add `idx_localtag_tag(identityId,driveId,tagId,fileId)` and pin the query to it with `INDEXED BY` — same reasoning as `QueryBatch`'s existing `idx_chatmessage_convid_userDate` hint, which was added for this same class of planner choice. `fileId` is in the index so the seek is covering.

```
SEARCH t USING COVERING INDEX idx_localtag_tag (identityId=? AND driveId=? AND tagId=?)
SEARCH m USING INDEX sqlite_autoindex_DriveMainIndex_1 (identityId=? AND driveId=? AND fileId=?)
```

Measured on 20k messages in one conversation, 2 of them pinned: **14.2ms → 0.028ms**.

Also moves `autoPinNewTypedMessages` + `refreshPinnedFor` off the open path into a launched `refreshPinsAfterOpen()`. Both were awaited before the message collector started, so they held the loading spinner up. The bar reaches the UI on its own `StateFlow` (`observePinnedMessages`), so it doesn't need to land before the messages do.

## No schema wipe

The index is over existing columns only, so `Schema.create()` adds it to already-installed databases. **No `DATABASE_VERSION` bump** — that would drop every table and force a full resync.

## Scope — read this before assuming it fixes the spinner

This removes a real O(conversation size) cost, but it is **not** what makes the open spinner visible today.

Measured on a Redmi Note 5 Pro, same conversation, before and after:

| | before | after |
|---|---|---|
| cold open | 1.27–1.36s | 1.279s |
| warm reopen (`reuse window`, zero DB) | — | 572ms |

End-to-end open time is unchanged. The warm reopen is the control: it logs `hasCached=true → reuse window`, does no database work at all, and still takes 572ms. Both traces show a silent gap — 718ms cold, 358ms warm — between `Pre-initializing empty scroll position` and `Restore detail pane`, with nothing logged and no query running. The spinner is dominated by detail-pane composition, which is a separate investigation.

Worth landing on its own: the cost removed here scales with history, so it bites a heavier account than the one measured.

## Tests

`MessageFetchPlanTest.pinnedBarShape_drivesJoinFromTagIndex` asserts via `EXPLAIN QUERY PLAN` that the tag table is the outer loop and is seeked via `idx_localtag_tag`. Mutation-checked — removing the `INDEXED BY` hint fails it.

`:homebase-api:jvmTest` and `:homebase-chat:jvmTest` green. Verified on device: pinned bar still populates on open (`1 of 5`).


2002Bishwajeet:
Closing. The complexity fix is real (O(conversation size) -> O(pinned count); 14.2ms -> 0.028ms isolated at 20k messages), but it was never shown to move end-to-end conversation-open time, and subsequent profiling on a non-debuggable build put the pinned-bar query well outside the top costs. Not worth carrying an index + a forced INDEXED BY hint on unproven benefit.

Branch `perf-chat-open-pinned-bar-query` is kept if this is worth revisiting against a genuinely large conversation.

PR #1259, opened 2026-08-11
fix(contactbook): hero the contact avatar into its own viewer branch (#1212)

### LINKED ISSUE #1212: Contact-detail avatar viewer: top bar sits a status-bar too low and there is no hero transition
Follow-up to #1190 / PR #1207, which made the contact photo open full-screen on contact detail. Two visible defects were found on device and deliberately deferred to keep that PR reviewable. They are filed together because **one change fixes both**.

## 1. The viewer's top bar sits one status-bar-height too low

On `ContactDetailScreen` the back-arrow + name row is visibly lower than a normal screen's app bar.

`AvatarFullScreenViewer` is rendered **inside the Scaffold's content slot** (`ContactDetailScreen.kt:413`):

```kotlin
fullScreenAvatar?.let { source ->
    AvatarFullScreenViewer(
        source = source,
        title = uiState.entry?.displayName.orEmpty(),
        onDismiss = { fullScreenAvatar = null },
    )
}
```

The Scaffold suppresses its own `topBar` while the viewer is open (`:220`), but it still applies window insets to its content lambda — so the viewer's `Box(Modifier.fillMaxSize())` already starts below the status bar. Then the viewer's own `TopAppBar` (`AvatarFullScreenViewer.kt:68`) applies `TopAppBarDefaults.windowInsets` on top. The status-bar inset is counted twice.

`GroupSettingsScreen` is unaffected: there the viewer lives inside the screen-level `AnimatedContent` and *replaces* the screen, so it never passes through the Scaffold and is inset exactly once.

## 2. No hero transition on contact detail

Group settings morphs the circular avatar into the full-screen image. Contact detail hard-cuts. Three separate reasons, all of which need addressing:

**a. The call site passes no scopes.** `sharedTransitionScope` and `animatedVisibilityScope` both default to null at `ContactDetailScreen.kt:414`. Compare `GroupSettingsScreen`, which wraps in `SharedTransitionLayout` + `AnimatedContent(targetState = fullScreenAvatar)` and passes `this@SharedTransitionLayout` / `this@AnimatedContent` to **both** the header avatar and the viewer.

**b. The shared key is null for a public avatar.** `AvatarFullScreenViewer.kt:50`:

```kotlin
val sharedContentStateKey = (source as? SubSamplingImageSource.Remote)?.imageData
    ?.let { "image-${it.fileId}-${it.payloadKey}" }
```

Only `SubSamplingImageSource.Remote` yields a key. A contact whose photo is a *published public avatar* resolves through the `Url` variant, so the cast fails and there is no key even if the scopes were passed. The hero could only ever fire for a drive-uploaded photo. `Url` needs a stable key derived from the URL.

**c. The small-avatar end registers nothing.** `PublicAvatar` does not declare a shared element at all, so even with (a) and (b) fixed there is nothing to morph *from*. Both ends must register the same key.

## Suggested direction

Restructure `ContactDetailScreen` to the group-settings shape — `SharedTransitionLayout` + `AnimatedContent(targetState = fullScreenAvatar)`, viewer as the non-null branch replacing the screen, Scaffold in the null branch. That applies the inset exactly once *and* puts both ends in the same scopes, fixing (1) and (2a) together.

Patching `windowInsets = WindowInsets(0)` onto the viewer's `TopAppBar` instead would hide this instance of the symptom while leaving the component fragile at any future call site that nests it differently, and does nothing for the hero.

## Landmine for whoever picks this up

Naively wrapping the Scaffold in `AnimatedContent` **resets the tab selection and per-tab scroll position**. `AnimatedContent` disposes the outgoing branch, and neither survives it:

- `var selectedTab by rememberSaveable(entry.uniqueId) { … }` (`ContactDetailScreen.kt:289`)
- `val scroll = remember(current) { ScrollState(0) }` (`:324`)

Both need hoisting above the `AnimatedContent`.

## Must not regress

- The shared-media viewer (`uiState.fullScreenMedia`) keeps working, and the two full-screen surfaces still never stack (`:220`).
- A contact with **no** image stays inert — no empty black viewer. PR #1207 fixed a leak here where `PublicAvatar`'s error branch passed `options` straight to `FallbackAvatar`, which applies `options.onClick` itself; don't reintroduce it.
- Group settings behaviour unchanged — it already works, this is a refactor of a working path.
- Back gesture / Esc still dismisses on Android, iOS and Desktop.

## Test

- Hero fires for **both** avatar branches: drive-uploaded photo and published public avatar.
- Top bar aligns with a normal screen's app bar.
- Tab selection and scroll position survive open → dismiss.
- Group settings regression check.
- Photoless contact: not tappable, no viewer.

## Related

- #1190 — the original request (contact photo full-screen), shipped in #1207
- #1188 — settings-screen owner avatar goes to avatar edit with no visual affordance; separate behaviour, not covered here

2002Bishwajeet:
Fixes #1212.

Both defects on the contact-detail avatar viewer have the same root cause — the viewer was nested inside the screen instead of replacing it — so one restructure fixes both.

## 1. Top bar sat one status-bar-height too low

`AvatarFullScreenViewer` was rendered inside `ContactDetailScreen`'s Scaffold content slot, whose `Box` applies `padding(padding)`. The viewer then applies `TopAppBarDefaults.windowInsets` through its own `TopAppBar`, so the status-bar inset was counted twice.

Not patched with `windowInsets = WindowInsets(0)` on the viewer — that hides this instance and leaves the component fragile at the next nesting call site. The viewer now replaces the screen, so it is inset exactly once.

## 2. No hero transition

Three separate gaps, all addressed:

- **Scopes.** `ContactDetailScreen` now wraps in `SharedTransitionLayout` + `AnimatedContent(targetState = fullScreenAvatar)` and passes `this@SharedTransitionLayout` / `this@AnimatedContent` to both the header avatar and the viewer — mirroring the working `GroupSettingsScreen`.
- **Key for the `Url` variant.** `AvatarFullScreenViewer` derived a key only from `SubSamplingImageSource.Remote`, so a published *public* avatar (which resolves through `Url`) got a null key. Key derivation moved to `SubSamplingImageSource.sharedElementKey()`, with `Url` keyed off the URL via the new `imageUrlSharedElementKey`. The `Remote` form is unchanged and still byte-identical to the key `HomebaseImage` registers.
- **The small end.** `PublicAvatar` declared no shared element. It now takes the two scopes (default null, so every existing call site is untouched) and registers `sharedBounds` on its container with the same URL-derived key, using the same `boundsTransform` / `RemeasureToBounds` as `HomebaseImage`. The key uses the *plain* published URL, never the `?v=` cache-busted one, because the viewer is opened with the plain URL and both ends must agree.

`ContactBookAvatar` threads the scopes into whichever branch rendered (drive photo → `HomebaseImage`, public avatar → `ContactAvatar` → `PublicAvatar`), so the hero fires for both avatar kinds. `DetailHeader` and `PendingRequestProfile` forward them.

## Hoisted state (the landmine in the issue)

`AnimatedContent` disposes the outgoing branch, and neither `rememberSaveable` nor `remember` survives that. Moved above the `AnimatedContent`:

- `selectedTab` (`rememberSaveable`)
- `detailsExpanded` (`rememberSaveable`) — same problem, not named in the issue
- the per-tab `ScrollState` (`remember(currentTab)`)

The tabs list is now a file-level `contactDetailTabs` so `currentTab` can be computed above the branch. The Scaffold body moved into a private `ContactDetailContent` that takes this state as parameters.

## Verified

- `:homebase-common:compileKotlinJvm :homebase-chat:compileKotlinJvm :homebase-core:compileKotlinJvm` — green
- `:homebase-common:compileAndroidMain :homebase-chat:compileAndroidMain :homebase-core:compileAndroidMain` — green
- `:homebase-common:jvmTest` — green (Konsist hardcoded-string / architecture guards)
- No behaviour change to `GroupSettingsScreen` — it is the reference shape and was not touched.
- Every other `ContactAvatar` / `PublicAvatar` call site passes null scopes (explicitly or by default), so the new shared element is inert everywhere except contact detail.
- The photoless path is unchanged: `onClickNeedsImage = true` still gates the tap on a loaded image, and `PublicAvatar`'s error branch still strips `options.onClick` before handing off to `FallbackAvatar` (the #1207 leak is not reintroduced).
- The shared-media viewer (`uiState.fullScreenMedia`) and the action-in-progress scrim stay exactly where they were, inside the Scaffold content. The app bar is still suppressed while the media viewer is open, so the two surfaces never stack; the avatar case no longer needs that guard because the viewer replaces the screen.

## Device verification

Not yet run on a device — needs a contact with a drive-uploaded photo and a contact with a published public avatar to confirm both hero branches, the top-bar alignment, and tab/scroll survival across open → dismiss.


2002Bishwajeet:
## Device verification — Redmi Note 5 Pro (Android 11)

Verified against a contact that resolves through the **published public avatar** branch — confirmed from the app log (`GET frodo.baggins.demo.rocks/pub/image`), i.e. the `SubSamplingImageSource.Url` path that defect 2b says could never hero before.

**1. Top bar inset.** Measured the back-arrow glyph's row span by pixel-scanning the left edge of a 30 fps screen recording:

| screen | back-arrow rows | centre |
|---|---|---|
| contact detail app bar | 136–172 | **154 px** |
| avatar viewer top bar | 136–172 | **154 px** |

Pixel-identical. With the double inset present the viewer's arrow would sit ~66 px (one status bar) lower.

**2. Hero transition.** Frame-stepped the recording: the circular avatar is caught mid-flight as a growing rounded rectangle before settling into the full-screen image — a morph, not a hard cut. Reverse morph on dismiss.

**3. Tab + scroll survival (the landmine).** Switched to Activity, opened the viewer, dismissed — landed back on Activity with the same scroll offset. The hoisting above `AnimatedContent` holds at runtime.

**4. Photoless contact stays inert.** Verified by construction: `ContactBookAvatar.kt:93` passes `options` to `FallbackAvatar` with no `onClick` ever set on that branch, and `:85`'s `onClickNeedsImage = true` keeps the container unclickable so only a loaded image is tappable. Two connected identities that *look* photoless (initials avatars) do open the viewer — correct, since those are real server-generated published avatar images, not the fallback.

Not verified on device: the drive-uploaded (`Remote`) branch — its key derivation predates this PR and was already working; it now also receives the scopes.

PR #1260, opened 2026-08-11
Turn a shared vCard into a contact card instead of an opaque .vcf (#1255)

### LINKED ISSUE #1255: Support sharing a contact into Homebase (parse vCard → contact card; optionally add to the chat-KMP address book)
## Summary (enhancement)

Support **sharing a contact into Homebase** (Android share sheet → Homebase). Today there's no contact/vCard awareness at all, so a shared contact falls through as a generic file and never becomes a usable contact. We should support it properly: turn a shared vCard into a **proper contact card** in the target conversation, and **optionally offer to also add the contact to the chat-KMP address book**.

## Current state (grounded — nothing contact-aware exists yet)

A contact is shared as a vCard — a `.vcf` file (`text/x-vcard` / `text/vcard` via `EXTRA_STREAM`) or a text summary in `EXTRA_TEXT`. Homebase appears in the share sheet (the `ShareReceiverActivity` manifest filters include `text/*` and `application/*`), but from there it's treated as a plain file at every stage:

- `SharedContentExtractor.copyUriToTemp` copies the vCard to a temp file with no vCard awareness.
- `ShareReceiverActivity.convertToAttachmentFiles` → the vCard hits the **`else` branch → `AttachmentPendingFile.File`** (only `image/`/`video/` are special-cased).
- `MessageAttachmentBuilder.build` → **`else` branch → generic `PayloadFile`** (only `image/`, `audio/`, `application/pdf` special-cased).

So the best case today is an opaque `.vcf` document card — never a contact — and the app's in-app **"attach contact" path (`AttachmentOptions.onContactClick`, `:411`) is never reached by an inbound share.

## What to build

1. **Detect a contact share** in the share-receive path: mime `text/vcard` / `text/x-vcard`, or a `.vcf` `EXTRA_STREAM`, or a vCard body in `EXTRA_TEXT`.
2. **Parse the vCard** (name, phones, emails, org, etc.).
3. **Route it into the existing contact-attachment path** (`onContactClick`) so it's sent as a **proper contact card/message**, not an opaque file. Normalize imported fields through `ContactFieldValidation` (E.164 phones, validated emails) — the same rules the in-app contact editor uses.
4. **Optional: offer to add it to the chat-KMP address book.** After (or alongside) sharing into the conversation, prompt "Also add to your contacts?" — on confirm, create a contact in the app's own address book (the `contactbook` feature: `ContactEditSheet` / `PhoneNumberField` / `ContactFieldValidation`, with a stable UUID). Pre-fill the editor from the parsed vCard so the user can review/correct before saving (respecting the E.164/email validation gates and the surrogate-safe name handling).

## Notes / constraints

- Handle both delivery shapes: vCard **file** via `EXTRA_STREAM` and vCard **text** via `EXTRA_TEXT`.
- Legacy/malformed vCard fields may not be E.164 — seed them into `PhoneNumberField` and flag with `isError` rather than dropping, and gate Save on validity (matches the existing contact-edit rules).
- Multi-contact vCards (a `.vcf` with several `VCARD` blocks) — decide whether to support 1 or N; at minimum handle a single contact cleanly and don't crash on multiples.
- Keep the raw-`.vcf`-as-document behavior only as a fallback if parsing fails (so a share is never silently dropped).

## Acceptance criteria

- [ ] Sharing a single contact into Homebase produces a **proper contact card** in the chosen conversation (not an opaque `.vcf`).
- [ ] The user is **optionally prompted to add the contact to the chat-KMP address book**, opening the contact editor pre-filled from the vCard for review before save.
- [ ] Imported phone/email fields are normalized/validated via `ContactFieldValidation` (E.164 + email), with legacy values shown-and-flagged rather than dropped.
- [ ] Works for both `EXTRA_STREAM` (`.vcf` file) and `EXTRA_TEXT` (vCard/text) shares; Homebase reliably appears in the share sheet.
- [ ] Parse failure falls back to sending the raw file rather than dropping the share.
- [ ] A test covering the vCard share intent → parse → contact-attachment (and the address-book pre-fill), mirroring `ShareReceiverActivityIntentTest`.

2002Bishwajeet:
Closes #1255.

## What this does

Sharing a contact into Homebase from the Android share sheet previously fell through as an opaque `.vcf` at every stage — `SharedContentExtractor.copyUriToTemp` had no vCard awareness, `ShareReceiverActivity.convertToAttachmentFiles` hit the `else` branch → `AttachmentPendingFile.File`, and `MessageAttachmentBuilder.build` hit its `else` branch → generic `PayloadFile`.

Now:

1. **Detect** a contact share — mime `text/vcard` / `text/x-vcard` / `text/directory`, a `*.vcf` `EXTRA_STREAM`, or a `BEGIN:VCARD` body in `EXTRA_TEXT`.
2. **Parse** it with a small in-tree reader (`VCardParser`, `homebase-chat/commonMain`, pure Kotlin → JVM-testable).
3. **Send** it as a **new typed message kind** — `ChatProtocol.ChatContactCardMessageDataType = 215`, `MessageContent.ContactCard`, `ContactCardBubble` — following `ADDING_TYPED_MESSAGE_KIND.md`. The picker previews the contact's name instead of a file chip, and neither the media editor nor "New Moment" is offered for a contact.
4. **Offer to add it to the address book.** After the card is sent, an "Add to your contacts?" dialog appears; on confirm the existing `ContactEditSheet` mounts **inside the share activity**, seeded from the vCard, and saves through `saveContactDraft` / `ContactRepository`.

> Note on where the editor lives: `AttachmentOptions.onContactClick` (`:411`) is referenced by the issue as "the existing contact-attachment path", but it is currently a **no-op** — it only closes the attachment sheet, and no contact `AttachmentOption` is even rendered in the LazyRow. There was no contact path to route into, so this PR builds one. Mounting `ContactEditSheet` in the share activity (rather than deep-linking into the contacts tab) keeps the user in the share flow and avoids a nav-event / `AppNavHost` / ViewModel round trip for a single sheet.

### Privacy

`MessageContent.ContactCard.notificationLabel` is overridden to the generic `"Contact"`. A shared contact's name is third-party PII and `notificationLabel` is what rides the wire as the push `unEncryptedMessage`; only `displayLabel` (local UI: conversation-list preview, search index) carries the real name.

## Parser: deliberate scope and its ceiling

`VCardParser` handles only what a contact card renders, across the vCard 2.1 / 3.0 / 4.0 basics every exporter emits:

- **Properties**: `FN`, `N`, `TEL`, `EMAIL`, `ORG`. Everything else is ignored.
- **Line unfolding** — a line starting with SPACE or TAB continues the previous one (RFC 6350 folding).
- **`GROUP.PROPERTY;PARAM=value:value`** splitting, including Apple's `item1.TEL` group prefixes.
- **Value escapes** — `\,` `\;` `\n` `\\`, with `;` component splitting that respects backslash escapes.
- Per-kind caps (10 phones / 10 emails), de-duplication, and surrogate-safe truncation via `String.truncateToCodePoints`.

**Not handled, on purpose:** `QUOTED-PRINTABLE` and `BASE64` encodings, `PHOTO`, `ADR`, `BDAY`, `URL`, `NOTE`, and vCard 4.0 `PREF=` ordering. A quoted-printable value surfaces raw rather than decoded (and its `=`-terminated soft line breaks degrade to ignored lines, not a crash). Add the QP decoder — and an `ADR`/`BDAY` mapping — the first time a real card in the wild needs it; that's ~10 lines and a test, and it isn't worth carrying a vCard dependency for.

## Multi-contact `.vcf`

`VCardParser.parse()` returns **every** block, but the share flow **sends only the first** and logs `"vCard carries N blocks; sending the first only"`. Simplest thing that can't crash or lose the primary contact.

**Ceiling:** sharing a 5-contact `.vcf` silently sends only contact #1. Lifting it means either a pick-which-contact step or N contact-card messages — both are UI decisions worth their own issue, and the list is already returned so the plumbing is there.

**Address-book ceiling:** only the *primary* phone/email seed the editor. Extra `TEL`/`EMAIL` values ride in the chat contact card but aren't imported into the address book — the app's own "new contact" path has the same limit (`saveContactDraft` takes no `additionalPhones`/`additionalEmails`; only the *edit* path writes them, via `ContactOverrideStore`, which needs a `versionTag` a fresh save doesn't return).

## Fallback: a share is never dropped

The descriptor is built **at detection time**, before the picker renders. If the vCard doesn't parse, or parses to nothing renderable (an `ORG:`-only card), `sharedContactDescriptor` stays null and the flow falls straight through to the existing raw-file / text send paths. Covered by three tests.

## Legacy / malformed values

`ContactCardImport` normalizes phones with `ContactFieldValidation.normalizePhone`. A value that still isn't E.164 afterwards is **kept verbatim** — it seeds `PhoneNumberField` (which already handles non-E.164 seeding), renders with `isError` + `contactbook_error_phone`, and the existing Save gate in `ContactEditSheet` keeps the button disabled until it's corrected. Same for an invalid email (`contactbook_error_email`). Nothing is silently dropped.

## Test output

```
homebase-common:jvmTest             tests=480  failures=0 errors=0   (incl. Konsist ArchitectureTest)
homebase-chat:jvmTest               tests=1106 failures=0 errors=0
homebase-core:jvmTest               tests=346  failures=0 errors=0
androidApp:testDebugUnitTest        tests=38   failures=0 errors=0
```

New tests:

- `VCardParserTest` (12) — single contact, multi-block, malformed → nothing, non-E.164 kept verbatim, emoji in the name, surrogate-safe capping, line unfolding + escapes, Apple group prefix + dedup, descriptor round-trip through `MessageContentParser`, broken descriptor → `ContactCard(null)` (never `null`), and the notification-label privacy assertion.
- `ContactCardImportTest` (7) — E.164 normalization, legacy phone kept-and-flagged (`phoneValid == false`, `isSavable == false`), editor prefill, no-structured-`N` fallback, `ORG`-only → no descriptor, invalid email flagged.
- `ShareReceiverVCardIntentTest` (11) — mirrors `ShareReceiverActivityIntentTest`: real `ACTION_SEND` intents through `SharedContentExtractor` + `ContactShareDetector` + `ContactCardImport`, for `EXTRA_STREAM`, `EXTRA_TEXT`, `text/plain`-declared vCards, generic-mime `.vcf` by filename, multi-block, unparseable → fallback, plus negative cases (image share, plain text share).

Compile checks: `:homebase-chat:compileKotlinJvm`, `:homebase-core:compileKotlinJvm`, `:homebase-core:compileKotlinIosSimulatorArm64` (transitively Android + iOS for `homebase-chat`), `:androidApp:compileDebugKotlin` — all BUILD SUCCESSFUL.

`MessageContentParserTest.parse_returns_unknown_for_unrecognized_non_zero_datatype` used `215` as its "future kind" sentinel; bumped to `216` now that 215 is reserved.

## Device repro

```
adb install -r <debug apk>            # package id.homebase.feed.debug
adb push contact.vcf /sdcard/Download/
```

Then either share a contact from the Contacts app (Contact → Share → Homebase) — AOSP Contacts sends `ACTION_SEND` with `type=text/x-vcard` and a `content://` `EXTRA_STREAM` — or drive it directly:

```
# EXTRA_TEXT delivery
adb shell am start -a android.intent.action.SEND -t text/x-vcard \
  -n id.homebase.feed.debug/id.homebase.feed.share.ShareReceiverActivity \
  --es android.intent.extra.TEXT $'BEGIN:VCARD\nVERSION:3.0\nFN:Ada Vance\nTEL:+14155550123\nEMAIL:ada@example.com\nEND:VCARD'
```

What to look for:

1. **Picker** — preview line reads the contact's **name** ("Ada Vance"), not "1 file". No "New Moment" option.
2. **After picking a conversation** — brief spinner, "Sent" toast, then the **"Add to your contacts?"** dialog naming the contact.
3. **"Not now"** — lands in the conversation showing a **contact card bubble**: avatar initials, name, org line, phone and email rows. Conversation list preview shows a contact icon + the name.
4. **"Review and add"** — the contact editor opens **pre-filled** (first name / surname / phone in `PhoneNumberField` / email). Save persists it; the contact appears under Contacts. Then it lands in the conversation.
5. **Legacy number** — share a vCard with `TEL;HOME:0207 946 0018`: the editor shows the number, flags it red with "Use international format…", and **Save is disabled** until corrected.
6. **Fallback** — share a `.vcf` containing garbage: it sends as an ordinary file attachment, exactly as before.
7. `adb shell run-as id.homebase.feed.debug cat files/logs/homebase.log | grep -E "ShareCold|ContactShare"` shows `initShareFlow: contactShare=true|false`.

## Not in this PR

- No attachment-sheet "attach contact" entry — `onContactClick` stays the dead parameter it is today. Composing a contact card from inside a conversation is a separate feature.
- Receivers can't save an incoming contact card to their own address book from the bubble.
- Typed kinds carry no payloads (`sendNewTypedMessage` forces `payloadBundle = null`), so the original `.vcf` isn't attached alongside the card.
- Android only — iOS share-extension parity is untouched.


2002Bishwajeet:
## Device verification — Redmi Note 5 Pro (Android 11), integration build

Fired a vCard share at `ShareReceiverActivity` and walked the flow end to end.

**Detection** — `initShareFlow: extract() files=0 hasText=true textLen=99` → **`initShareFlow: contactShare=true`**.

**Picker preview** — header reads **"Ada Vance"** (the parsed name), not "1 file".

**Send** — "Sent" toast, then the **"Add to your contacts?"** dialog: *"Ada Vance was shared to your chat. Save them to your contacts too?"* / Not now / Review and add.

**Contact card bubble** — renders as a real card, not an opaque `.vcf`: AV initials avatar, "Ada Vance", `Green Dragon`-style org row (Acme Corp), `+14155550123`, `ada@example.com`, each with its own icon.

**Pre-filled editor** — "Review and add" opens the contact editor seeded from the vCard, with the phone correctly decomposed into a `PhoneNumberField` country selector (`+1`) plus national number `4155550123`, and the email seeded. (The `+1` flag renders as Canada — pre-existing `PhoneNumberField` behaviour for the shared NANP calling code, not from this PR.)

### Not device-verified
The `EXTRA_STREAM` (`content://` `.vcf` file) delivery shape. `am start` from the shell cannot grant read permission on a MediaStore URI it does not own, and the only other route was creating a contact that would sync into a real Google account — not worth the side effect for a path that differs only in how the bytes are obtained before reaching the same parser, and that `ShareReceiverVCardIntentTest` already covers. Worth one manual pass from the Contacts app share sheet before merge.

Issue #854, opened 2026-06-25
iOS share: sharing an image into Homebase sends it as a sticker (auto-sticker transparency heuristic fires without user intent)


Seifert69:
## Summary
Sharing an image **into** Homebase via the iPhone share sheet sends it as a **sticker** (frameless cut-out) instead of a normal image. The user never chose "send as sticker" — it just arrives stickered.

![Shared image rendered as a sticker](https://raw.githubusercontent.com/homebase-id/chat-kmp/issue-assets/docs/issue-assets/ios-share-image-sticker/shared-image-rendered-as-sticker.png)

## Environment
- Platform: iOS (share extension / share sheet → Homebase)
- Build: debug
- Surface: incoming-share → send to a conversation

## Steps to reproduce
1. In another iOS app, share an image to Homebase.
2. Pick a conversation; it sends.
3. **Actual:** the image renders as a sticker (no bubble frame). **Expected:** a normal image.

## Root cause (prime suspect — not "fix the MIME")
Shared images flow through the normal attachment builder, which **auto-classifies stickers by transparency**:

`MessageAttachmentBuilder.build()` (`homebase-chat/.../services/builder/MessageAttachmentBuilder.kt` ~50-61):
```kotlin
val isSticker = attachment.forceSticker ||
    ((format == "image/png" || format == "image/webp") &&
        ImageUtils.hasNonOpaquePixels(thumbs.sourceBytes))
```
- The share path sets **`forceSticker = false`** (`sharedMediaAttachment()` builds `FileImage` without it; default false in `ConversationListUiState.kt` ~334-340). So the user did NOT opt in.
- But the **second clause auto-stickers any PNG/WebP image with non-opaque pixels** — format is re-detected from the actual bytes (`ImageFormatDetector.detectFormat`), independent of any MIME label. So a shared PNG/WebP that has any transparency (alpha channel, anti-aliased edges) silently becomes a sticker.

**Why the audit's "iOS share extension hardcodes `image/jpeg`" angle is NOT the fix:** the KMP side ignores that MIME and re-detects format from magic bytes, then applies the transparency heuristic. Correcting the iOS MIME alone wouldn't stop a genuine PNG/WebP-with-alpha from auto-stickering. (It's still a real secondary data-correctness bug — see below — but not the cause here.)

## What to verify (evidence before fix)
Confirm which trigger fired for the reported image:
[...]

PR #1261, opened 2026-08-11
feat(share): recognise a shared vCard on the descriptor path, so iOS sends a contact card

2002Bishwajeet:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

> **Stacked on #1260** — review that one first. This layer does not compile without it.

## Why

#1260 made a shared vCard become a contact card instead of an opaque `.vcf` attachment — but only on Android, where the interception lives in `ShareReceiverActivity`. iOS never goes through that code.

iOS shares travel a different route: the share extension writes a `SharedContentDescriptor` into the App Group and the main app picks it up in `MessageActionsHandler.processPendingSharedContent`. A shared `.vcf` misses every specific branch in `SharedContentSaver.swift` and lands in the generic `UTType.data` one, so iOS sends the raw file.

Verified against the system UTI database rather than assumed:

```
public.vcard  conforms to public.text:        true
              conforms to public.plain-text:  false   <- skips the plainText branch
              conforms to public.data:        true    <- lands in the generic data branch
UTType(filenameExtension: "vcf")  ->  public.vcard  ->  text/vcard
```

That last line is what `SharedContentSaver.mimeTypeForExtension` runs, so the descriptor arrives as `fileNames = ["Alice.vcf"]`, `mimeTypes = ["text/vcard"]`.

## What

`SharedVCardDetector` (homebase-chat, commonMain) recognises a vCard on the descriptor path, and `processPendingSharedContent` sends `MessageContent.ContactCard` instead of a file. Detection semantics mirror Android's `ContactShareDetector`: the three vCard mime types with any `;charset=` stripped, or a `.vcf` filename, or a vCard body in the share text; 1 MiB read cap; first block of a multi-contact card.

Because it hooks the descriptor rather than an Android `Intent`, it also covers any other descriptor-driven entry point — `ShareCacheStorage` has a `jvmMain` actual.

**Layering.** `ContactCardImport.toDescriptor` needs `ContactFieldValidation` for E.164, and both live in `:homebase-core`, which `:homebase-chat` does not depend on (`homebase-chat/build.gradle.kts:51-58`). So the conversion is injected: `fun interface VCardDescriptorFactory` declared in homebase-chat, SAM-bound in `AppModule` to `ContactCardImport::toDescriptor`. No new module dependency, no relocated files, nothing duplicated.

## Fallback contract

Anything that isn't a parseable vCard — wrong mime, oversize, unreadable, parses to nothing renderable, or a descriptor the importer rejects — falls through to the existing file/text send completely unchanged. A share is never dropped and never becomes a blank bubble.

One place this deliberately diverges from Android: **a multi-file share is left alone.** The contact card replaces the *whole* share, so anything shared alongside the vcf would vanish. Android's detector picks the first vcf out of the share and discards the sibling files with it — `photo.jpg` + `Alice.vcf` sends only the card. Here the contact-card path requires the vcf to be the share's only file, and the text-vCard path requires there be no files at all; a mixed share goes down the ordinary multi-file send with everything intact.

The Android path has the same defect and is worth a follow-up — this PR does not touch it, since #1260 is device-verified and shouldn't be disturbed.

## Tests

18 in `SharedVCardDetectorTest`, 2 in `SharedVCardDescriptorPathTest`. JVM.

The seam is the detector — the same split #1260 used (it tested `ContactShareDetector`, not the Activity), because `ChatMessageSenderService` is a concrete class and the repo ships no mocking framework. `SharedVCardDescriptorPathTest` crosses the module seam with the *real* `ContactCardImport` factory, so the injected-conversion wiring is covered rather than just the chat-side half.

Both multi-file regression tests were checked against the pre-fix detector and fail there, so they pin behaviour rather than merely passing.

```
homebase-chat: tests=1124 failures=0 errors=0 skipped=6
homebase-core: tests=348  failures=0 errors=0 skipped=2
```

## Not in this PR

- **The sender-side "add to your contacts?" prompt.** On Android that's a Compose surface owned by `ShareReceiverActivity` (`ShareScreenState.ContactSent`). The descriptor path has no equivalent host, so porting it means a new `ConversationListUiDialog` arm — its own change.
- **Any Swift.** `SharedContentSaver.swift:157` handles only `item as? URL` in the data branch, while the image branch above it also handles in-memory `Data` (added for #854). If Contacts.app vends the vCard as `Data`, the extension captures nothing and this layer never fires. I have no evidence either way yet, so no speculative fix — see the verification note below.

## Verification status

JVM tests prove the descriptor path produces a contact card. They do **not** prove iOS populates the descriptor in the first place — that needs a signed-in simulator or device build, sharing a contact from Contacts.app and confirming a card renders. Until that runs, treat the iOS end-to-end as unverified.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/contactcard/SharedVCardDetector.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/contactcard/SharedVCardDetector.kt
@@ -1,0 +1,94 @@
+package id.homebase.chat.contactcard
+
+import co.touchlab.kermit.Logger
+import id.homebase.core.share.SharedContentDescriptor
+import kotlin.coroutines.cancellation.CancellationException
+
+/**
+ * Recognises a contact share that arrived as a [SharedContentDescriptor] and turns it into a
+ * [VCardContact].
+ *
+ * This is the non-Android delivery shape: the iOS share extension writes the descriptor to the
+ * App Group, and `ShareCacheStorage` has a jvmMain actual too. A `.vcf` misses the extension's
+ * image/movie/url/plainText branches — `public.vcard` conforms to `public.text`, not
+ * `public.plain-text` — so it lands as a generic `FILE` with `mimeTypes = ["text/vcard"]`.
+ *
+ * Semantics mirror Android's `ContactShareDetector`. Returns null for anything that isn't a
+ * parseable contact; the caller then sends the raw file/text as before, so a share is never
+ * dropped. A multi-contact `.vcf` yields its FIRST block; the rest are logged and ignored.
+ */
+object SharedVCardDetector {
+
+    private const val TAG = "ContactShare"
+
+    /** Bigger than any plausible vCard; stops a mislabelled 500 MB file being slurped into RAM. */
+    const val MAX_VCARD_BYTES = 1024L * 1024
+
+    private val VCARD_MIME_TYPES = setOf(
+        "text/vcard",
+        "text/x-vcard",
+        // Pre-RFC-6350 exporters still emit text/directory for a vCard.
+        "text/directory",
+    )
+
+    suspend fun detect(
+        descriptor: SharedContentDescriptor,
+        fileSize: (fileName: String) -> Long,
+        readFileText: suspend (fileName: String) -> String,
+    ): VCardContact? {
+        // A lone vcf only: the contact card replaces the whole share, so anything shared
+        // alongside it would be silently dropped.
+        val vCardFile = descriptor.fileNames.singleOrNull()?.takeIf {
+            isVCardMime(descriptor.mimeTypes.firstOrNull()) || it.endsWith(".vcf", ignoreCase = true)
+        }
+
+        val source = when {
+            vCardFile != null -> readCapped(vCardFile, fileSize, readFileText)
+            // Some senders put the card body straight in the share text.
+            descriptor.fileNames.isEmpty() -> descriptor.text
+            else -> null
+        } ?: return null
+
+        if (!VCardParser.looksLikeVCard(source)) return null
+
+        val cards = VCardParser.parse(source)
+        if (cards.isEmpty()) {
+            Logger.w(tag = TAG) { "vCard detected but nothing parsed; falling back to raw share" }
+            return null
+        }
+        if (cards.size > 1) {
+            Logger.i(tag = TAG) { "vCard carries ${cards.size} blocks; sending the first only" }
+        }
+        return cards.first()
+    }
+
+    private suspend fun readCapped(
+        fileName: String,
+        fileSize: (String) -> Long,
+        readFileText: suspend (String) -> String,
+    ): String? = try {
+        if (fileSize(fileName) > MAX_VCARD_BYTES) {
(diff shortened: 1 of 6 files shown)


2002Bishwajeet:
Filed #1263 for the iOS share-extension gap referenced in the *Not in this PR* section — `SharedContentSaver.swift` drops a file item vended as `Data` rather than a `URL` in the generic branch. It is a blocker for *verifying* this PR on iOS, not for reviewing or merging it: this layer is correct on any path, and its unit tests stand on their own.

2002Bishwajeet:
## iOS device-verified — a shared vCard now arrives as a contact card

Ran on the iPhone 17 Pro simulator (iOS 26.5), signed in, with a **freshly built** app from this stack's tip installed with its `ShareExtension.appex`. Replaces my earlier draft finding, which was based on a stale Jul-30 build and was wrong.

### Result: the full path works

Contacts.app → *Save to Files* → share the `.vcf` from Files → **Homebase** → pick conversation → Send → deep link → a **contact card bubble** renders (`KB / Kate Bell`, `4155553695`, `5555648583`, `kate-bell@mac.com`). Not a `.vcf` attachment.

The pending descriptor is consumed afterwards — `shared_content.json` and `shared_files/` are both gone, so `cleanup()` ran.

### The descriptor this PR was written for, captured from a real device

```json
{
  "contentType": "FILE",
  "fileNames": ["share_1786464846652_1868.vcf"],
  "mimeTypes": ["text/vcard"],
  "targetConversationId": "e4ef2382-…"
}
```

Single file, `text/vcard` — matches `SharedVCardDetector` on both the mime rule and the `.vcf` filename rule, and satisfies the lone-vcf guard that stops a co-shared file being dropped. Exactly the shape predicted from the `UniformTypeIdentifiers` analysis in the PR body.

Payload was a genuine Apple vCard 3.0 (`PRODID:-//Apple Inc.//iPhone OS 26.5//EN`).

Note the phones arrive **not** in E.164 — Contacts.app vends US national-format numbers with no country code, and `normalizePhone` can't synthesise one. They're kept verbatim rather than dropped, which is the documented contract: the editor seeds them, flags them, and blocks Save until corrected.

### One real limitation, and it is not this PR's

**Homebase is not offered in the share sheet when sharing directly from Contacts.app** — only *More* and *Save to Files* appear. Sharing the same vCard from **Files** does offer Homebase. Controlled on one build, minutes apart:

| source | Homebase offered |
|---|---|
| Safari (URL) | yes — proves the extension is registered |
| Contacts.app → Share Contact | **no** |
| Files.app → share the saved `.vcf` | yes |

So the detection logic here is reachable on iOS, just not by the route a user would take first. The cause is upstream in `ShareExtension/Info.plist`'s activation rule, which uses the dictionary shorthand; `public.vcard` conforms to `public.text` and `public.data` but not `public.plain-text`, and empirically none of the generated predicates match it. That is a separate plist change and should be verified with the same share-sheet test rather than assumed.

### #1263 is not implicated on this path

The extension captured the file correctly (`shared_files/share_…vcf`, 335 bytes), so the `Data`-vs-`URL` gap in `SharedContentSaver`'s generic branch did not bite here. #1263 stands on its own merits for other senders.

Issue #1265, opened 2026-08-11
Decide whether a contact card can carry a Homebase identity (odinId), not just vCard fields


2002Bishwajeet:
## Summary

A design question, not a defect. Filed so it gets decided rather than drifting into a picker PR by accident.

The contact card added in #1255 / #1260 is vCard-shaped: display name, given/surname, organization, phones, emails. In a Homebase context the more valuable thing to pass along is often not a phone number but an **identity** — "here's my friend, connect with them".

`ContactCardDescriptor` has no `odinId` field today. `ContactBookEntry` does (`odinId: String?`), and `ContactBookEntry.SOURCE` already distinguishes `CONNECTION` ("public", synced from a Homebase identity profile) from `IMPORTED` and `MANUAL`. So the app already knows which of your contacts *are* identities; the card just can't say so.

## What it would change

An optional `odinId` on the descriptor lets the receiving bubble offer **Connect** alongside (or instead of) **Add to contacts**, and lets the card render the identity's live published avatar rather than initials.

This is legitimate under the descriptor rules in `ADDING_TYPED_MESSAGE_KIND.md` — the prohibition there is on duplicating **envelope** fields (`createdByOdinId`, the *author*). The *subject* of the card is genuinely new data the envelope does not carry.

## Open questions to settle before anyone builds it

1. **Same message kind or a new one?** An optional field on dataType 215 keeps one bubble and one parser, but means "contact card" covers two fairly different things. A separate kind is cleaner semantically and costs another dataType integer plus a second bubble.
2. **What does an older client do?** A receiver on the current build ignores an unknown field and shows a normal contact card — acceptable degradation, worth confirming against the parser's nullable-descriptor contract.
3. **Does a shared identity card leak anything?** The odinId is public, but a card asserting "X knows Y" is a social-graph disclosure the sender may not intend. Probably fine for an explicit share; worth a deliberate decision rather than an accident.
4. **Does it belong in the picker at all?** If the picker in the sibling issue lists both identity-backed and plain contacts, sending an identity-backed one *without* the odinId would quietly discard the most useful part.

Question 4 is why this is filed now: it constrains the picker's design even if the answer is "not yet".

2002Bishwajeet:
## This is now forced by two concrete requirements

Filed as "decide later", but two requests make it blocking rather than optional:

1. **Reply to / message a contact.** You can't open a conversation against a card that carries only a name, phones and emails. There is nothing to target.
[...]

PR #1266, opened 2026-08-11
feat(chat): add a Contact option to the attachment sheet

### LINKED ISSUE #1264: Add a Contact option to the chat attachment sheet (onContactClick is a dead parameter)
## Summary

You can receive a contact card and import it, and (after #1255 / #1260 / #1261) you can share one *into* a conversation from the OS share sheet. You cannot originate one from inside a chat. The attachment sheet has no Contact option.

## The socket already exists and does nothing

`AttachmentOptions` takes an `onContactClick` parameter (`AttachmentOptions.kt:411`), and `ConversationContent.kt:1746` implements it as:

```kotlin
onContactClick = {
    showAttachmentSheet = false
},
```

No `AttachmentOption` row renders it — the sheet shows Gallery, File, Location, Event and Groodle. So the parameter is threaded, called by nothing, and closes the sheet if it ever were.

This is worth flagging because #1255 was written on the assumption that this was a working integration point ("route it into the existing contact-attachment path, `AttachmentOptions.onContactClick`"). It never was, which is why #1255 became a new typed message kind instead of a routing change.

## What to build

A **Contact** row in the attachment sheet that picks from the **Homebase contact book** and sends a `MessageContent.ContactCard` — the same message kind #1260 added, so the bubble, the parser, and the receiver-side import all already work.

Follow the **Location** option exactly; it is the established pattern for an attachment that needs a screen owned by `:homebase-core`:

```
AttachmentOption tap
  -> onUiAction(ConversationListUiAction.OpenShareLocation(conversationId))   ConversationContent.kt:1752
  -> ConversationListViewModel:1117
  -> sendEvent(ConversationListUiEvent.NavigateToShareLocation(id))
  -> AppNavHost routes to the core-owned screen
```

This matters because `:homebase-chat` does not depend on `:homebase-core` (`homebase-chat/build.gradle.kts:51-58`), so the picker cannot live in the attachment sheet's module. `:homebase-core` *does* depend on `:homebase-chat`, so a core-owned picker can call `ChatMessageSenderService.sendNewTypedMessage` directly once a contact is chosen.

`CircleMemberPickerScreen` is an existing contact picker to model on.

## Mapping

`ContactBookEntry` already carries everything `ContactCardDescriptor` wants:

| descriptor | entry |
|---|---|
| `displayName` / `givenName` / `surname` | same names |
| `phones` | `phone` + `additionalPhones` |
| `emails` | `email` + `additionalEmails` |
| `organization` | check — `ContactBookEntry` may not have one; leave blank if so |

Reuse `ContactCardImport`'s normalization (`ContactFieldValidation.normalizePhone` for E.164, `truncateToCodePoints` for the caps) rather than writing a second conversion, and respect `ContactCardDescriptor.MAX_VALUES_PER_KIND`.

## Explicitly not in scope

**A device address-book picker.** That needs a runtime `CONTACTS` permission on both platforms plus an `expect`/`actual` picker, to deliver what the OS share sheet already does well — Contacts → Share → Homebase works today. Bad trade.

**Sharing a Homebase identity** rather than a vCard-shaped contact — that changes what the card *means* and is tracked separately.

2002Bishwajeet:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

Closes #1264. **Stacked on #1261 → #1260** — review those first; this does not compile without them.

## Why

You could receive a contact card, and (after #1260/#1261) share one *into* a conversation from the OS share sheet. You could not originate one from inside a chat.

`AttachmentOptions` already took an `onContactClick` parameter and `ConversationContent.kt:1746` implemented it as `showAttachmentSheet = false` — no row rendered it, so it was dead. That dead parameter is what #1255 was originally written against ("route it into the existing contact-attachment path"), which is why that issue turned into a whole new message kind instead of a routing change.

## What

A **Contact** row in the attachment sheet → a contact-book picker → sends `MessageContent.ContactCard`, the kind #1260 added. Bubble, parser and receive side already work, so this is purely the outbound affordance.

Routing mirrors the **Location** option hop for hop — each new line sits beside its Location twin:

```
AttachmentOption tap
  -> ConversationListUiAction.OpenShareContact(conversationId)   ConversationContent.kt:1749
  -> ConversationListViewModel.kt:1121
  -> ConversationListUiEvent.NavigateToShareContact(id)
  -> ConversationListScreen.kt:216
  -> AppNavHost.kt:1010 -> Route.ShareContact
```

Same `Uuid`→`String`→`Uuid.parse` threading and the same manual Koin `viewModel { params -> }` block, for the same reason Location has one (`viewModelOf` would try to autowire the runtime parameter).

**Where the picker lives.** `:homebase-chat` does not depend on `:homebase-core` (`homebase-chat/build.gradle.kts:51-58`), so the picker cannot live next to the attachment sheet — the contact book is all in core. It sits beside `CircleMemberPickerScreen`, which it is modelled on. Since core *does* depend on chat, the picker's ViewModel calls `sendNewTypedMessage` directly, copying `ShareReceiverActivity.sendContactCard`'s shape.

**Select-then-send, not tap-to-send.** An accidental tap would otherwise drop a message into someone's chat with no undo.

## Mapping

`ContactBookEntry` has **no organization field** — confirmed, not assumed (`ContactBookEntry.kt:43-90`). `descriptor.organization` is left blank and a test pins that; `shortBio` / `status` were not repurposed into it. `additionalName` likewise has no descriptor slot and is dropped.

`phones` = `phone` + `additionalPhones`, `emails` = `email` + `additionalEmails`, both normalized to E.164 through `ContactFieldValidation` and capped at `MAX_VALUES_PER_KIND`. This is a second `ContactCardImport.toDescriptor` overload — the private normalization helpers were refactored from `VCardContact` extensions to `List<String>` extensions so the vCard path and the picker path share one rule set rather than drifting.

## Tests

`ShareContactPickerTest` (12) and 2 added to `AttachmentOptionsTest` (now 5; the 3 pre-existing cases are untouched, diff is additive only).

Full gate green: **1,966 tests, 0 failures**. Konsist `ArchitectureTest` passes and uses `Konsist.scopeFromProject()`, so it genuinely scans the new core and chat composables rather than only `homebase-common`.

## Known exposure

**This repo has no Koin graph verification test.** The new DI registration is validated only at runtime — the same exposure `ShareLocationViewModel` already carries. Compiling proves the types line up, not that the graph resolves, so opening the attachment sheet is a worthwhile smoke test before merge. The registration deliberately copies the proven Location pattern to minimise the risk.

Not device-verified — the picker needs a populated contact book. The send path is the same `sendNewTypedMessage` call the already-verified share flow uses.

## Not in scope

`odinId` on the card (#1265). The picker doesn't foreclose it: `ShareContactCandidate` holds the whole `ContactBookEntry`, which carries `odinId`, and the descriptor is built in exactly one place, so adding it later is one property plus one line in the mapper. Nothing was pre-built. The open question there is what the *inbound* vCard path puts in that field, since a vCard has no odinId.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListScreen.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListScreen.kt
@@ -135,6 +135,7 @@ fun ConversationListScreen(
     onNavigateToLiveLocationMap: () -> Unit = {},
     onNavigateToLocationSetup: () -> Unit = {},
     onNavigateToShareLocation: (conversationId: String) -> Unit = {},
+    onNavigateToShareContact: (conversationId: String) -> Unit = {},
     onNavigateToContactInfo: (odinId: String) -> Unit,
     onNavigateToConversationSettings: (conversationId: String) -> Unit,
     onNavigateToGroupSettings: (conversationId: String) -> Unit,
@@ -218,6 +219,11 @@ fun ConversationListScreen(
             is ConversationListUiEvent.NavigateToShareLocation -> {
                 viewModel.eventConsumed()
                 onNavigateToShareLocation(event.conversationId)
+            }
+
+            is ConversationListUiEvent.NavigateToShareContact -> {
+                viewModel.eventConsumed()
+                onNavigateToShareContact(event.conversationId)
             }
 
             is ConversationListUiEvent.NavigateToContactInfo -> {
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiAction.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiAction.kt
@@ -422,6 +422,11 @@ sealed interface ConversationListUiAction {
     /** Open the full-screen share-location screen (attachment sheet → Location). */
     data class OpenShareLocation(val conversationId: Uuid) : ConversationListUiAction
 
+    /** Open the contact book picker that sends a contact card (attachment sheet → Contact). The
+     *  contact book lives in :homebase-core, which this module can't see — hence a navigation
+     *  event rather than an in-place sheet. */
+    data class OpenShareContact(val conversationId: Uuid) : ConversationListUiAction
+
     /** Open the location setup screen — from the "set up location" prompt shown when a live share
      *  can't start because location isn't ready. */
     data object OpenLocationSetup : ConversationListUiAction
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiEvent.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListUiEvent.kt
@@ -38,6 +38,9 @@ sealed interface ConversationListUiEvent {
     /** Open the full-screen share-location screen for a conversation (attachment sheet → Location). */
     data class NavigateToShareLocation(val conversationId: String) : ConversationListUiEvent
 
+    /** Open the contact book picker for a conversation (attachment sheet → Contact). */
+    data class NavigateToShareContact(val conversationId: String) : ConversationListUiEvent
+
     /** Open the location setup screen (from the "set up location" prompt before a live share). */
     data object NavigateToLocationSetup : ConversationListUiEvent
 }
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListViewModel.kt
@@ -1122,6 +1122,10 @@ class ConversationListViewModel(
                 sendEvent(ConversationListUiEvent.NavigateToShareLocation(action.conversationId.toString()))
             }
 
+            is ConversationListUiAction.OpenShareContact -> {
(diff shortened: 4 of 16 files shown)


2002Bishwajeet:
## Device-verified on a Redmi Note 5 Pro (AOSP 11)

Built from the stack tip and installed on hardware. Full chain works:

1. **Contact row renders** in the attachment sheet — Gallery · File · **Contact** · Location.
2. **Picker opens** — "Send a contact" with search and the contact book listing identities.
3. **Select shows a checkmark and reveals the Send FAB** — confirming select-then-send rather than tap-to-send, so a stray tap can't drop a message into a conversation with no undo.
4. **Send returns to the conversation and the contact card bubble appears.**

### This also closes the DI exposure flagged in the PR body

The PR notes that this repo has no Koin `checkModules()` test, so the new `ShareContactPickerViewModel` registration is validated only at runtime — CI compiling proves the types line up, not that the graph resolves. Opening the attachment sheet and then the picker exercises exactly that path. No crash, and the picker populated from the real contact book.

### Note on a name-only contact

The contact I sent (`app.tester.demo.rocks`) has no phone or email in the book, so the card rendered as avatar + name only. That is correct — `ContactCardDescriptor.isValid()` accepts a name-only card and the bubble handles it — but worth knowing that sending an identity with no stored phone/email produces a card whose only content is the name. If that is undesirable, the picker is where it would be filtered, not the bubble.

### Not covered here

The receiving side's **Save to contacts** flow is not exercised by this PR — it belongs to the stacked bubble work — and could not be device-tested at all, because Save only appears on a card received from a *different* identity and both available devices are signed into the same account.

PR #1310, opened 2026-08-18
Web two pane layout

toddmitchell:
Running the app in a browser, chat stayed single-column no matter how wide the window
was — no conversation list on the left with the thread on the right the way desktop does.

## Cause

`ConversationListScreen.kt` gated its `ListDetailPaneScaffold` on `isDesktop() && width >= 800dp`.
`isDesktop()` is hardcoded `false` on wasm, so `maxHorizontalPartitions` was always 1 and the
width check never got a chance to run. Moments was gated the same way.

The app shell was already width-driven — `AppNavHost`'s navigation rail is a pure width check —
so chat and moments were the outliers.

## Approach

Rather than sprinkle `isDesktopOrWeb()` at each site, this adds one helper in `homebase-common`:

```kotlin
@Composable
fun isExpandedLayout(): Boolean =
    isDesktopOrWeb() && currentWindowAdaptiveInfo().windowSizeClass
        .isWidthAtLeastBreakpoint(ExpandedWidthBreakpointDp)
```

The expression was already duplicated in three places at two different breakpoints (800 in the
chat scaffold, 840 in both moments and the conversation list pane), so it was drifting. All three
now share one definition at 800dp.

Web belongs on the same side of this check as desktop: both are pointer-driven and freely
resizable. The width term still keeps wide phones and tablets on a single column, and a phone
browser is well under 800dp so it stays single-pane too.

## Changes

| Site | Before | After |
|---|---|---|
| `ConversationListScreen.kt:767` | `isDesktop() && >=800dp` | `isExpandedLayout()` |
| `ConversationListScreen.kt:869` | `BackHandler` on `!isDesktop()` | `!isExpanded` |
| `ConversationListPane.kt:104` | `>=840dp` (container color) | `isExpandedLayout()` |
| `MomentsScreen.kt:201` | `isDesktop() && >=840dp` | `isExpandedLayout()` |
| `MomentsScreen.kt:206` | Reels preference coerced to Timeline on desktop | `isDesktopOrWeb()` |
| `MomentsScreen.kt:628` | feed autoplay off on desktop | `isDesktopOrWeb()` |
| `MomentsScreen.kt:1512` | Reels hidden from desktop view menu | `isDesktopOrWeb()` |

## Behavior changes beyond web

- **The chat back handler is now layout-gated, not platform-gated.** It previously refused to pop
  the detail pane on desktop unconditionally; a desktop user who dragged the window below the
  breakpoint got a back gesture that did nothing. It now follows the layout you are actually in.
- **Moments splits at 800dp instead of 840dp.** Chat is unchanged at 800.
- **The 800–840dp band is no longer inconsistent in chat** — the scaffold was two-pane there while
  the list pane still painted its single-pane container color.

## Tradeoff worth a look

The three moments flags are gated on the platform, not the width, so a **phone browser** also loses
Reels and feed autoplay. `isWeb()` cannot distinguish a phone browser from a desktop one. This
matches what the existing `isDesktopOrWeb()` sites already do (Enter-to-send and the formatting
toolbar in `MessageInputBar`), so it is at least consistent — but autoplay in particular is the one
where a metered or battery-constrained browser tab argues the other way. Easy to move to a
width-based gate if reviewers prefer that.

## Not included

Hover affordances on message bubbles (reply / react / info) are still `isDesktop()`-gated, so a
mouse user in the browser does not get them. Real gap, separate and larger change.

## Testing

`homebase-common`, `homebase-chat`, `homebase-core` compile on JVM, wasmJs, Android and
iosSimulatorArm64. `:homebase-common:jvmTest` and `:homebase-chat:jvmTest` pass, including the
Konsist architecture test.

Pre-existing and unrelated: `:homebase-chat:compileTestKotlinIosSimulatorArm64` fails on
`VCardParserTest.kt:165` (`Name contains illegal characters: ","`). Confirmed identical on `main`
with these changes stashed.

Not yet exercised in a running browser — worth a `./gradlew webApp:wasmJsBrowserDevelopmentRun
--no-configuration-cache` and a window drag across the 800px line before merging.

### DIFF
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListScreen.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/conversationlist/ConversationListScreen.kt
@@ -66,7 +66,7 @@ import id.homebase.core.connections.ConnectRequestViewModel
 import id.homebase.core.localization.TranslationUtil
 import id.homebase.core.ui.theme.HomebaseTheme
 import id.homebase.core.util.getUriHandler
-import id.homebase.core.util.isDesktop
+import id.homebase.core.util.isExpandedLayout
 import id.homebase.core.widget.DialogButtons
 import id.homebase.core.widget.DialogCard
 import id.homebase.core.widget.DialogText
@@ -764,7 +764,7 @@ fun ConversationListUi(
 ) {
     val windowAdaptiveInfo = currentWindowAdaptiveInfo()
     val defaultDirective = calculatePaneScaffoldDirective(windowAdaptiveInfo)
-    val isExpanded = isDesktop() &&  windowAdaptiveInfo.windowSizeClass.isWidthAtLeastBreakpoint(800)
+    val isExpanded = isExpandedLayout()
     val scaffoldDirective = PaneScaffoldDirective(
         maxHorizontalPartitions = if (isExpanded) 2 else 1,
         horizontalPartitionSpacerSize = 0.dp, // Remove the white border
@@ -866,7 +866,7 @@ fun ConversationListUi(
         scope.launch {
             if (messagesUiState.fullScreenOverlay != null) {
                 onUiAction(ConversationListUiAction.CloseFullScreenOverlay)
-            } else if (!isDesktop()) {
+            } else if (!isExpanded) {
                 scaffoldNavigator.navigateBack(BackNavigationBehavior.PopUntilContentChange)
             }
         }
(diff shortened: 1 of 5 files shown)


toddmitchell:
Closing as superseded by #1312.

Both commits here are already ancestors of that branch:

- `0089a95` Split chat and moments into two panes on web
- `47bfcb0` Give moments the desktop treatment on web

#1312 carries all five files this PR touches, plus `MomentsViewModel.kt`, `main.wasm.kt` and `index.html`. Merging both would leave the second as a no-op.

Nothing from this PR has reached `main` yet — the full diff still applies cleanly there — so the work is not lost, just consolidated into #1312.

PR #1314, opened 2026-08-18
Bundle a colour emoji font for web

toddmitchell:
Emoji render as missing glyphs in the wasm app.

On Android, iOS and desktop, Skia resolves emoji through the OS font manager. In
the browser there is no system font to fall back on, and the repo ships no font of
its own — there is no `FontFamily` anywhere in the codebase and no fonts directory.
The emoji *data* is all present (`emoji_data.json`, the picker); only the glyphs
are missing.

## What this adds

Noto Color Emoji (**COLRv1**, 4.8 MB) behind an `expect`/`actual` seam:

```kotlin
@Composable expect fun emojiFontFamily(): FontFamily?
```

- **wasm** returns the bundled family
- **Android / JVM / native** return `null` — their typography is provably unchanged

`Typography.withFontFamily(family)` applies it across all 15 text styles and returns
`this` untouched when the family is null, so `HomebaseTheme` is a no-op everywhere
except web.

## Bundle impact is web-only

The font is in `wasmJsMain/composeResources/font/`, not `commonMain`. Verified after
building: the generated accessor lands only in `wasmJsMainResourceAccessors`, so the
Android and iOS bundles do not carry the 4.8 MB.

## Why COLRv1

Vector glyphs — sharp at any size, and about half the size of the CBDT bitmap build.
Confirmed the file is genuinely COLRv1 (`COLR version=1`, no `CBDT`/`sbix` tables).
Swapping to the bitmap build means replacing the file in that directory; no code
change.

## What the browser spike must answer

The emoji family is applied as *the* typography font, so ordinary text depends on
Skia's paragraph shaper falling back to its default font manager for glyphs Noto
Color Emoji lacks — it has no Latin. That is standard `FontCollection` behaviour and
I expect it to hold, but **it is unverified**: it cannot be exercised without running
the browser, and I would rather flag it than have it found in review.

Check in this order:

1. **Does ordinary Latin text still render?** If it goes blank or tofu, the fallback
   assumption is wrong.
2. **Do simple emoji render in colour?** 🎉
3. **Do ZWJ sequences render as one glyph?** 👨‍👩‍👧 — separates real support from partial.

If step 1 fails, the fix is to bundle a text font (~200 KB) and use
`FontFamily(textFont, emojiFont)` so the order is explicit rather than inferred —
a one-line change to the same seam.

## Notes

- No LFS in this repo, and it already tracks 100 MB ffmpeg binaries directly, so a
  4.8 MB font follows existing convention.
- Compiles on JVM, Android, wasmJs and iosSimulatorArm64; `homebase-common` JVM tests pass.
- Consider lazy-loading the font so it never blocks first paint — not done here, and
  worth measuring against the existing wasm bundle size before deciding.


toddmitchell:
Pushed a second, unrelated commit onto this branch (`e629f59`): temporary logging for the Moments portrait-photo re-crop on web.

It does not belong in this PR long-term — flagging so it is stripped before merge, or split out if the font work lands first.

**What it does.** The feed cell uses `ContentScale.Crop`, so the symptom is not distortion; the cell locks to a wrong aspect and chops the top and bottom. The height comes from `aspectRatioFor`, which reads `previewThumbnail ?: thumbnails.lastOrNull()`. Two candidates fit:

1. metadata arrives late, so the first composition falls back to `1f` (square) — plausible on web, where the descriptor is fetched over the network, and consistent with it being fine on phone
2. `previewThumbnail` wins the elvis and carries worse dimensions — the tiny thumb is capped at 20px (`ThumbnailGenerator.kt:40`)

The log is keyed on `aspect`, so it fires on the initial value and on every change. **A second line for the same payload key is the bug reproducing**, and the fields show which source supplied the number.

Filter the browser console on `MomentAspect`.

toddmitchell:
Closing — superseded by #1315.

This branch accumulated the exploration: bundling the font, applying it as the typography face (which broke all non-emoji text), reverting that, scoping it to spans, then sweeping too many surfaces and narrowing back. Useful to arrive at the answer, not useful to review.

#1315 is one commit off `main` with only what is needed — the font, the span helper and its tests, and eight lines of wiring across the three intended surfaces (emoji selector, chat screen, moment comments).

The Moments squish logging (`e629f59`) also lives on this branch and is not in #1315, since it is unrelated to emoji. It is recoverable from that SHA if wanted.

PR #1340, opened 2026-08-21
Pick emoji by typing ":" in the composer

toddmitchell:
richeditor already ships a trigger system built for @mentions — Trigger, activeTriggerQuery, and a suggestions popup. Registering ':' on it gets the detection for free, including the word-boundary rule that keeps "10:30" and "https://" from opening the popup.

Commit goes through replaceTextRange, not insertToken: insertToken splices an atomic styled Token span, and the composer ships toMessageMarkdown(), so an emoji has to land as literal text. That rules out the stock TriggerSuggestions composable, whose onSelect is typed to return a Token — hence the local popup.

It anchors to the editor's bounds rather than the caret. Both textFieldWindowPosition and textLayoutResult are internal, and the caret rect that is public is relative to the inner field, offset from the decoration box by the leading icon.

Enter needed a guard. Enter-to-send lives in an onPreviewKeyEvent on the editor's outer modifier, and preview events run root-to-leaf, so that handler sees Enter before the editor does; without giving the popup first refusal it sends ":sm" instead of committing the highlighted emoji.

Ranking is label/tag based because the bundled emojibase dump carries no shortcodes field. Tiers run label-exact, label-prefix, word-prefix, tag-exact, tag-prefix, substring, then a separator-insensitive pass so ":thumbs_up" and ":thumbsup" reach "thumbs up". Promoting tag-exact above label-prefix was tried and measurably lost accuracy — tags are shared by dozens of emoji, so order then picks arbitrarily among them.

Conventional shortcodes unrelated to the Unicode label (":white_check_mark" for a "check mark button") stay unreachable; those need a real shortcode table.


2002Bishwajeet:
Closing as its superseeded with new prs

PR #1431, opened 2026-08-31
Show an HD indicator on photos sent at high quality

### LINKED ISSUE #1384: Show an HD indicator on sent/received photos (quality is not recorded on the wire)
## What

The composer lets you send a photo in HD (the `HD` FilterChip added in #1376), but once the message
is sent **nothing anywhere shows which quality it was sent at** — not the bubble, not the fullscreen
viewer, not any media-info surface.

`chat_media_quality_hd` has exactly one consumer today: the composer chip at
`MediaAttachmentEditor.kt:627`. Every other `MediaQuality` reference is send-side plumbing — the
toggle state in `ConversationListUiState`, and the encode decision in `MessageAttachmentBuilder` /
`MessageThumbnailGenerator` / `StandardQualityImage`.

## Why it matters

With STANDARD as the default, the overwhelming majority of photos are now downscaled to ~1600px.
That is the right default, but it makes the exception invisible: a sender who deliberately chose HD
gets no confirmation it took effect, and a recipient has no way to tell a full-quality photo from a
downscaled one before deciding whether to download or re-share it.

## The catch: this is a protocol change, not a UI change

**The quality is not recorded on the wire.** Nothing in `PayloadDescriptor` carries it, and it
cannot be inferred receiver-side. The two things that look like tells both collapse:

- **Payload dimensions.** STANDARD caps at 1600px — but a HIGH photo whose original was already
  ≤1600px is byte-for-byte the same shape. Small originals are indistinguishable.
- **Thumbnail-ladder length.** STANDARD ships `standardThumbSizes` (2 rungs) vs HIGH's 4 — but
  `getRevisedThumbs` already drops rungs at or above the source size, so a small HIGH image also
  arrives with 2 rungs.

So an indicator needs an explicit flag on the payload descriptor. Cheap in bytes, but it is a
wire-format addition with the usual consequences:

- Older senders will never set it, so **every existing message reads as "unknown"**.
- That forces a product decision: does absent mean *not HD*, or *don't render a badge at all*?
  Treating absent as "not HD" mislabels every HD photo sent before the flag ships.

## Open questions

- [ ] **Do we want this at all?** Prior art is split — WhatsApp badges HD photos in the bubble;
      Signal does not. #1369's final decision did not ask for one.
- [ ] **Where does it go?** The image bubble already overlays a timestamp + delivery-status scrim.
      Adding a second overlay element needs a Material 3 pass so it does not turn into clutter on
      small thumbnails, in grids, or on the fullscreen viewer.
- [ ] **Sender-side only, or both ends?** Showing it just to the sender ("your HD choice applied")
      needs no protocol change at all — the sending client knows. Showing it to the recipient is
      what costs a descriptor field. These are separable and the cheap half could ship first.
- [ ] **Does video need the same treatment?** The toggle also drives `VideoQuality`
      (STANDARD ≈720p / HIGH ≈1080p), and the same invisibility applies there.

## Not blocking anything

#1376 is complete and correct without this — it is an additive enhancement, deliberately kept out of
that PR to keep its diff reviewable as "encode smaller by default".

2002Bishwajeet:
Closes #1384.

The composer's HD chip changed the encode but left no trace on the wire, so once a photo
was sent neither end could tell a full-quality photo from a downscaled one. This records
the quality and badges it.

## The wire-format addition

`DescriptorContent.ImageFile` — the per-image descriptor that already rides on
`PayloadDescriptor.descriptorContent` and today carries `isSticker` / `format` — gains an
optional `quality: MediaQuality?`. No new protocol surface, no new payload, no header-budget
pressure: the object is `{"isSticker":false,"quality":"high"}`, ~34 bytes, and only on photos
that actually claim HD. `MediaQuality` becomes `@Serializable` with `@SerialName` values
matching its existing `code` (`"standard"` / `"high"`) so the wire spelling is the same one
the rest of the codebase already persists. `MessageAttachmentBuilder` already had
`mediaQuality` threaded in, so the send side is a three-line change.

**STANDARD writes nothing at all.** It keeps the legacy `""` descriptor, so an ordinary photo
is byte-identical on the wire to what it was before this PR — the same conservatism the
sticker feature used. The descriptor is only emitted when it carries information.

**Stickers never record quality.** A sticker skips `standardQualityImage` and is bounded at
512px by `StickerImageProcessor` regardless of the toggle, so "HD" would be a false claim.

## Absent means unknown, and renders nothing

Absent is **not** treated as "not HD" in the UI. Every photo sent before this ships reads as
absent, and many of those genuinely were HD — rendering them as standard would mislabel the
entire existing corpus. So absent draws nothing.

Standard also draws nothing: a badge on the overwhelmingly common case is noise. That makes
absent and standard render identically, which is the intended outcome — a *missing* badge on
an old HD photo is much cheaper than a *wrong* label on one.

## Where the badge renders, and where it deliberately does not

- **Solo photo bubble — yes.** Top-start, a bold `HD` on a `colorScheme.scrim` chip. Top-start
  rather than bottom: the media-only bubble's timestamp + delivery-status scrim is a 40dp
  bottom gradient drawn by the parent *over* this box, so a bottom badge would be dimmed by
  it, and top-start is also the one position that works identically for captioned bubbles
  (which have no scrim and put the timestamp below the media).
- **Fullscreen viewer — yes.** Appended to the app bar's timestamp line as `12:34 · HD`. This
  is the surface where "should I download or re-share this?" is actually decided, it has room,
  and it needs no overlay on the picture at all. It is per-payload, so it follows the pager.
- **Gallery / album cells — deliberately no.** The three-image layout's cells are ~104×69dp and
  the four-plus layout already overlays a "+N" scrim. Per-cell HD chips in a 4-up album would
  be clutter for a surface whose whole job is "tap to open" — and opening it lands on the
  fullscreen viewer, which does badge. This is the one place the badge is intentionally absent
  on media that qualifies for it.

The badge is text, not `Icons.Default.Hd`. I rendered both to PNG: the filled glyph is a solid
plate with HD knocked out of it, so white-on-scrim comes out as a white blob nested inside the
chip's own rounded rect; `Icons.Outlined.Hd` is two nested outlines. Bold `labelSmall` text is
one shape and legible at bubble-corner size. It also reuses `chat_media_quality_hd` — the
string the issue notes had exactly one consumer — so the marker on the photo is literally the
same word as the control the sender tapped.

## Video is out of scope

Video does **not** ride this descriptor and should not be bolted onto it:

1. Its descriptor is `VideoMetadata`, a different serialized type.
2. `MediaQuality.HIGH` for images is a byte-for-byte contract. For video it is only a
   *requested* encoder target — `VideoPayloadProcessor` can fall back to the input path
   entirely (`skipped=${compressedPath == payload.filePath}`), and a 480p source encoded at
   HIGH is still 480p. Recording the request would badge non-HD video as HD.
3. `VideoMetadata` already records `widthPx`/`heightPx` probed from the delivered output, so if
   a video badge is wanted the honest source is measured height, not the requested enum.

That is a different design and deserves its own issue rather than widening this diff.

## Tests

29 tests, JVM (`homebase-api`, `homebase-chat`), 0 skipped:

- `MessageAttachmentBuilderQualityTest` — a HIGH send records `high`; a STANDARD send records
  nothing and is not badged; a sticker never records quality even when sent as HIGH.
- `MediaQualityBadgeTest` — the badge predicate across HIGH / explicit standard / legacy blank /
  null / malformed / unknown-enum-value / link-preview JSON-array / video payloads.
- `HdBadgeRenderTest` — Compose render assertions: HIGH draws one badge, explicit standard and
  no-recorded-quality draw zero. This pins the absent→nothing decision at the UI layer, not
  just in the predicate.
- `PayloadDescriptorImageTest` — serialize/deserialize round trip, lowercase wire spelling,
  quality omitted when unrecorded, sticker+quality coexistence with an unknown sibling key, and
  an unknown quality value coercing to null rather than to a quality.

**Mutation-checked** (production logic deliberately broken, tests confirmed to fail, then
reverted):

| Mutation | Result |
|---|---|
| `isHighQualityImage()` treats absent as HD (`!= STANDARD`) | 9 tests fail |
| `MessageAttachmentBuilder` records quality for STANDARD too | `standardQualityPhoto_recordsNothing_andIsNotBadged` fails |
| Drop `@SerialName("high")` so the wire says `HIGH` | `quality_roundTripsThroughTheWire` fails |
| Absent-as-HD, checked against the *render* test specifically | `photoWithNoQualityRecorded_drawsNothing` fails |

Compiles on all four targets: JVM, iOS (`compileKotlinIosSimulatorArm64`), Android, wasmJs.

## Verified on an iOS simulator

Run for real on two simulators, one identity each: `frodo.baggins.demo.rocks` sending on an
iPhone 17 Pro Max, `samwise.gamgee.demo.rocks` receiving on an iPhone 17e.

- HD sent bubble, light and dark; badge top-start, legible over a bright sky, clear of the
  timestamp scrim.
- **HD received bubble on the second identity**, light and dark — the flag survives transit,
  which is the whole reason it is on the wire.
- Captioned HD bubble: no scrim, timestamp below the media, badge still top-start.
- Fullscreen viewer app bar reading `1m ago · HD`.
- Standard-quality photo and a pre-existing photo from **Aug 18** — both render no badge.
- 2×2 album of four HD photos — no badge on any cell.

The accessibility tree gives this machine-readably, since the badge sets a contentDescription:
a HD bubble reads `"Image attachment, Sent in high quality, …"`, a standard or pre-existing one
reads only `"Image attachment"`. The descriptor itself was observable on device as
`{"isSticker":false,"quality":"high"}`.

Device testing also turned up a real bug and fixed it: `FullScreenMediaViewer` used
`payload.descriptorContent` as the image's `contentDescription`, which was harmlessly empty
before this branch but now made VoiceOver read the descriptor JSON aloud. It uses the proper
string resource now.

## What was NOT verified

- **Android** — iOS simulator only.
- **Video**, which this PR deliberately leaves alone.

### DIFF
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/client/drives/files/PayloadDescriptor.kt
@@ -2,6 +2,7 @@ package id.homebase.api.client.drives.files
 
 import androidx.compose.runtime.Immutable
 import co.touchlab.kermit.Logger
+import id.homebase.api.image.MediaQuality
 import id.homebase.api.serialization.OdinSystemSerializer
 import id.homebase.api.video.VideoMetadata
 import kotlinx.serialization.Serializable
@@ -138,11 +139,16 @@ sealed interface DescriptorContent {
      * Additive and backward-compatible: a blank or legacy descriptor parses to
      * `ImageFile(isSticker = false, format = null)` and older receivers ignore [format]
      * (OdinSystemSerializer has ignoreUnknownKeys = true).
+     *
+     * [quality] is the send quality the sender chose. null means "not recorded" — every
+     * image sent before this field shipped, which is why a null must never render as
+     * "not HD": that would mislabel every HD photo sent before it.
      */
     @Serializable
     data class ImageFile(
         val isSticker: Boolean = false,
         val format: String? = null,
+        val quality: MediaQuality? = null,
     ) : DescriptorContent {
         /**
          * Filename extension for a downloaded sticker, derived from [format]. Defaults to
@@ -180,10 +186,15 @@ sealed interface DescriptorContent {
         /**
          * Serialize an [ImageFile] descriptor (e.g. `{"isSticker":true,"format":"image/png"}`)
          * for the wire. [format] is the detected image content type — carried so a receiver
-         * can name a downloaded sticker "StickerFile.<ext>".
+         * can name a downloaded sticker "StickerFile.<ext>". [quality] is the chosen send
+         * quality; leave it null to record nothing.
          */
-        fun descriptorContentFromImage(isSticker: Boolean, format: String? = null): String {
-            return OdinSystemSerializer.serialize(ImageFile(isSticker, format))
+        fun descriptorContentFromImage(
+            isSticker: Boolean,
+            format: String? = null,
+            quality: MediaQuality? = null,
+        ): String {
+            return OdinSystemSerializer.serialize(ImageFile(isSticker, format, quality))
         }
     }
 }
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/image/MediaQuality.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/image/MediaQuality.kt
@@ -1,13 +1,21 @@
 package id.homebase.api.image
 
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
 /**
  * How hard an outgoing photo or video is compressed before upload.
  *
  * [HIGH] is a byte-for-byte contract: the picked file's bytes reach the wire unmodified
  * (bar the EXIF scrub every image already gets).
  */
+@Serializable
 enum class MediaQuality(val code: String) {
+    // Wire names match `code` so a descriptor written by any client reads the same.
+    @SerialName("standard")
     STANDARD("standard"),
+
+    @SerialName("high")
     HIGH("high");
 
     companion object {
--- a/homebase-api/src/jvmTest/kotlin/id/homebase/api/client/drives/files/PayloadDescriptorImageTest.kt
+++ b/homebase-api/src/jvmTest/kotlin/id/homebase/api/client/drives/files/PayloadDescriptorImageTest.kt
@@ -1,5 +1,6 @@
 package id.homebase.api.client.drives.files
 
+import id.homebase.api.image.MediaQuality
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
@@ -80,6 +81,62 @@ class PayloadDescriptorImageTest {
         val padded = imageDescriptor("""   [{"url":"https://example.com"}]""").descriptorInfo()
         assertTrue(padded is DescriptorContent.ImageFile, "Expected ImageFile, got $padded")
         assertFalse(padded.isSticker)
+    }
+
+    @Test
+    fun quality_roundTripsThroughTheWire() {
+        val serialized = DescriptorContent.descriptorContentFromImage(
+            isSticker = false,
+            quality = MediaQuality.HIGH,
+        )
+        assertTrue(
+            serialized.contains("\"quality\":\"high\""),
+            "Quality must serialize with its lowercase code, got $serialized",
+        )
+        val info = imageDescriptor(serialized).descriptorInfo()
+        assertTrue(info is DescriptorContent.ImageFile, "Expected ImageFile, got $info")
+        assertEquals(MediaQuality.HIGH, info.quality)
+
+        val standard = DescriptorContent.descriptorContentFromImage(
+            isSticker = false,
+            quality = MediaQuality.STANDARD,
+        )
+        val standardInfo = imageDescriptor(standard).descriptorInfo()
+        assertTrue(standardInfo is DescriptorContent.ImageFile)
+        assertEquals(MediaQuality.STANDARD, standardInfo.quality)
+    }
+
+    @Test
+    fun qualityIsOmittedWhenNotRecorded() {
+        // Absent must stay absent — it is what every pre-flag image reads as, and it must
+        // never be coerced to STANDARD, which would mislabel old HD photos.
+        val serialized = DescriptorContent.descriptorContentFromImage(isSticker = true)
+        assertFalse(serialized.contains("quality"), "Unexpected quality key in $serialized")
+
+        val info = imageDescriptor(serialized).descriptorInfo()
+        assertTrue(info is DescriptorContent.ImageFile)
+        assertNull(info.quality)
+
+        assertNull((imageDescriptor("").descriptorInfo() as DescriptorContent.ImageFile).quality)
+    }
+
+    @Test
+    fun stickerAndQualityCoexist_andQualitySurvivesAnUnknownSibling() {
+        val info = imageDescriptor(
+            """{"isSticker":true,"format":"image/png","quality":"high","futureField":42}"""
+        ).descriptorInfo()
+        assertTrue(info is DescriptorContent.ImageFile)
+        assertTrue(info.isSticker)
+        assertEquals(MediaQuality.HIGH, info.quality)
+    }
+
+    @Test
+    fun unknownQualityValue_coercesToNull_notToAQuality() {
+        // coerceInputValues turns an unrecognised enum into the default (null) rather than
+        // throwing, so a newer sender's vocabulary degrades to "not recorded".
+        val info = imageDescriptor("""{"isSticker":false,"quality":"ultra"}""").descriptorInfo()
+        assertTrue(info is DescriptorContent.ImageFile, "Expected ImageFile, got $info")
+        assertNull(info.quality)
     }
 
     @Test
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/builder/MessageAttachmentBuilder.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/builder/MessageAttachmentBuilder.kt
@@ -70,11 +70,20 @@ object MessageAttachmentBuilder {
                         // as a normal image (issue #854). When it IS a sticker we detect the
                         // real format so the download is named "StickerFile.<ext>"; ordinary
                         // photos keep the legacy "" so nothing changes for them.
+                        //
+                        // Quality is recorded only for a HIGH ordinary photo: STANDARD keeps
+                        // the legacy "" (absent reads as standard), and a sticker's bytes are
+                        // bounded at 512px regardless of the toggle, so "HD" would be a lie.
                         val descriptorContent =
                             if (attachment.forceSticker) {
                                 DescriptorContent.descriptorContentFromImage(
                                     isSticker = true,
                                     format = ImageFormatDetector.detectFormat(thumbs.sourceBytes),
+                                )
+                            } else if (mediaQuality == MediaQuality.HIGH) {
+                                DescriptorContent.descriptorContentFromImage(
+                                    isSticker = false,
+                                    quality = MediaQuality.HIGH,
                                 )
                             } else {
                                 ""
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/FullScreenMediaViewer.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/widget/FullScreenMediaViewer.kt
@@ -68,6 +68,7 @@ import id.homebase.core.util.formatTimestamp
 import id.homebase.resources.MR
 import id.homebase.resources.chat_image_unavailable
 import id.homebase.resources.chat_media_quality_hd
+import id.homebase.resources.chat_message_image_attachment
 import id.homebase.resources.chat_options
 import id.homebase.resources.menu_back
 import id.homebase.resources.share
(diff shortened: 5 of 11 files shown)


2002Bishwajeet:
## Verification status — on-simulator run is blocked, and here is exactly why

I was asked to verify this on the iOS simulator. **I could not**, and I want to be precise
about the reason rather than hand back a softer claim.

There is no logged-in Homebase account on any simulator on this machine:

- The app is installed on exactly one simulator — **iPhone 17e** (`E6F1C627-…`), bundle
  `id.homebase.feed.dev`. Nothing on the other seven.
- Its container has only `Library/Preferences` and `Documents/logs`. **There is no
  `odin-2.db`** — I searched every simulator device container, no database anywhere.
- Its log covers today, 11:31 → 11:49, and every one of the 10 `AuthState` lines reads
  `AuthState: Unauthenticated`. It has never been signed in.
- `adb devices` is empty, so there is no Android fallback either.

Reaching a chat bubble requires an identity, and I have no credentials. So shots of a real
sent/received bubble, a captioned bubble, the fullscreen app bar, and a gallery cell could
not be taken. I did not send anything to anyone.

**I have built this branch and am installing it on iPhone 17e** so it can be signed into and
checked directly — that is the part only someone with the credentials can do.

## What I did verify, and one claim I can now state more strongly

### Badge vs. the bottom timestamp scrim — this is geometry, not a screenshot

The PR body claimed top-start "keeps it clear of the bottom timestamp scrim". That is
provable for every bubble size rather than for one screenshot:

- `MediaTimestampOverlay` is `height(40.dp)`, `align(BottomStart)` — it occupies the bottom
  40dp only.
- `HdBadge` is at top-start: 6dp inset + ~18dp tall (labelSmall + 2dp vertical padding),
  so it ends around y = 24dp.
- Every solo media bubble is at least `Dimens.MediaBubble.minHeight = 100.dp`
  (`fillsBubble` → 320dp, `narrowCaptioned` → `coerceIn(minHeight, maxHeight)`, otherwise
  `heightIn(min = minHeight)`).

Worst case 100dp tall: badge ends at 24dp, scrim starts at 60dp — **a 36dp gap**. They
cannot overlap at any size, so the badge is never dimmed by the scrim. The captioned path
has no scrim at all, so it is clear there by construction too.

### Legibility over a real photo, including the worst case

Rendered the real `HdBadge` at true 240×320dp bubble scale over a real photograph with a
bright sky in the top-left corner, and over pure white, with a replica of the scrim (same
40dp, same transparent → black-0.6 gradient, same BottomEnd 12dp inset) so both can be seen
together. The badge reads cleanly in both, and sits far clear of the timestamp.

The contrast is computable, so it does not depend on which photo I happened to pick. The
chip is `scrim` at 55% over whatever pixel is behind it, with white bold text:

| Photo content behind the badge | Chip luminance | Contrast vs. text |
|---|---|---|
| pure white (absolute worst case) | 114.8 | **4.76:1** |
| bright sky (230) | 103.5 | 5.61:1 |
| mid grey (128) | 57.6 | 11.44:1 |
| black | 0 | 21.00:1 |

Even against a pure-white photo the badge clears WCAG AA for normal text (4.5:1), and it is
bold, which only needs 3:1. There is no photo content that makes it illegible.

Nothing here contradicted the PR body. The scrim claim is now stronger than it was written —
it holds by construction rather than by inspection.

## Still unverified, and needs a signed-in run

Everything that requires real messages:

1. Sent and received HD bubbles, light and dark.
2. A captioned HD bubble.
3. The fullscreen viewer's `12:34 · HD` app-bar line.
4. A standard-quality photo and a pre-existing (absent-flag) photo showing no badge —
   asserted by `HdBadgeRenderTest` but not seen on a device.
5. A gallery cell with a qualifying photo showing the badge deliberately absent.

The render PNGs are on this machine if they are useful; say the word and I will attach them
somewhere referenceable rather than leaving them as local files.

2002Bishwajeet:
## Verified on the iOS simulator — full set, including a genuine received bubble

Retract my earlier comment: I had said there was no signed-in account. There was no *database
with data*, but with the demo credentials I was able to sign in and run the whole thing for
real. Everything below is from a device, not a render.

- **Sender:** `frodo.baggins.demo.rocks` on **iPhone 17 Pro Max** (`4DD5589F-…`)
- **Receiver:** `samwise.gamgee.demo.rocks` on **iPhone 17e** (`E6F1C627-…`) — a second
  simulator, so frodo was never logged out
- Build: this branch, `iosApp Dev` scheme, installed on both

### The screenshots

Local paths (I can't attach binaries to a PR comment from the CLI; say the word and I'll put
them somewhere referenceable):

`/private/tmp/claude-501/-Users-biswa-Documents-GitHub-chat-kmp/4fd80a9a-202a-4127-a5ce-e453d5f9b740/scratchpad/`

| File | What it shows |
|---|---|
| `shot1-hd-sent-light.png` | HD photo, **sent** bubble, light. Badge top-start over a bright sky; timestamp scrim bottom-right, no interaction |
| `shot2-hd-received-light.png` | HD photo, **received** bubble on Samwise's device, light. Badge present |
| `shot2-hd-received-dark.png` | Same, dark |
| `shot2-hd-sent-dark-album-dark.png` | Sent HD bubble + album, dark |
| `shot3-hd-captioned-light.png` | **Captioned** HD bubble — no scrim, timestamp below the media, badge still top-start |
| `shot4-fullscreen-hd.png` | Fullscreen viewer app bar reading `1m ago · HD` |
| `shot5-hd-vs-standard-light.png` | HD badged directly above a standard photo with **no badge** |
| `shot5b-preexisting-no-badge.png` | A photo from **Aug 18**, long before this build — **no badge** |
| `shot6-album-no-badge-light.png` | 2×2 album of four HD photos, **no badge on any cell**, below a solo HD bubble that has one |
| `band-verdict-clean-hd-vs-banded-standard.png` | The black-band control (see below) |

### Machine-readable proof, straight from the iOS accessibility tree

This is stronger than the pixels. The badge sets `contentDescription = "Sent in high quality"`,
so the bubble's own label states whether it is badged:

```
HD photo:          AXButton "Image attachment, Sent in high quality, 1m ago, Sent"
                   AXGroup  "Sent in high quality"
standard photo:    AXButton "Image attachment"
pre-existing (Aug 18): AXButton "Image attachment"
captioned HD:      AXButton "Image attachment, Sent in high quality, Lake Tahoe, captioned, Just now, Sent"
delivered to peer: AXButton "Image attachment, Sent in high quality, Just now, Delivered"
fullscreen app bar: AXStaticText "1m ago · HD"
```

The wire format is also directly observable on device — the fullscreen viewer's image element
carried the literal descriptor:

```
{"isSticker":false,"quality":"high"}
```

34 bytes, exactly as the PR body claims.

### The received bubble is the one that matters

Frodo sent, Samwise received on a different device after a fresh login, and the badge renders
on the received bubble. That is the whole justification for putting the flag on the wire rather
than keeping it sender-local, and it now has direct evidence.

The badge also survived two app restarts, so it is being read from the persisted descriptor,
not from in-memory send state.

## The black band: not a regression, and not an app bug — it was my test photo

**Verdict: my fault, in the fixture, not in the code.** Evidence, in order of decisiveness:

**1. The band is baked into the JPEG.** I generated the test photo by cropping a 3840×2160
wallpaper with a crop box ending at y=2412 — past the bottom edge. PIL pads out-of-bounds
regions with black. Sampling the file directly:

```
row at 80%: mean RGB 102.3 151.4 140.0
row at 85%: mean RGB 104.0 148.5 133.7
row at 90%: mean RGB   0.0   0.0   0.0
row at 95%: mean RGB   0.0   0.0   0.0
```

Pure black from ~88% down, in the source file, before it ever reaches the app.

**2. Controlled experiment on device.** I sent the *same* photo at **standard** quality. The
band is identical and there is no badge. Then I sent a correctly-cropped photo at **HD**: badge
present, **no band**. Both are in `band-verdict-clean-hd-vs-banded-standard.png`. So the band
tracks the *photo* and the badge tracks the *quality* — the two are independent, and the
apparent correlation was me happening to use the banded photo for the HD sends and a stock
photo for the standard one.

I did not rebuild `main` for this. The same-photo/both-qualities control is a stricter test than
a cross-branch comparison — it holds the image constant and varies only the code path this PR
touches — and the pixel dump settles it outright. If you'd still like the `main` build as
belt-and-braces, say so and I'll run it.

## One real bug this found, now fixed

`FullScreenMediaViewer` passed `payload.descriptorContent` as the image's `contentDescription`
in three places. That was harmlessly empty for ordinary photos before this branch — but once a
HD photo carries a descriptor, VoiceOver reads `{"isSticker":false,"quality":"high"}` out loud.
I only saw it because the accessibility tree showed the raw JSON as the image's label.

Fixed in `b6b7a855` to use the same string the bubble already uses. Tests and iOS/JVM compiles
still green.

## Also worth knowing: an unrelated Compose crash

The app hit `EXC_BAD_ACCESS` twice during screen transitions while I was driving it. Both
crashes are identical and contain **no frames from this branch**:

```
kfun:androidx.compose.ui.platform.AccessibilityElement.<get-node>#internal
kfun:androidx.compose.ui.platform.AccessibilityElement.contentOffset#internal
kfun:androidx.compose.ui.platform.AccessibilityElement.bounds#internal
-[CMPAccessibilityElement accessibilityFrame]
UIAccessibility ... _handleNonMainThreadCallback
```

A null-deref in Compose Multiplatform's iOS accessibility layer, reading a disposed semantics
node when the accessibility runtime asks for element frames off the main thread. It is provoked
by automated accessibility inspection (and would presumably affect VoiceOver users); a normal
user without assistive tech would not hit it. Both sends completed successfully regardless.
Pre-existing and out of scope here, but probably worth its own issue.

## Not verified

- Android. iOS simulator only.
- Video, which this PR deliberately does not touch.

Issue #1054, opened 2026-07-10
iOS (HIGH): "Create account" does nothing — create-account path calls the iOS no-op launchAuthBrowser; sign-up URL never opens (blocks onboarding)


Seifert69:
**Priority: HIGH** — blocks new-user onboarding on iOS (a prospective user can't create an
account at all). Filing now; the reporter's exact iPhone model will be added later.

## Summary

On **iOS**, tapping **"Create account" ("Opret konto")** on the login screen **does nothing** —
no browser/sign-up page opens. Confirmed root cause: the create-account path calls the iOS
`launchAuthBrowser`, which is a **no-op** on iOS. The reporter also notes the **keyboard keeps
popping up** on this screen (secondary; see below).

## Environment

- Platform: **iOS / iPhone** (confirmed from the screen recording — iOS keyboard with `.dk` +
  globe, dynamic island). Exact model **TBD** (reporter sending shortly).
- Screen: initial **Login** screen (`LoginScreen`), Danish locale.

## Evidence

![Login screen — "Opret konto" (Create account) below "Prøv igen"; keyboard up](https://raw.githubusercontent.com/homebase-id/chat-kmp/issue-assets/docs/issue-assets/ios-create-account-noop/login-screen.png)

Reporter (paraphrased): *"nothing happens when I press the create new account link, and the
keyboard keeps popping up."* A screen recording exists and can be attached.

## Root cause (confirmed in code)

On iOS, `rememberAuthBrowserLauncher()` returns a **no-op that returns `true`**:

`homebase-common/src/nativeMain/kotlin/id/homebase/core/ui/auth/AuthBrowserLauncher.native.kt`
```kotlin
@Composable
actual fun rememberAuthBrowserLauncher(): (url: String) -> Boolean =
    remember { { _: String -> /* No-op */ true } }   // iOS launches via BrowserLauncher.onAuthBrowserOpened instead
```

It's intentionally a stub because the iOS **login** flow launches `ASWebAuthenticationSession`
separately via `BrowserLauncher.onAuthBrowserOpened(...)`. But the **create-account** flow
never makes that second call:

- **Create account (broken):** `LoginUiAction.CreateAccount` → `uiEvent =
  OpenUrl("https://homebase.id/sign-up")` (`LoginViewModel.kt:73-75`) → `LoginScreen.kt:137-139`
  `OpenUrl -> { eventConsumed(); launchAuthBrowser(uiEvent.url) }`. On iOS
[...]

PR #1473, opened 2026-09-03
Redesign the login screen against the Homebase design manual v2

2002Bishwajeet:
Top layer of a two-PR stack — **base is #1472**, review that one first. This PR is `homebase-auth`, the strings the screen needs, and one line in `homebase-core`'s `AppNavHost`.

The old screen was a bounded card floating on a flat surface, with the identity field, the CTA and the brand mark all competing at the same weight. It carried none of the manual's marks: no constellation, no Northern Lights palette, no Montserrat, and the wordmark rendered positive on a dark ground, against p9.

## Layout

The two-pane split needs a landscape-ish frame, so it's gated on `isExpandedLayout()` **and** `width > height`. `isExpandedLayout()` is the shared 840dp gate the navigation rail and the list/detail scaffolds use and stays shared — but it tests width alone, which put a 13" iPad in portrait into a two-pane whose seam ran down a 1366dp screen with a void either side. The orientation test is local to this screen; a second global breakpoint would only make the app disagree with itself.

Landscape is a 45/55 split with the brand pane full-bleed. Dropping the bounded panel also deleted the figure/ground problem it was creating: the old card's ground sat *between* the two halves' values in both themes, and there was no mapped surface role outside them to fix it with.

**Portrait is one path at every width — phone and tablet alike — and carries no brand artwork at all.** The logo, the headline and the form on the plain surface, form width-capped, metrics scaled off the pane. A coloured band above the form read as decoration the screen hadn't earned, and splitting portrait into a phone strip and a tablet hero meant maintaining two compositions to say the same thing.

The portrait logo is the **bare glyph filled with the brand purple→blue gradient**. The manual's mark is the *negative* version — a white glyph drawn for the blue ground — so on a plain surface it has to be recoloured, and neither brand blue nor the M3 `primary` lavender is a designed positive version. The full-colour app-icon tile was tried and dropped too: a rounded square exists to survive a home-screen grid, not a login screen.

`Icon` takes a flat `tint` only, so the fill is the glyph's alpha used as a mask — an offscreen `graphicsLayer` plus `drawRect(brush, BlendMode.SrcIn)`. Without `CompositingStrategy.Offscreen`, `SrcIn` composites against the window and paints a solid block; the 6× crops in review show the glyph's counters painted in surface colour, so the mask is real.

**The ramp is lifted for dark.** `HomebaseBrand.Purple`/`Blue` measure **2.10:1** and **2.43:1** against the dark surface — dimmer than the flat blue this replaced. `PurpleOnDark` `#9473CA` and `BlueOnDark` `#748EBD` are the same two hues raised in OkLCh lightness with hue and chroma held, at **4.51:1** and **5.15:1**, at 100% of original chroma (no pastel trade needed). Light uses the brand values unchanged (7.91:1 / 6.82:1).

The lift is **one uniform ΔL applied to both endpoints**, not each pushed to its own target. Blue is fractionally lighter than Purple in the base pair (L 0.4654 vs 0.4437); per-endpoint lifting would be +0.180/+0.146 and inverts that ordering into a different gradient.

64dp, matching `BrandMarkSize` in the landscape pane, so both layouts show the mark at the same size. Landscape keeps the flat negative mark, correct on its blue pane — proved pixel-identical by rendering 1440×900 from `git show HEAD:` sources and diffing (`bbox=None`).

## Insets

`LoginScrollColumn` takes `windowInsetsPadding(WindowInsets.safeDrawing)` **outside** both its content padding and its scroll, so the safe area is a boundary the viewport can't cross.

Without it the brand mark rendered under the status bar on a phone. The field autofocuses on arrival (pre-existing, #1054), so the keyboard is up the moment the screen appears; that shrank the viewport below the content height, and a centred column with no slack top-aligns against the raw window edge. Measured on an iPhone 17e: the mark moved from **y=34pt to y=84pt** against a 59pt inset.

`safeDrawing` rather than `statusBarsPadding()` / `safeDrawingPadding()` because those publish consumption into a legacy modifier-local that `windowInsetsPadding` and the M3 app bars can't read on skiko. Its per-side union also counts a raised keyboard once instead of stacking the ime on the navigation bar, which had been double-padding the bottom on iOS.

The landscape form pane had the same latent bug — its top element sat 95pt below a 24pt inset on an iPad, which was luck, not a rule. Same fix covers it.

## Identity

`IdentityPreviewResolver` debounces the typed homebase-id and swaps the mark for the public profile avatar once it resolves. `LastIdentityCard` offers the previous identity as a one-tap pill and replaces the field entirely while it's showing.

Pill buttons are scoped to this screen only, per your call.

## Desktop

The login route opts out of the `AppNavHost` top inset so the brand pane runs under the transparent macOS title bar. Safe only because its top-left is empty ground — the traffic lights have nothing to collide with. `safeDrawing` contributes 0dp on desktop (nothing populates `LocalPlatformWindowInsets` there), so there's no double inset.

## Tests

`LoginUiTest` goes 7 → 17. The headline pin is a geometry canary and was mutation-tested rather than just asserted: dropping the state-slot reserve to 196dp fails it.

The star-clearance canary went with the band fade it guarded — a fade and the asterism no longer coexist in any layout, so the code it defended is gone rather than merely untested.

Verified on the iPhone 17e and iPad Air 13" simulators, portrait and landscape, keyboard up and down, both themes.

## Known follow-ups, not fixed here

- At 2560 the field's floating label doesn't shrink on float — 148px → 148px, against 108 → 88 at 1280. One-line fix via `labelTextStyle`. Only surfaces in the resolved/error/loading states.
- The headline subtitle stays `bodyLarge` against a flat 24dp reserve, so it reads small under a scaled tablet headline. Scaling it needs the reserve scaled with it or the headline pin breaks.
- **Separate CMP-iOS bug, not from this PR:** focus the field, rotate to landscape, rotate back — the whole scene returns translated ~295pt up, clipped and unscrollable, until the keyboard is dismissed. It's a translation applied above the Compose content root, so no in-tree padding can defend against it. Worth its own issue.

https://claude.ai/code/session_01D6McDnpKJjiPQN6b4w5ofx


2002Bishwajeet:
Superseded by #1474 — GitHub auto-closed this when the #1472 squash-merge deleted its base branch `login-brand-primitives`. Same commit, rebased onto `main`.

PR #1481, opened 2026-09-07
Show an iOS app-icon badge for undismissed notifications (#1347)

### LINKED ISSUE #1347: iOS: show an app-icon badge/number for undismissed notifications (currently blank, while Android shows a dot) — approach is Bishwa's call
## Goal

**The iOS app icon should show a badge/number when there is a notification the user hasn't pressed or dismissed.** Today it shows **nothing** in that case, while Android already shows a dot — so iOS users get no icon indicator that they have missed messages. Fix that.

**Bishwa has free hands on the implementation** — the notes below are context to make an informed choice, not a prescribed design.

## Current state (context, grounded)

- **iOS shows nothing for a killed-app message.** iOS has no "dot" concept — only a numeric red badge or blank. And right now nothing sets it in the common case:
  - The NSE (`iosApp/NotificationServiceExtension/NotificationService.swift`) runs but **does not set `content.badge`**.
  - The push payload has **no `aps.badge`** (`Odin.PushNotification` builds `Aps` with `ContentAvailable`/`MutableContent`/`Alert`, no `Badge`).
  - `BadgeManager.increment()` (Kotlin) only runs while the main app is alive, so it can't fire for a killed app.
- **Android already indicates it** — the launcher dot is automatic from posting a notification; `RichNotificationDisplayer.android.kt:70` also sets `setNumber(...)` for launchers that show a count.
- Useful building blocks that already exist: `BadgeManager` (per-platform), a **"Badge Count" settings section** (`NotificationSettingsScreen.kt:257`, incl. an *include-muted-chats* toggle), and the true unread total via `ChatReadCountWrapper.selectAllUnreadCount`.
- The iOS NSE already runs on every push (`MutableContent = true`), so setting `content.badge` on-device is readily available if wanted.

## Requirement (the must-have)

- iOS app icon shows a **badge/number** whenever there are **undismissed / unread** notifications the user hasn't opened — including when the message arrived while the app was **killed/backgrounded**.
- It **clears** when the user reads/opens/dismisses (down to 0 when nothing is pending).
- Exact value (a running count vs. true unread from `selectAllUnreadCount`) and mechanism are **Bishwa's call**.

## One note worth weighing (not a mandate)

If the count is put in the **push payload** (server `aps.badge`), it passes through the **central `Odin.PushNotification` relay + Firebase (Google)** — a small extra metadata leak (the payload already carries sender/conversation/message ids, so it's incremental, not new-in-kind). Setting the badge **on-device in the NSE** (from a shared App-Group value the app maintains) avoids putting the count in the payload, which fits Homebase's data-minimization goals. Bishwa can pick whichever he prefers given effort vs. that trade-off.

## Acceptance criteria

- [ ] With an undismissed/unread notification (incl. arrived while app killed), the iOS app icon shows a badge/number — not blank.
- [ ] The badge clears/decrements as the user reads/opens/dismisses, reaching 0 when nothing is pending.
- [ ] Behavior is sensible on cold-launch (badge reflects reality after the app has synced).
- [ ] Android indicator continues to work (no regression).
- [ ] Honors the existing include-muted-chats setting where practical.

2002Bishwajeet:
Closes #1347.

## What

The iOS app icon showed **nothing** when a notification arrived that the user hadn't opened — including the common case of a message landing while the app was killed. Android already shows a launcher dot. The NSE ran on every push but never set `content.badge`, the payload carried no `aps.badge`, and `BadgeManager.increment()` only runs while the main app is alive.

The count is now maintained **on-device**. No `aps.badge`, no server or `Odin.PushNotification` change: the unread count never enters the push payload, so it doesn't transit the central relay or Firebase.

- The NSE reads/increments an integer in the shared App Group `UserDefaults` and sets `content.badge`, applied immediately after the `silent` guard so all four terminal `contentHandler` paths carry it. Silent pushes and the payload-parse bail don't increment.
- The app stays the source of truth: `BadgeManager.setCount(n)` writes both the icon badge and the shared integer, so the NSE's next increment starts from truth rather than drifting. `resetCount()`/`cancelAll()` clear the shared value too, so logout can't leave the extension resuming from a stale number.
- `AppViewModel.onResumed()` re-asserts the true unread total instead of calling `resetCount()`. Blind-zeroing was wrong under the new behaviour — with messages still unread the icon must keep showing them.

The group id is read from `Info.plist` (`AppGroupIdentifier`), never hardcoded: it is `group.id.homebase.feed.dev` on Debug/Dev and `group.id.homebase.feed` on Release, so a literal would have silently no-op'd every debug build. The NSE's `Info.plist` was the one target missing that key; it's added. **No pbxproj or Xcode-UI change is needed** — the setting reaches the NSE through the project-level `baseConfigurationReference` to `Config.xcconfig`/`Dev.xcconfig`. Verified against the built `.appex` in both configurations, where `AppGroupIdentifier` matches that configuration's entitlements file.

## Two changes beyond "set a badge" that reviewers should look at

**1. iOS no longer calls `BadgeManager.increment()` from `NotificationService`.** The NSE runs for every mutable-content push whether or not the app process is alive. With the NSE counting, a backgrounded-but-alive app counting too would double-count every message. Android and Desktop keep their `increment()`.

**2. Android's `badgeCount` changes slightly.** With `resetCount()` on resume replaced by `setCount(trueTotal)`, the value feeding `RichNotificationDisplayer.android.kt:70`'s `setNumber(...)` becomes the real unread total at resume instead of 0. This is required rather than incidental: if `setCount` were a no-op on Android, dropping the resume reset would let `badgeCount` grow unbounded. The launcher dot is unaffected — it comes from having any posted notification.

`totalUnreadCount` is gated on `enrichment.hasUnreadCounts`: before that pass every row reads 0, and emitting it would blank the badge on every cold start.

## Known limitations

- **`includeMutedChatsInBadge` is not honoured, because there is no per-conversation mute in this codebase.** A repo-wide search finds only video mute and `Mutex`; `ConversationUiModel` has no mute field and `selectAllUnreadCount` has no mute filter. The preference has a settings toggle and zero consumers today. Honouring it means first building conversation muting — out of scope here, and it was not faked.
- **The NSE increment races.** Two pushes arriving together can run concurrent extension processes, and cross-process `UserDefaults` read-modify-write is not atomic, so a burst can undercount. Survivable only because the app rewrites the absolute total on foreground/read; there's a one-line note naming that ceiling. No try/catch, no sleep.
- **Cold launch shows no badge for one beat** until the unread-enrichment pass lands, then shows the true remaining unread.
- `resetCount()` now has zero callers. Kept (its iOS actual clears the shared value) but flagged as dead API.

## Verification

Real `BUILD SUCCESSFUL` for: `:homebase-common:compileKotlinIosSimulatorArm64`, `:compileKotlinJvm`, `:compileAndroidMain`, `:compileKotlinWasmJs`; `homebase-common:jvmTest`, `homebase-chat:jvmTest`, `homebase-core:jvmTest`; `:homebase-core:compileKotlinIosSimulatorArm64` and `linkDebugFrameworkIosSimulatorArm64`.

Swift: `swiftc -parse` clean, and `xcodebuild` produced `NotificationService.o` (arm64 + x86_64) and a `NotificationServiceExtension.appex` in both Debug and Release.

**Not verified: a badge was never observed on an actual icon.** That needs a signed build on a real device plus a real APNs push through the extension. Reviewers should treat the end-to-end behaviour as unproven until someone runs it on hardware.

The full `iosApp` scheme build fails from the CLI at the "Compile Kotlin Framework" script phase (`:homebase-core:syncComposeResourcesForIos > Cannot query the value of this provider because it has no value available`). That's the Compose Gradle plugin under CLI-driven builds and is unrelated to this change — the Kotlin compiles and links, per `linkDebugFrameworkIosSimulatorArm64`. Building from Xcode is untested.


2002Bishwajeet:
Reworked away from the stored counter, per the discussion on approach.

**The badge is now the count of notifications still in the tray**, computed in the NSE via `getDeliveredNotifications { content.badge = $0.count + 1 }`. The App Group integer, its cross-process read-modify-write race, and all the persistence are gone — `AppGroupIdentifier` is removed from the NSE's `Info.plist` again, and no `UserDefaults` read remains in the extension.

Why the counter was the wrong shape:

- **Dismissal never decremented it.** The acceptance criterion is that the badge clears as the user reads/opens/**dismisses**; a stored counter has no idea the tray was emptied.
- **NSE invocation isn't guaranteed** — iOS throttles and skips extensions under memory/budget pressure, and a 30s timeout hands off to `serviceExtensionTimeWillExpire`. Every skipped run was permanent drift. Reading the tray is self-correcting: a skipped run costs one push, not the running total.

**This needed its missing other half.** `cancelConversationNotifications` was a no-op on iOS ("no Kotlin-side handle to cancel an individual conversation's notification"), so notifications lingered in the tray after the message was read on any device — a bug in its own right, and one that would make a tray-derived badge wrong. It's now implemented via `removeDeliveredNotifications(withIdentifiers:)`, matching on the `threadIdentifier` the NSE already sets. That's the same string equality the Android path already depends on, so no new assumption.

The `expect` was **narrowed rather than widened**: `cancelConversationNotifications(messageId: Int, summaryId: Int)` → `(conversationId: String)`. Both ints were always derived from the id, so that derivation moved into the Android actual instead of shipping Android notification ints to iOS.

Sequencing in the NSE is **nested, not raced** — the tray read is a local IPC call returning in milliseconds against a 30s budget, so serialising costs nothing, and nesting guarantees `content.badge` is set before any terminal `contentHandler` can run. A 2s fallback fires the continuation unbadged if the tray read never calls back, so a stuck read can't strand the push. `bestAttemptContent` is the same object as `content`, so `serviceExtensionTimeWillExpire` carries the badge automatically.

`cancelAll()` now clears the tray on logout too — a tray-derived badge is incoherent if the icon zeroes but the previous identity's notifications stay for the next push to recount.

**Worth a reviewer's opinion:** two writers now encode different definitions — the NSE writes *tray count*, `AppViewModel` writes *unread conversation total*. Both are absolute `setCount`s, last-writer-wins. They visibly disagree only in a narrow, transient, foreground case: the app clears a conversation's tray entries while its unread count doesn't change (already read elsewhere). Happy to drop the re-assert and let `AppViewModel` own the badge whenever the app is alive, if you'd prefer one definition.

Verification: `:homebase-common:compileKotlinIosSimulatorArm64` + jvm/android/wasm compiles, `homebase-common:jvmTest` / `homebase-chat:jvmTest` / `homebase-core:jvmTest` all green after merging main. Swift `xcodebuild` of the NSE target succeeds in Debug and Release, and `plutil -p` on both built `Info.plist`s confirms `AppGroupIdentifier` is gone.

**Still not verified: no badge has been observed on an actual icon.** That needs a signed build on hardware plus a real APNs push. Per-conversation removal is likewise unproven at runtime.

PR #1482, opened 2026-09-07
Media settings screen with opt-in auto-save of incoming chat media (#1154)

### LINKED ISSUE #1154: Auto-save incoming chat photos to the device photo album (opt-in)
## What

Add an opt-in setting that automatically saves **incoming** chat images to the phone's own photo album (Android gallery / iOS Photos), the way WhatsApp's "Save to camera roll" does.

Today the only way to get a received image into the camera roll is manual: open it full screen → ⋯ → Save (`FileSystemHandler.saveFile`, wired at `ChatMediaFullScreenHost.kt:132`/`:207`). That's one image at a time.

## Why it's cheap

The platform save path already exists and is used in production:

- `FileSystemHandler.saveFile(file, suggestedName, onSuccess, onError)` — `homebase-common/.../util/FileUtilities.kt:31`
- Android: MediaStore `Images`/`Video` insert, no runtime permission on API 29+ — `FileUtilities.android.kt`
- iOS: `PHPhotoLibrary` **add-only** authorization + `PHAssetCreationRequest` — `FileUtilities.native.kt:123`

So the feature is a setting + a trigger, not new platform plumbing.

## Proposed scope (deliberately small)

1. One preference in `UserPreferences` — `autoSaveIncomingMedia`, **default off**.
2. One Settings toggle: *"Save incoming photos to my photo album"*.
3. On an incoming image message whose payload is available locally, call the existing `saveFile`.
4. Skip anything already saved (dedupe by message uniqueId in prefs) so a re-sync doesn't duplicate the album entry.

Images only for v1. Video is the same call but a far bigger bandwidth/storage bill — separate decision.

## Open questions

**A. Download trigger.** Chat media is fetched lazily (thumbnail in the bubble, full payload on open/prefetch — `MediaItem.kt:281`, `FullScreenImageRequest.kt`). "Auto-save on receive" therefore implies **auto-download on receive**, which is a real cost on mobile data. Two options:

- *Lazy:* auto-save only when the full payload already got downloaded (user opened it, or prefetch pulled it). Zero new network. Weaker promise — an unopened photo never lands in the album.
- *Eager:* download every incoming image in the background so the album is complete. Matches WhatsApp's behaviour, but needs at least a Wi-Fi-only guard.

Recommend deciding this explicitly before implementing — it's the whole substance of the feature.

**B. Per-contact / per-group opt-out** (raised in the request). Recommendation: **don't build it yet.** WhatsApp's media-autosave settings are a three-state matrix × three network types × per-chat override, and they're widely considered unintelligible. Ship the single global toggle, see whether anyone actually asks for exceptions. If they do, the cheapest escape hatch is a per-conversation "Don't auto-save media from this chat" switch in the conversation's detail sheet — one boolean keyed by conversationId, no new settings screen.

## Test

Unit-test the decision function (`shouldAutoSave(message, prefs, alreadySaved)`) rather than the platform write: setting off → no save; non-image → no save; already-saved uniqueId → no second save; own outgoing message → no save.

## Acceptance

- [ ] Toggle exists, defaults off, persists.
- [ ] With it on, a newly received photo appears in the device gallery / Photos.
- [ ] iOS asks for add-only Photos access at the right moment, and a denial doesn't spam the user on every subsequent image.
- [ ] Same image received twice (or re-synced) doesn't produce two album entries.
- [ ] With it off, behaviour is exactly as today.

2002Bishwajeet:
Closes #1154.

## What

A new **Media** settings screen. The media-quality section moves into it out of Data and storage, and the new auto-save toggle hangs off the same screen rather than becoming a third place to look for a media setting.

**Auto-save**: incoming chat photos and videos are downloaded in the background and written to the device photo album. Default **off**, with an unmetered-only sub-toggle that defaults **on** and is hidden while auto-save is off.

Scope decisions taken with @Seifert69's thread in mind: eager download (so the album is complete without opening anything), photos and video under one toggle, chat only. No per-conversation override and no Moments — deliberately not built, per the "global and local settings are hard for users to find" concern in the issue.

## The parts that weren't a flag flip

**The receive path has two lanes and only one is obvious.** Live WebSocket pushes emit `BatchReceived`; the REST `DriveSync` lane is deliberately silent and only reports `DriveEvent.Stopped`. Hooking the WebSocket worker alone would have missed every photo that arrived while the app was closed — i.e. most of them. The service listens to both, cloning the dual-lane shape `ChatMessageStream` already uses.

**Album saving had to leave Compose.** `FileSystemHandler.saveFile` was only reachable through `@Composable getUriHandler()`, so a background coroutine couldn't call it. The Android MediaStore body and the iOS `PHPhotoLibrary` body **move** into an injectable `AlbumSaver` (−118 / −183 lines from the two `FileUtilities`), and the composable `saveFile` now delegates — no duplicated platform logic.

**There was no connectivity detection anywhere in the repo.** Zero hits for `ConnectivityManager`, `NWPathMonitor`, `isMetered`; the existing `isOnline` is WebSocket-derived and binary. `NetworkMonitor` is new across all four targets (Android `NET_CAPABILITY_NOT_METERED`, iOS `nw_path_monitor`, JVM/wasm hardcoded unmetered), plus `ACCESS_NETWORK_STATE` in the manifest.

**Dedupe is mandatory, not an optimisation.** `BatchReceived` re-fires on edits, reactions and delivery-status fan-out, so without a durable marker a reaction to an old photo re-saves it and a re-sync duplicates the album. New `AutoSavedMedia` table keyed on `(fileId, payloadKey)`. `DATABASE_VERSION` is **not** bumped — `Schema.create()` runs on every open and creates missing tables, whereas a bump would rebuild the whole local DB.

**iOS permission behaviour is fixed as a side effect.** Only `NotDetermined` now requests authorization; `Denied`/`Restricted` fail fast, so a denial no longer re-prompts on every image. This also applies to the existing manual Save menu item — same error message, so no visible UX change, but the OS no longer gets a no-op request on that path.

## Auto-save applies from when you switch it on

The sync lane rescans a window of recent messages, so on the first `Stopped` after enabling, the naive version would download and dump up to 200 historical photos into the camera roll — a surprise, a large burst, and painful to undo by hand. A timestamp is stamped when the toggle flips on and older media is skipped. Say the word if you'd rather it backfill.

## Excluded from auto-save

Own outgoing messages, soft-deleted files, stickers, and the payload keys that carry plumbing rather than user photos: `dflt_key`, `chat_web*` (long-message bodies and event cover photos), `chat_links` and `chat_loc` — the last two carry an `image/*` content type but are link and location previews.

**Segmented HLS video is skipped**, with a one-line note naming the ceiling. Saving it needs `downloadAndRemuxHlsToMp4`, a large FFmpegKit bill on every incoming video, and its concurrent `ffmpeg_execute` calls are a known iOS crash class. Enabling it wants a single-permit semaphore first.

## Test

`shouldAutoSave(...)` is a pure function — 20 tests, no Compose, DB or network: setting off, non-media content type, each excluded payload key, sticker, own outgoing, already-saved, segmented HLS, metered-with-guard vs guard-off, and the two backfill-boundary cases. Plus `UserPreferences` round-trip tests and 6 `MediaSettingsUiTest` cases.

`./gradlew homebase-chat:jvmTest homebase-common:jvmTest homebase-core:jvmTest homebase-api:jvmTest` — 3820 tests, 0 failures (Konsist `ArchitectureTest` included). `:homebase-common:compileKotlinIosSimulatorArm64` green, as are the Android/JVM/wasm compiles for all four new platform files.

**Not device-verified.** End-to-end needs two identities and a real photo send. The album write itself is the same MediaStore/PHPhotoLibrary code the existing Save menu item already uses, now reached from a coroutine.

## Android media lands in a `Homebase` subfolder

Saved media no longer drops into the top-level `Pictures/`, `Movies/` and `Download/`, the way it did when the album code moved out of `FileUtilities`. It goes into the `Homebase` subfolder of each — `Pictures/Homebase`, `Movies/Homebase`, `Download/Homebase` — following the `Pictures/WhatsApp` convention. Both Android save paths are covered: `MediaStore.RELATIVE_PATH` on API 29+, and `File(publicDir, "Homebase")` on the API 27-28 legacy path, which is live at `minSdk 28`. `onSuccess` reports the relative path, so the existing `file_saved_to` snackbar reads "Saved to Pictures/Homebase".

iOS is deliberately unchanged. A named album needs `PHAccessLevelReadWrite` in place of the current add-only access, which is not worth the extra permission — it keeps saving to Recents.

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -17,6 +17,8 @@
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
         android:maxSdkVersion="28" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <!-- Metered-vs-unmetered for the "only on Wi-Fi" auto-save guard. -->
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <!--    <uses-permission android:name="android.permission.CAMERA" />-->
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/AutoSavedMediaWrapper.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/AutoSavedMediaWrapper.kt
@@ -1,0 +1,23 @@
+package id.homebase.api.sync.database
+
+import app.cash.sqldelight.db.SqlDriver
+
+class AutoSavedMediaWrapper(
+    driver: SqlDriver,
+    private val databaseManager: DatabaseManager,
+) {
+    private val delegate = AutoSavedMediaQueries(driver)
+
+    suspend fun isSaved(fileId: String, payloadKey: String): Boolean =
+        databaseManager.readValue("autoSavedMedia.isSaved") {
+            (delegate.isSaved(fileId, payloadKey).executeAsOneOrNull() ?: 0L) > 0L
+        }
+
+    suspend fun markSaved(fileId: String, payloadKey: String) {
+        databaseManager.withWriteValue { delegate.markSaved(fileId, payloadKey) }
+    }
+
+    suspend fun deleteAll() {
+        databaseManager.withWriteValue { delegate.deleteAll() }
+    }
+}
--- a/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/DatabaseManager.kt
+++ b/homebase-api/src/commonMain/kotlin/id/homebase/api/sync/database/DatabaseManager.kt
@@ -232,6 +232,7 @@ class DatabaseManager(
         // on logout — exactly the class of bug that leaks Outbox rows across sessions.
         internal val TABLE_NAMES = listOf(
             "AppNotifications",
+            "AutoSavedMedia",
             "ChatReadCount",
             "ConnectionCache",
             "DriveLocalTagIndex",
@@ -319,6 +320,9 @@ class DatabaseManager(
         }
     }
 
+    val autoSavedMedia: AutoSavedMediaWrapper by lazy {
+        AutoSavedMediaWrapper(driver, this)
+    }
     val appNotifications: AppNotificationsWrapper by lazy {
         AppNotificationsWrapper(
             driver,
--- a/homebase-api/src/commonMain/sqldelight/id/homebase/api/sync/database/AutoSavedMedia.sq
+++ b/homebase-api/src/commonMain/sqldelight/id/homebase/api/sync/database/AutoSavedMedia.sq
@@ -1,0 +1,20 @@
+-- One row per payload the chat media auto-save has already written to the device album.
+-- Required, not an optimisation: BatchReceived re-fires for edits, reactions and delivery-status
+-- fan-out, and a re-sync replays the whole window, so without a durable marker a reaction to an
+-- old photo re-saves it and a re-sync duplicates the album.
+CREATE TABLE IF NOT EXISTS AutoSavedMedia(
+  fileId TEXT NOT NULL,
+  payloadKey TEXT NOT NULL,
+  PRIMARY KEY(fileId, payloadKey)
+);
+
+markSaved:
+INSERT OR IGNORE INTO AutoSavedMedia(fileId, payloadKey)
+VALUES (?, ?);
+
+isSaved:
+SELECT COUNT(*) FROM AutoSavedMedia
+WHERE fileId = ? AND payloadKey = ?;
+
+deleteAll:
+DELETE FROM AutoSavedMedia;
--- a/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMediaAutoSaveRules.kt
+++ b/homebase-chat/src/commonMain/kotlin/id/homebase/chat/services/ChatMediaAutoSaveRules.kt
@@ -1,0 +1,58 @@
+package id.homebase.chat.services
+
+import id.homebase.api.client.drives.files.DescriptorContent
+import id.homebase.api.client.drives.files.PayloadDescriptor
+
+const val HLS_PLAYLIST_CONTENT_TYPE = "application/vnd.apple.mpegurl"
+
+/**
+ * Whether one payload of one chat message should be downloaded and written to the device album.
+ * Pure so the policy can be tested without Compose, a database or a network.
+ */
+fun shouldAutoSave(
+    payload: PayloadDescriptor,
+    isIncoming: Boolean,
+    isSoftDeleted: Boolean,
+    autoSaveEnabled: Boolean,
+    unmeteredOnly: Boolean,
+    isUnmetered: Boolean,
+    alreadySaved: Boolean,
+    messageTimestampMs: Long,
+    enabledSinceMs: Long,
+): Boolean {
+    if (!autoSaveEnabled) return false
+    if (unmeteredOnly && !isUnmetered) return false
+    if (!isIncoming || isSoftDeleted || alreadySaved) return false
+    // Only what arrived after the switch was flipped: the recent-message window the sync lane
+    // rescans still holds older media the user never asked to have copied into their album.
+    if (messageTimestampMs < enabledSinceMs) return false
+    if (isNonMediaPayloadKey(payload.key)) return false
+
+    val contentType = payload.contentType ?: return false
+    return when {
+        contentType.startsWith("image/") ->
+            (payload.descriptorInfo() as? DescriptorContent.ImageFile)?.isSticker != true
+
+        // ponytail: segmented video is skipped rather than remuxed. Saving it needs
+        // MediaDownloadHandler.downloadAndRemuxHlsToMp4, whose concurrent ffmpeg_execute calls
+        // crash on iOS — put the remux behind a single-permit semaphore before enabling it here.
+        contentType == HLS_PLAYLIST_CONTENT_TYPE -> false
+
+        contentType.startsWith("video/") ->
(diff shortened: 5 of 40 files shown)


2002Bishwajeet:
Pushed `5d18930d2` — Android saved media now lands in a `Homebase` subfolder, plus a settings-text truncation fix.

### Folder change (Android only)

`AlbumSaver.android.kt` has two save paths and both needed it. Folder name is a single `private const val ALBUM_FOLDER = "Homebase"`.

**API 29+ (MediaStore).** `RELATIVE_PATH` was the bare `Environment.DIRECTORY_*` constant; it is now `"$directory/$ALBUM_FOLDER"`, so images go to `Pictures/Homebase`, video to `Movies/Homebase`, everything else to `Download/Homebase`. MediaStore creates the directory itself, and the path is built from the platform constant rather than a hardcoded `"Pictures"`.

**API 27-28 (legacy external storage).** `minSdk` is 28, so this branch is live on real devices. `Environment.getExternalStoragePublicDirectory(directory)` is now the parent of `File(publicDir, ALBUM_FOLDER)`; the existing `mkdirs()` creates the intermediate directory, and `MediaScannerConnection.scanFile` already scans `destFile.absolutePath`, which now points inside the subfolder.

**What `onSuccess` reports.** Both branches pass the relative path — `Pictures/Homebase` — rather than the bare directory. That string has two consumers:

- `ChatMediaAutoSaveService.saveOne` logs it (`"Saved … to $location"`).
- `MR.string.file_saved_to` (`"Saved to %1$s"`), the snackbar behind the manual Save menu item in `ChatMediaFullScreenHost` and `MomentDetailScreen`.

So the snackbar goes from "Saved to Pictures" to "Saved to Pictures/Homebase". Left as a path rather than a new string resource deliberately: it is a filesystem location, and the surrounding `file_saved_to` template is already the translatable half. The previous value was likewise a raw untranslated `Environment.DIRECTORY_*` constant — putting the path in `strings.xml` would let a translation drift away from where the file actually is.

**iOS untouched.** `AlbumSaver.native.kt` still saves to Recents/Library. A named album would mean escalating `PHAccessLevelAddOnly` to `PHAccessLevelReadWrite` — a materially wider permission for a folder label. JVM and wasmJs actuals are also unchanged.

### Also in this commit

`SettingsRow.kt` — dropped `maxLines = 2, overflow = TextOverflow.Ellipsis` from the `supportingContent` `Text`. Longer setting descriptions were being clipped mid-sentence on device. The title keeps its two-line cap.

### No test added

The change is inline `ContentValues` / `File` plumbing inside `androidMain`, which no JVM test source set can reach — `homebase-common:jvmTest` compiles `commonMain` + `jvmMain` only. There is no pure seam here without inventing one, and a Robolectric or instrumentation harness for a two-line path concatenation is not worth its maintenance. The existing suites that touch this feature all still pass.

### Verification

All run on this branch at `5d18930d2`.

```
./gradlew :homebase-common:compileAndroidMain :homebase-common:compileKotlinJvm \
          :homebase-common:compileKotlinWasmJs :homebase-common:compileKotlinIosSimulatorArm64
BUILD SUCCESSFUL in 1m 6s

./gradlew :androidApp:compileDebugKotlin
BUILD SUCCESSFUL in 1m 8s

./gradlew homebase-chat:jvmTest homebase-common:jvmTest homebase-core:jvmTest homebase-api:jvmTest --rerun-tasks
BUILD SUCCESSFUL in 2m 2s
```

3886 tests, 0 failures, 0 errors, 8 skipped (Konsist `ArchitectureTest` included). The suites this could have broken are green: `MediaSettingsUiTest` 6/6, `ChatMediaAutoSaveRulesTest` 20/20, `AutoSaveMediaPreferenceTest` 2/2. All four `homebase-common` targets matter here because `SettingsRow.kt` is in `commonMain`.

`./gradlew androidApp:installDebug` — installed on a Redmi Note 5 Pro (API 30, `id.homebase.feed.debug`, `versionName=1.4.1505`) for hands-on verification. Not driven from here: flipping auto-save writes to a real camera roll.

Branch note: the local worktree had been left on an older base. Its PR commit was patch-identical to the pushed one (`git patch-id` `106dc866f…`), so it was re-synced onto `origin/feat-1154-media-settings` and this commit went on top — the push was a clean fast-forward `369398177..5d18930d2`, no force.

`gh pr edit` failed silently again (Projects-classic GraphQL), so the description above was applied through `gh api -X PATCH repos/.../pulls/1482`, which worked.

2002Bishwajeet:
Device-tested on a Redmi Note 5 Pro (API 30, `id.homebase.feed.debug`).

**Verified working:**
- The Media settings screen — quality section migrated, both toggles persist, "Only on Wi-Fi" correctly appears only once auto-save is on.
- The `AlbumSaver` refactor via the manual Save menu item: snackbar read "Saved to Pictures/Homebase" and `/sdcard/Pictures/Homebase/chat_web0.gif` (2557720 bytes) landed on disk.
- The settings-row description no longer truncates mid-sentence.

**Not working — filed as #1496:** auto-save of incoming media is a **no-op**. `ChatMediaAutoSaveRules.kt:56` excludes `key.startsWith("chat_web")`, but `MessageActionsHandler.kt:777` keys every user attachment `chat_web<index>`. So `shouldAutoSave` is false for all incoming media and the download/dedupe/metered-guard/album-write path is unreachable.

Note that `MessageBubble.kt:533-538` and `MediaDownloadHandler.kt:118-123` filter the same constant with an **exact** `.contains(it.key)` match and treat `chat_web0…N` as genuine user media — only this rule uses a prefix match.

The fix is not simply prefix→exact: `EventComposerSheet.kt:299` gives event cover photos `chat_web0` too, so the two can only be separated by message dataType (Event = 210). Details and acceptance criteria in #1496.

Two caveats on this PR's own test evidence:
- `ChatMediaAutoSaveRulesTest.kt:54` asserts `chat_web0` is excluded, so the 20 `shouldAutoSave` tests and the full 3886-test suite pass while the feature does nothing.
- The backfill guard's device check was a **false pass** — no folder appeared, but nothing was eligible, so the guard never ran. It needs re-testing after #1496.

Issue #1485, opened 2026-09-07
WASM/web: app hangs permanently — a Kotlin exception at the JS-callback boundary traps (unreachable executed) and kills Compose's FlushCoroutineDispatcher


Seifert69:
## Symptom

The **WASM/web chat hangs** — the UI freezes completely and never recovers. Observed in **Firefox** on Linux against `michael.seifert.page`.

Not yet reproduced on demand; the original triggering exception scrolled out of the console before it could be captured (see *Missing evidence* below). **Priority deliberately left normal until we have a smoking gun.**

## Measured: the frame loop is dead, not spinning

At the time of the hang, all Firefox content processes measured **0% CPU** over a 2-second sample (`/proc/<pid>/stat` utime+stime delta). Tab memory was modest — **321–462 MB RSS** — so this is **not** a WASM OOM trap. Nothing is running; the app is not stuck in a loop, it has simply stopped scheduling frames.

## Failure chain (from the browser console)

```
Uncaught RuntimeError: unreachable executed
    org.w3c.dom.events.__convertKotlinClosureToJsClosure_((Js)->Unit)/<   chat-kmp-webApp.import-object.mjs:217
    a9ad1ba2dc3bad409bf0.wasm:15628841:1

kotlinx.coroutines.CoroutinesInternalError: Fatal exception in coroutines machinery for
  CancellableContinuation(DispatchedContinuation[FlushCoroutineDispatcher@1510514787,
  androidx.compose.animation.core.$callWithFrameNanosCOROUTINE$@1265811028]){Completed}
… (repeats for $callWithFrameNanos, SharedFlowImpl.$collect, BufferedChannel.$receive, …)

Uncaught RuntimeError: unreachable executed
    org.w3c.dom.__convertKotlinClosureToJsClosure_((Double)->Unit)/<     chat-kmp-webApp.import-object.mjs:259
    requestAnimationFrame_$external_fun                                  chat-kmp-webApp.import-object.mjs:426
```

Reading:

1. A Kotlin exception escaped inside a **DOM event handler** at the Kotlin/Wasm → JS boundary (`__convertKotlinClosureToJsClosure_((Js)->Unit)`). Wasm cannot propagate a Kotlin exception into a JS callback, so it **traps** — `unreachable executed`. **This is the original fault.**
2. The trap killed the in-flight continuation on Compose's **`FlushCoroutineDispatcher`** — note the **same dispatcher instance (`@1510514787`) in every subsequent error**.
[...]

Issue #1278, opened 2026-08-17
10-bit HLG High-10 video ships with allow-10-bit OFF (regression of #959): prober returns wrong-positive 8-bit → compression skipped → undecodable


Seifert69:
## Summary

A **10-bit HLG (BT.2020, H.264 High-10) video shipped even though the "allow 10-bit" developer flag is OFF** — the sender's own device then can't decode it ("cannot play 10-bit video"). This is a real-world reproduction / **regression of #959** (closed via `afba893c0` "fail closed on undeterminable bit depth"). The fail-closed pin is in the code but is being **defeated by the prober returning a confident *wrong-positive* 8-bit/SDR** for a genuinely 10-bit HLG capture.

## Log evidence (Android, 2026-08-17)

**1) Compression was skipped — no downconvert ran** (`homebase.log` ~08:52:10):
```
(VideoProcessing) compression finished in 791.788us: input=15045525B → compressed=15045525B quality=STANDARD trim=null..null skipped=true
(VideoProcessing) HLS decision: compressed=15045525B threshold=5242880B → SEGMENT (hls)
```
`skipped=true` with `input == compressed` = the `isAlreadyOptimal` short-circuit fired, so ffmpeg (and the 8-bit downconvert) never ran; the original bytes were HLS-segmented and sent as-is.

**2) The shipped stream is 10-bit High-10 HLG and won't decode** (~08:55:10, playback of the sent video):
```
player error: code=ERROR_CODE_DECODING_FAILED
  format=Format(…, video/avc, avc1.6E0028, …, ColorInfo(BT2020, Limited range, HLG, false, 10bit Luma, 10bit Chroma)), format_supported=NO_EXCEEDS_CAPABILITIES
Caused by: MediaCodecVideoDecoderException: Decoder failed: c2.exynos.h264.decoder
```
`avc1.6E0028` = H.264 **High 10** profile; `10bit Luma/Chroma`, HLG/BT2020 → the exact #959 signature.

## Why the fail-closed pin didn't catch it (root cause)

`FfmpegCompressPlanner.isAlreadyOptimal` fails closed correctly **on nulls**:
```kotlin
if (!allowTenBit) {
    if (bitDepth != 8) return false   // null or >8 → re-encode
    if (isHdr != false) return false  // null or true → re-encode
}
```
[...]

PR #1488, opened 2026-09-07
Move web byte crossings off Base64 onto a linear-memory bridge

### LINKED ISSUE #1387: Web: replace the Base64 Wasm↔JS byte bridge with a zero-copy typed array (video will hit this first)
## Summary

Every byte payload that crosses the Kotlin/Wasm ↔ JS boundary on the web target goes as a **Base64
string**. It is the established idiom here and it is fine for small payloads, but it costs roughly
**5–7× the payload size in peak memory** plus a **synchronous per-byte JS loop on the main thread**.

That is invisible for a thumbnail or a PDF. It is not invisible for video, which is where this will
bite first.

## Where it is today

| File | Crossing |
|---|---|
| `homebase-chat/.../video/HtmlVideoOverlay.web.kt:121` | whole clip → `<video>` blob |
| `homebase-api/.../video/BrowserVideoDecoder.kt:126-130` | whole clip → blob (okio fallback) |
| `homebase-api/.../video/FFmpegBridge.web.kt` | clip in **and** out of ffmpeg.wasm |
| `homebase-common/.../audio/AudioPlayer.web.kt:97` | whole audio file → blob |
| `homebase-common/.../clipboard/ClipboardPlatformFileFactory.web.kt:45` | pasted image |
| `homebase-common/.../util/FileUtilities.web.kt` | file download (added in #1386) |

**This is already a known, written-down concern** — two of these files predicted it:

> `HtmlVideoOverlay.web.kt:16-17` — *"a direct Uint8Array bridge is a possible follow-up if
> large-clip playback proves slow"*

> `FFmpegBridge.web.kt:13-14` — *"A direct Uint8Array bridge for the two largest crossings is a
> planned follow-up if large-clip compression is slow."*

This issue is that follow-up, filed before the pain rather than after.

## The cost

```
ByteArray (wasm linear memory)   n
Base64 Kotlin String             ~2.67n   (1.33n chars × UTF-16)
→ JS string conversion           ~1.33n–2.67n
atob() binary string             ~n
Uint8Array + charCodeAt loop     n
Blob copy                        n
```

Peak ≈ **5–7n**. The `charCodeAt` fill is per-byte, synchronous, on the main thread — order tens of
MB/s, so it *blocks the UI* for the duration.

A 100 MB clip therefore means ~600 MB peak and seconds of frozen UI, with a real chance of an
OOM'd tab. `FFmpegBridge` is the worst case: the clip crosses **twice** (in and out), so compression
pays it on both sides.

## What to do instead

1. **Zero-copy typed-array view over Wasm linear memory** — `new Uint8Array(memory.buffer, ptr, len)`,
   the `HEAPU8.subarray()` equivalent. No copy at all. This is the general fix and it benefits every
   row in the table above, not just one.
2. **File System Access API** for the *download* path specifically — `showSaveFilePicker()` +
   `createWritable()` streams straight to disk and never holds the file in memory. Chrome/Edge only,
   so it needs a fallback to the current anchor+objectURL route.
3. **Service-worker streaming** (StreamSaver.js pattern) for very large downloads cross-browser.
   Most machinery; only worth it if (2) proves insufficient.

(1) is the one that pays for itself across the whole web target. (2) and (3) are download-specific
and can wait.

## Acceptance criteria

- [ ] A single shared Wasm↔JS byte bridge that does not round-trip through Base64, used by at least
      the two video crossings and the download path.
- [ ] Large-clip playback and download do not block the main thread for a perceptible period.
- [ ] Peak memory for a crossing is within a small constant of the payload size, not ~6×.
- [ ] The `Uint8Array`-bridge follow-up comments in `HtmlVideoOverlay.web.kt` and
      `FFmpegBridge.web.kt` are removed once they are no longer true.

## Not a regression

Nothing here is newly broken and nothing needs reverting. #1386 deliberately reused the existing
idiom rather than inventing a private bridge inside a download bugfix — replacing the boundary is
repo-wide work and belongs in its own change.

2002Bishwajeet:
Closes #1387.

## The premise needed correcting first

The issue proposes `new Uint8Array(memory.buffer, ptr, len)` — the Emscripten `HEAPU8.subarray()` idiom. **That does not apply to a `ByteArray` on this target.** Kotlin/Wasm compiles to Wasm GC, not Emscripten linear memory: a `ByteArray` is a GC `(array i8)` living in the GC heap, with no address in `memory.buffer`. JS cannot form a view over one, so true zero-copy is not reachable.

What *is* reachable, verified against this build rather than from memory:

1. **`kotlin.wasm.unsafe` linear memory is real and addressable.** `withScopedMemoryAllocator` hands out a `Pointer` with an `address`, and `storeByte`/`loadByte` are plain wasm `i32.store8`/`i32.load8_s` instructions — no boundary crossing.
2. **The module exports `memory`.** The generated glue ends with:
   ```js
   const exports = wasmInstance.exports
   export const { memory, _start } = exports
   ```
3. **`js()` bodies can reach it.** They are compiled into `chat-kmp-webApp.import-object.mjs`, the same module that declares — with this comment, emitted by the Kotlin compiler itself:
   ```js
   // Placed here to give access to it from externals (js_code)
   let wasmExports;
   ```
4. **Compose Multiplatform already does exactly this.** In the same generated glue:
   ```js
   'org.jetbrains.compose.resources.jsExportInt8ArrayToWasm' : (src, size, dstAddr) => {
       const mem8 = new Int8Array(wasmExports.memory.buffer, dstAddr, size);
       mem8.set(src);
   }
   ```
   So this is the sanctioned first-party mechanism, not a private hack.

## Mechanism chosen: one bulk copy, not zero-copy

Bytes stream through a **fixed 1 MiB staging window** in linear memory. An in-wasm store loop fills the window; one typed-array view hands the whole window to JS. Per direction that is one bulk copy and **one JS call per megabyte instead of one per byte**, with no Base64 string at any point.

Being explicit, since the acceptance criteria said "zero-copy": **this is one bulk copy.** It cannot be zero-copy for the reason above. It still removes the two things that actually hurt — the ~6x Base64 expansion and the synchronous per-byte `charCodeAt` loop on the main thread. For a 100 MB clip: ~100 JS calls instead of ~100 million, and peak memory a small constant over the payload instead of ~600 MB.

The window is **fixed rather than payload-sized on purpose**: wasm linear memory never shrinks, so staging a whole clip would permanently grow the tab's heap by the clip size for the rest of the session.

`kotlinx-browser`'s `ByteArray.toInt8Array()` was evaluated and **rejected** — its source is an element-wise loop (`for (index in indices) result[index] = this[index]`), which costs one wasm→JS interop call per byte and would be *slower* than the Base64 it replaces. No new dependency was added; `kotlinx-browser` is already present transitively via Compose, and the bridge follows house style (file-private `js("…")` functions, no `external`/`@JsFun`).

## Converted vs deliberately left

New shared bridge: `homebase-api/src/wasmJsMain/kotlin/id/homebase/api/util/WasmByteBridge.kt` — `toJsBlob`, `toBlobObjectUrl`, `toJsUint8Array`, `jsUint8ArrayToByteArray`. It lives in `homebase-api` because both `homebase-common` and `homebase-chat` already `api(project(":homebase-api"))`.

**Converted** (acceptance minimum in bold):
- **`HtmlVideoOverlay.web.kt`** — `bytesToObjectUrl` deleted; its three callers (`VideoPlayerSurface`, `LocalVideoPlayerSurface`, and the overlay itself) now use `toBlobObjectUrl`
- **`FFmpegBridge.web.kt`** — the worst case, clip in *and* out: `probe`/`writeFile` now pass a `Uint8Array`, `readFile` returns one. `webApp/.../odin-ffmpeg.js` updated to match; both of its Base64 helpers are gone
- **`FileUtilities.web.kt`** — the download path
- `BrowserVideoDecoder.kt` — the okio fallback object URL
- `AudioPlayer.web.kt`
- `ClipboardPlatformFileFactory.web.kt` — builds the `File` from a Blob

**Deliberately left:**
- `WebSqlDriver` — sql.js has its own contract, and the issue scoped it out
- `ClipboardImageReader.web.kt` / `ClipboardImagePasteEffect.web.kt` and `BrowserVideoDecoder`'s poster/filmstrip JS — these are *inbound* small-JPEG paths where JS does `btoa` on canvas output. Converting them means reshaping `Promise<JsString>` into `Promise<JsAny>`, which is not the mechanical swap the rest of this change is
- Issue items (2) File System Access API and (3) service-worker streaming — deferred by the issue itself

The two now-untrue follow-up comments in `HtmlVideoOverlay.web.kt` and `FFmpegBridge.web.kt` are removed, along with several other comments that asserted the Base64 idiom.

## Verification

| Command | Result |
|---|---|
| `./gradlew :homebase-api:compileKotlinWasmJs :homebase-common:compileKotlinWasmJs :homebase-chat:compileKotlinWasmJs` | BUILD SUCCESSFUL |
| `./gradlew :webApp:compileDevelopmentExecutableKotlinWasmJs` | BUILD SUCCESSFUL (full wasm link) |
| `./gradlew homebase-api:jvmTest homebase-common:jvmTest homebase-chat:jvmTest --rerun-tasks` | BUILD SUCCESSFUL |
| `./gradlew :homebase-api:wasmJsBrowserTest` (full suite, headless Chromium 152) | all suites green |

**Runtime evidence.** `homebase-api/src/wasmJsTest/.../WasmByteBridgeTest.kt` runs in the repo's existing Karma/ChromeHeadless harness and exercises the real linear-memory path: a 2.6 MB round trip spanning several staging windows and ending mid-window, all 256 byte values (the signed-`Byte`/unsigned-`Uint8Array` boundary), the empty array, and Blob assembly across windows. 4 tests, 0 failures.

The test was **mutation-checked** rather than assumed useful: changing `emit(address, offset, length)` to `emit(address, 0, length)` makes `roundTripsAcrossStagingWindows` fail with a real element mismatch, and reverting restores green.

**Honest gap:** the ffmpeg browser tests pass but do *not* validate the ffmpeg conversion at runtime — they early-return green on web because no fixture is wired there (`FfmpegDecoderCommonTest`'s own doc: "iOS / Web (no fixture wired up yet)"), and every case reports `time="0.0"`. That is pre-existing and untouched by this change. The `FFmpegBridge`/`odin-ffmpeg.js` conversion is therefore verified by compile, full wasm link, and inspection of the generated glue, not by a live ffmpeg run. I did not run the dev server to play a clip end-to-end.

### DIFF
--- a/homebase-api/src/wasmJsMain/kotlin/id/homebase/api/util/WasmByteBridge.kt
+++ b/homebase-api/src/wasmJsMain/kotlin/id/homebase/api/util/WasmByteBridge.kt
@@ -1,0 +1,108 @@
+@file:OptIn(kotlin.js.ExperimentalWasmJsInterop::class, kotlin.wasm.unsafe.UnsafeWasmMemoryApi::class)
+
+package id.homebase.api.util
+
+import kotlin.wasm.unsafe.withScopedMemoryAllocator
+
+/*
+ * Bulk byte bridge to JS, replacing the Base64 idiom (~6x the payload in peak memory plus a
+ * synchronous per-byte `charCodeAt` loop on the main thread).
+ *
+ * Kotlin/Wasm is Wasm GC, so a ByteArray is a GC `(array i8)` with no address in `memory.buffer`
+ * and JS cannot view one directly — the Emscripten `HEAPU8.subarray()` trick does not apply here.
+ * `kotlin.wasm.unsafe` linear memory does have addresses, the module exports it as `memory`, and
+ * the compiler puts `wasmExports` in scope for `js()` bodies precisely so they can reach it
+ * (`compose.resources` moves its own bytes the same way). So bytes go through a staging window:
+ * an in-wasm store loop fills it, one typed-array view hands the window to JS.
+ *
+ * The window is fixed rather than payload-sized because linear memory never shrinks — staging a
+ * whole clip would grow the tab's heap by the clip size for the rest of the session.
+ *
+ * Not kotlinx-browser's `ByteArray.toInt8Array()`: that is an element-wise loop, so it costs one
+ * wasm->JS interop call per byte.
+ */
+private const val STAGING_WINDOW_BYTES = 1 shl 20
+
+/** Blob rather than a Uint8Array: the bytes land off the JS heap and may spill to disk. */
+fun ByteArray.toJsBlob(mimeType: String): JsAny {
+    val parts = newJsArray()
+    stageThroughLinearMemory { address, _, length -> appendBlobPart(parts, address, length) }
+    return blobFromParts(parts, mimeType)
+}
+
+/** The caller owns the revoke. */
+fun ByteArray.toBlobObjectUrl(mimeType: String): String = objectUrlFromBlob(toJsBlob(mimeType))
+
+fun ByteArray.toJsUint8Array(): JsAny {
+    val sink = newUint8Array(size)
+    stageThroughLinearMemory { address, offset, length ->
+        copyIntoUint8Array(sink, offset, address, length)
+    }
+    return sink
+}
+
+fun jsUint8ArrayToByteArray(source: JsAny): ByteArray {
+    val total = uint8ArrayLength(source)
+    val bytes = ByteArray(total)
+    if (total == 0) return bytes
+    withScopedMemoryAllocator { allocator ->
+        val window = minOf(STAGING_WINDOW_BYTES, total)
+        val base = allocator.allocate(window)
+        val address = base.address.toInt()
+        var offset = 0
+        while (offset < total) {
+            val length = minOf(window, total - offset)
+            copyFromUint8Array(source, offset, address, length)
+            for (i in 0 until length) bytes[offset + i] = (base + i).loadByte()
+            offset += length
+        }
+    }
+    return bytes
+}
+
+private inline fun ByteArray.stageThroughLinearMemory(
+    crossinline emit: (address: Int, offset: Int, length: Int) -> Unit,
+) {
+    if (isEmpty()) return
+    val source = this
+    withScopedMemoryAllocator { allocator ->
+        val window = minOf(STAGING_WINDOW_BYTES, source.size)
+        val base = allocator.allocate(window)
+        val address = base.address.toInt()
+        var offset = 0
+        while (offset < source.size) {
+            val length = minOf(window, source.size - offset)
+            for (i in 0 until length) (base + i).storeByte(source[offset + i])
+            emit(address, offset, length)
+            offset += length
+        }
+    }
+}
+
+// Every body re-reads `wasmExports.memory.buffer`; growing linear memory detaches the previous
+// ArrayBuffer, so a cached view would silently become unusable.
+
+private fun newJsArray(): JsAny = js("[]")
+
+private fun appendBlobPart(parts: JsAny, address: Int, length: Int): Unit = js(
+    "{ parts.push(new Blob([new Uint8Array(wasmExports.memory.buffer, address, length)])); }"
+)
+
+private fun blobFromParts(parts: JsAny, mimeType: String): JsAny =
+    js("new Blob(parts, { type: mimeType })")
+
+private fun objectUrlFromBlob(blob: JsAny): String = js("URL.createObjectURL(blob)")
+
+private fun newUint8Array(size: Int): JsAny = js("new Uint8Array(size)")
+
+private fun copyIntoUint8Array(sink: JsAny, offset: Int, address: Int, length: Int): Unit =
+    js("{ sink.set(new Uint8Array(wasmExports.memory.buffer, address, length), offset); }")
+
+private fun uint8ArrayLength(source: JsAny): Int = js("source.length")
+
+private fun copyFromUint8Array(source: JsAny, offset: Int, address: Int, length: Int): Unit = js(
+    """{
+        new Uint8Array(wasmExports.memory.buffer, address, length)
+            .set(source.subarray(offset, offset + length));
+    }"""
+)
--- a/homebase-api/src/wasmJsMain/kotlin/id/homebase/api/video/BrowserVideoDecoder.kt
+++ b/homebase-api/src/wasmJsMain/kotlin/id/homebase/api/video/BrowserVideoDecoder.kt
@@ -4,6 +4,7 @@ package id.homebase.api.video
 
 import id.homebase.api.file.systemFileSystem
 import id.homebase.api.util.isBlobUrl
+import id.homebase.api.util.toBlobObjectUrl
 import kotlin.io.encoding.Base64
 import kotlin.js.Promise
 import kotlinx.coroutines.await
@@ -24,8 +25,7 @@ import okio.Path.Companion.toPath
  *    read into wasm and nothing is base64'd. We do NOT revoke it; its owner (the attachment editor)
  *    does.
  *  - an okio path (e.g. an ffmpeg-produced file, or `FFmpegUtils.grabThumbnail`). We read the bytes
- *    and mint a short-lived blob URL for the `<video>`, then revoke it. This is the only path that
- *    still base64s, and it's off the interactive hot path.
+ *    and mint a short-lived blob URL for the `<video>`, then revoke it.
  */
 internal class BrowserVideoDecoder : VideoDecoder {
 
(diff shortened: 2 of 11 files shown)


2002Bishwajeet:
Merged `main` into this branch (merge commit, not a rebase — the branch is signed and will be squash-merged anyway). `main` had moved 2 commits ahead: `6ac00e8fd` (#1485) and `cd95e0bb2` (#1487).

## The conflict

One file: `homebase-api/.../video/FFmpegBridge.web.kt`. #1485 wrapped this file's ffmpeg progress callback in `guardJsCallback`; this branch had rewritten the same file's byte crossings from Base64 to `Uint8Array`. The textual conflict was confined to the **import block** — git auto-merged both sides' bodies, since the guard touches `exec()` and the bridge touches the `ff*` signatures.

Resolved by keeping **both** sides' imports and dropping only what the code no longer references:

| Line | From | Kept/dropped |
|---|---|---|
| `import id.homebase.api.browser.guardJsCallback` | main (#1485) | **kept** |
| `import id.homebase.api.util.jsUint8ArrayToByteArray` | this PR | **kept** |
| `import id.homebase.api.util.toJsUint8Array` | this PR | **kept** |
| `import kotlin.io.encoding.Base64` | main | **dropped** — no Base64 call sites remain in the file |
| `@file:OptIn(..., ExperimentalEncodingApi::class)` | main | **dropped** — opt-in existed only for the removed Base64 |

## #1485's guard survived — verified, not assumed

Diffing the resolved file against **both** parents is the proof:

- vs `de555f2cd` (this branch's tip): the *only* delta is #1485's `guardJsCallback` import plus its `exec()` wrap. Nothing else changed.
- vs `origin/main`: the *only* delta is this PR's `Uint8Array` bridge. The guard does not appear as a removal.

Repo-wide, all **12** `guardJsCallback("…")` tags from #1485 are present post-merge, an identical set to `main`'s: `audio.ended, audio.error, audio.play, audio.progress, audio.resume, clipboard.paste, ffmpeg.progress, trimVideo.loaded, trimVideo.progress, video.ended, video.progress, youauth.message`. #1485's own `RepeatedErrorFilterTest` also runs green post-merge (2/2).

Three files were touched by both sides but auto-merged; each was checked against both parents rather than trusted for being textually clean, and all three are semantically correct — the two changes occupy disjoint regions (the guard wraps *callback bodies*, the bridge changes *byte/URL construction*):

- `LocalVideoPlayerSurface.web.kt` — guard in the trim progress/loaded listeners; bridge in `resolvePlayerSrc`
- `VideoPlayerSurface.web.kt` — guard in the progress/ended listeners; bridge at `toBlobObjectUrl`
- `AudioPlayer.web.kt` — guard in the four audio callbacks; bridge replacing `audioBytesToObjectUrl`

## Drive-by doc fix

`FFmpegUtils.web.kt`'s `compressVideo` KDoc claimed it "short-circuits (already-optimal / small) by returning null — caller falls back to the original file", and that "an in-budget H.264 clip skips ffmpeg here exactly as it does on Android/iOS/Desktop". Both are false: the function returns `String` (never null) and its body has a single `return` at the end with no skip branch. `FfmpegCompressPlan`'s own KDoc records the truth — *"There is no 'skip ffmpeg' outcome: every video is re-encoded so the output is unconditionally 8-bit"* (#1278). The wrong sentences are deleted rather than reworded. No behaviour change.

## Verification (post-merge, all re-run)

| Command | Result |
|---|---|
| `:homebase-api:compileKotlinWasmJs :homebase-common:… :homebase-chat:…` | BUILD SUCCESSFUL |
| `:webApp:compileDevelopmentExecutableKotlinWasmJs` (full wasm link) | BUILD SUCCESSFUL |
| `homebase-api:jvmTest homebase-common:jvmTest homebase-chat:jvmTest --rerun-tasks` | **3191 tests, 0 failures** (6 skipped) |
| `:homebase-api:wasmJsBrowserTest` (headless Chromium 152) | **536 tests, 0 failures**; `WasmByteBridgeTest` 4/4 |

The bridge test was **re-mutation-checked after the merge**, not assumed still meaningful: changing `emit(address, offset, length)` to `emit(address, 0, length)` makes `roundTripsAcrossStagingWindows` fail with a real element mismatch; restoring it returns 536/0. So the linear-memory path is genuinely exercised on the merged tree.

<details><summary>Harness note: <code>wasmJsBrowserTest</code> can exit 0 on a genuine failure</summary>

Worth knowing for CI, and pre-existing — not introduced here. Under the mutation, the failing assertion's message is enormous (it dumps a 2.6 MB array), which overflows the Kotlin Gradle plugin's TeamCity stream handler:

```
org.jetbrains.kotlin.gradle.internal.testing.TCServiceMessageOutputStreamHandler.overflow(TCServiceMessageOutputStreamHandler.kt:95)
  at org.gradle.process.internal.streams.ExecOutputHandleRunner.forwardContent(...)
```

That kills output forwarding, so no test-results XML is written and the task reports `BUILD SUCCESSFUL` (exit 0) despite `testFailed name='roundTripsAcrossStagingWindows'` in the stream. The failure is only visible in the raw log. It does not affect the green runs above (they write all 62 XML files and overflow nothing), but a huge assertion message could mask a real failure in CI.
</details>

PR #1489, opened 2026-09-07
Web push notifications on the wasmJs target (#1359)

### LINKED ISSUE #1359: Wasm/web: check for, offer, and register web push notifications
## Goal

The **Wasm/web app should check for, offer, and register web push notifications** so a logged-in web user receives message pushes like the native apps.

**Web push is already fully supported and working in odin-core and the JS/React web client** — so this is a **client-side port** of an existing, proven flow into the wasm app, not new infra. Mirror what the JS client already does (same server endpoints, same token/subscription shape, same service worker); no server work is expected.

## Current state (grounded)

Web has **no notifications at all** in the wasm app:

- `homebase-common/.../wasmJsMain/.../notifications/NoopNotificationBackend.web.kt` — "No push / local notifications on wasmJs yet".
- `RichNotificationDisplayer.web.kt` — no-op.
- No service worker registered in `webApp/src/wasmJsMain/resources` (only `webpack.config.d/html-plugin.js`).

The plumbing to reuse is already present:

- `PushNotificationApi.subscribe(deviceToken, devicePlatform, friendlyName)` → `POST /notify/push/subscribe-firebase`, **`getSubscription()`** (`/notify/push/subscription`), `unsubscribe()` — the same endpoints the JS client hits.
- `devicePlatform()` on web **already returns `"web"`** (`DeviceDisplayName.web.kt`).
- Token/subscription acquisition needs a wasm JS-interop path (KMPNotifier has no wasmJs backend) — port the JS client's implementation.

## What the wasm client needs to do (mirror the working JS client)

1. **Capability check** — feature-detect `Notification`, `navigator.serviceWorker`, `PushManager`; degrade gracefully where absent.
2. **Check existing registration first** — before offering, determine whether this device is already registered and working: call **`getSubscription()`** (server side) and cross-check the browser side (`Notification.permission` + `serviceWorker.getRegistration()` / `pushManager.getSubscription()`). If it's already registered and healthy, **don't offer**; if the server has no/stale registration but the browser is permitted, re-register silently.
3. **Actively offer it** — when not registered, **proactively surface an in-app offer** (a prompt/banner/dialog), don't wait for the user to dig into settings. The offer includes a **"don't ask again" checkbox** so a user who declines isn't re-nagged (persist that choice). The offer's **"Enable" button is the user gesture** that then invokes `Notification.requestPermission()` — that's the only reason a gesture is involved; it does **not** mean the offer must be passive. Also expose the toggle in notification settings so a user who dismissed/opted-out can turn it on later.
4. **Register** — on Enable, register the service worker, obtain the token/subscription exactly as the JS client does, and call `subscribe(...)` with `devicePlatform = "web"`. Re-subscribe on token rotation; `unsubscribe()` on logout/opt-out.
5. **Display** — implement `RichNotificationDisplayer.web.kt` for foreground; background notifications from the service worker (as in the JS client).

## Reference & caveats

- **Source of truth: the existing JS/React web client + odin-core.** Match its VAPID/FCM choice, service-worker file, and payload handling rather than re-deriving — the server already speaks this.
- **Service-worker scope:** the wasm app loads sql.js / ffmpeg / skiko via the page — scope the SW so it doesn't interfere with those fetches or the SPA bootstrap (watch the historical `.cvr` SPA-fallback MIME gotcha on this server).
- **iOS Safari:** web push only fires when the site is **installed to the Home Screen as a PWA** (iOS 16.4+) — relevant since the web app also runs on phones. Desktop browsers work in-tab.
- Web target is currently **partial/disabled** in `settings.gradle.kts` — this rides on web being buildable/enabled.

## Acceptance criteria

- [ ] When a logged-in web user is **not** registered (and hasn't chosen "don't ask again"), the app **proactively shows an offer** to enable notifications — not buried in settings.
- [ ] The offer has a **"don't ask again"** checkbox that is persisted and honored (no re-nagging); the toggle is still reachable in notification settings afterward.
- [ ] Registration status is determined via **`getSubscription()`** + browser state, and an already-registered/healthy device is **not** offered; a stale/missing server registration with browser permission re-registers.
- [ ] Enabling registers the service worker + token and calls `subscribe(..., devicePlatform = "web", ...)`; `getSubscription()` reflects it.
- [ ] A real message push is **received and displayed** on web — foreground (page) and background (service worker).
- [ ] Parity with the JS client (same server endpoints/shape; no server change needed).
- [ ] Unsupported browsers degrade gracefully; permission denial handled.
- [ ] Logout / opt-out calls `unsubscribe()` and tears down the subscription.

2002Bishwajeet:
Closes #1359.

## The issue's premise was wrong — this is raw VAPID, not `subscribe-firebase`

The issue asks for "a client-side port … same `subscribe-firebase` endpoint, same shape". Building that would have produced a subscription the server accepts and then **silently never delivers to**. There is no Firebase in the JS web client — `grep -rln -i firebase packages/` over `dotyoucore-js` returns nothing, and `subscribe-firebase` appears nowhere in it.

`PushNotificationService.cs` routes on one line:

```csharp
if (string.IsNullOrEmpty(subscription.FirebaseDeviceToken))
    tasks.Add(WebPushAsync(...));       // raw VAPID
else
    ... DevicePushAsync(...);           // the FCM relay
```

Empty `FirebaseDeviceToken` ⇒ raw VAPID via the `WebPush` NuGet package. `WebpushConfig` appears nowhere in `DotYouCore`, so an FCM-routed web subscription is a dead end.

|  | KMP before | Web needs |
|---|---|---|
| Endpoint | `POST /api/v2/notify/push/subscribe-firebase` | `POST /api/v2/notify/push/subscribe` |
| Body | `{DeviceToken, DevicePlatform, FriendlyName}` | `{friendlyName, endpoint, expirationTime, auth, p256DH}` |
| Key | FCM token via KMPNotifier | VAPID key from `GET /api/guest/v1/public/keys/notifications_pk` |

Three more premises corrected along the way:

- **`devicePlatform() == "web"` is irrelevant** — `/notify/push/subscribe` has no platform field at all.
- **The web target is not disabled.** `include(":webApp")` is present and it builds. What is memory-only is the *SQLite database* (sql.js); `createSettings()` on wasmJs is `StorageSettings()` over `localStorage`, so preferences are durable — which is where the "don't ask again" flag lives.
- **`getSubscription()` alone cannot decide health.** `PushNotificationSubscription.Redacted()` drops `Endpoint`/`Auth`/`P256DH`, so a *healthy* browser subscription reads back with a null `firebaseDeviceToken`, which `checkSubscription` maps to `TOKEN_MISMATCH` — and the self-heal is gated on `subscription?.firebaseDeviceToken != null`, so it never fires. Every healthy web subscription would report a permanent, unrecoverable mismatch. Web health is a separate decision (below).

## What landed

All six phases of the implementation plan.

**1 — honest permissions.** `PermissionsManager.web.kt` returned `isPermissionGranted = true` unconditionally, which is exactly why the permission card in `NotificationSettingsScreen` never rendered on web. `NOTIFICATION` now reports the real `Notification.permission` and `askPermission` drives `Notification.requestPermission()`. The blind `askPermission(NOTIFICATION)` in `AppNavHost` is skipped on web — a browser only prompts from a user gesture — and the permanently-denied "Open Settings" button is hidden there, since `launchSettings()` is a no-op a browser denial can't be undone from.

*Deviation from the plan:* the plan also proposed returning `false` for every non-`NOTIFICATION` type instead of `true`. I kept `true`. Flipping `GALLERY`/`RECORD_AUDIO` would change the visible gates in `GalleryPermissionState` and `RecordAudioPermissionState` — camera/mic/file access on web reaches the browser through pickers and `getUserMedia`, which prompt on their own — and that is an unrelated composer-behaviour change I can't verify without a live tenant. Worth a follow-up, not this PR.

**2 — manifest + service worker.** `manifest.json` did not exist even though `index.html:9` has linked it all along (a live 404 on every page load). It needs `display: "standalone"`: iOS Safari 16.4+ fires web push *only* for a Home-Screen-installed PWA. `sw.js` has **deliberately no `fetch` handler** — push/`showNotification`/`notificationclick` all work without one, and a pass-through handler forces a worker start on every request and breaks the HTTP `Range` requests this app streams media over. The payload it decodes is a *batch*, not the FCM single-payload shape; notifications coalesce on `options.typeId` for chat and `options.tagId` elsewhere. Registration is lazy and scoped from `document.baseURI`, so the `-PpublicPath=/apps/chat-wasm/` mount keeps its scope, and a dev-server MIME failure (`historyApiFallback` + `disableDotRule` answers a missed `/sw.js` with `index.html`) fails quietly instead of throwing.

**3 — the JS bridge.** `odin-push.js` exposing `globalThis.__odinPush`, matching the house style (`odin-ffmpeg.js` / `odin-sqljs.js`); no new Gradle dependency. `kotlinx-browser` has no Push API at any published version and `kotlin-wrappers` would pin us to an old Kotlin, so the Kotlin side is file-private `fun x(): T = js("…")` — the repo has zero `external`/`@JsFun` declarations. Two details that are easy to get wrong: `applicationServerKey` takes the server's base64url string **directly** (no `urlBase64ToUint8Array`), and `p256dh`/`auth` are base64**url**-normalised on the way out — one of the public prior-art repos emits plain base64 here, which is a latent server interop bug.

Every op **resolves** rather than rejects. On kotlinx-coroutines 1.10.2 the wasmJs `await()` collapses a JS rejection into `IllegalStateException("Unexpected non-Kotlin exception…")` and **the `DOMException.name` is lost** — and `NotAllowedError` (denied) vs `AbortError` (push service unreachable) is exactly what has to be told apart, so the name is extracted JS-side. Capability detection distinguishes *"iOS, needs Home Screen install"* from *"unsupported"*: on a non-installed iOS Safari tab the APIs are absent, not denied, so the UI says "Add to Home Screen" instead of showing a dead button.

**4 — API client.** `subscribeWebPush()` and `getVapidPublicKey()` on `PushNotificationApi`. The VAPID fetch is a *guest* endpoint outside `/api/v2` — unauthenticated, unencrypted, and it answers with a bare string, not JSON — so `encryptedGet`/`deserialize` don't apply and it uses a plain `httpClient.get`. `expirationTime` is sent for JS parity but never read back: `PushNotificationSubscribeDeviceRequest.ExpirationTime` is accepted server-side and then never copied onto the subscription.

**5 — the user-facing surface is Notification Settings, not a banner.** This started as a top-of-screen offer banner; it was replaced at the repo owner's request after he ran it locally, because settings is where enabling, inspecting and repairing all belong together. The screen is now the whole web-push surface and every web-push affordance on it is behind `isWeb()` / `WebPushService.isSupported`, so nothing changes on Android, iOS or Desktop:

- **Enable** — the existing permission card's *Enable Notifications* button. That click is the user gesture the browser gates `Notification.requestPermission()` on; the grant callback then runs `evaluate()`, which sees granted-but-unsubscribed as `NEEDS_REPAIR` and subscribes.
- **Repair** — the existing *Re-register Push Notifications* card, routed to `WebPushService.enable()` on web. It had to be: the FCM path it used to run there calls `api.unsubscribe()` first and then finds no device token, so it destroyed a working web subscription. Same card, same `isReRegistering` / `reRegisterResult` feedback.
- **See** — the *Push notification status* row is no longer behind the five-tap debug gesture on web, where it is a user's only view of whether push works. The device-token and FCM server-verification rows are hidden there instead, which retires the `TOKEN_MISMATCH` rough edge this PR previously shipped with; `verifyServerSubscription()` no longer runs on web at all.
- **Nothing to do** — `WebPushHealth`, not the permission callback, decides which dead end the card explains, so a denial that predates the visit is recognised as one rather than offering a button the browser silently ignores. `NEEDS_INSTALL` says add to Home Screen; a denial gets its own copy (`settings_notifications_denied_browser_body`) because `launchSettings()` is a no-op and the phone wording pointed at device settings.

Gone with the banner: `WebPushOfferBanner`, its `AppNavHost` overlay, `AppViewModel`'s `showWebPushOffer` + `refreshWebPushOffer`/`enableWebPush`/`dismissWebPushOffer` and its `WebPushService` dependency, the `webPushOfferSuppressed` localStorage flag, `shouldOfferWebPush`, and the four `web_push_offer_*` strings.

**6 — health, teardown.** A web-specific health check replacing `verifySubscription()` (see the Redacted point above), re-run on every launch rather than once: Chrome began auto-revoking granted notification permissions for low-engagement sites in Oct 2025, so `granted` can silently revert to `default` months later. A granted-but-unsubscribed browser is repaired silently. Logout adds the browser-side `pushManager.getSubscription()?.unsubscribe()` next to the already-wired server unsubscribe.

`RichNotificationDisplayer.web.kt` stays a no-op, but its KDoc claim ("handled by the service worker / KMPNotifier") is now accurate: `userVisibleOnly: true` obliges `sw.js` to display a notification for *every* push, foreground included, and no push path reaches Kotlin on web — so it is never on a live path.

## Verification

Actually run, with outcomes:

- `./gradlew :homebase-api:compileKotlinWasmJs :homebase-common:compileKotlinWasmJs :homebase-core:compileKotlinWasmJs :homebase-core:compileKotlinJvm` — **BUILD SUCCESSFUL**
- `./gradlew :homebase-common:compileAndroidMain :homebase-common:compileKotlinIosSimulatorArm64` — **BUILD SUCCESSFUL** (the three no-op `webPushBridge()` actuals)
- `./gradlew homebase-api:jvmTest homebase-common:jvmTest homebase-chat:jvmTest homebase-core:jvmTest homebase-auth:jvmTest --rerun-tasks` — **BUILD SUCCESSFUL, 3892 tests, 0 failures / 0 errors.** Includes `homebase-common`'s Konsist `ArchitectureTest` (no hardcoded `Text("…")`).
- `./gradlew webApp:wasmJsBrowserDevelopmentExecutableDistribution --no-configuration-cache` — **BUILD SUCCESSFUL**; `sw.js`, `odin-push.js`, `manifest.json`, `icon-192.png`, `icon-512.png` all land in `dist/`, and the emitted `index.html` carries `<base href="/">`, the manifest link and `odin-push.js` in the bootstrap script list.
- `./gradlew webApp:wasmJsBrowserDevelopmentRun --no-configuration-cache` plus a real Chrome session against `localhost:8080`:
  - `/sw.js` → 200 `application/javascript`, `/manifest.json` → 200 `application/json`, `/odin-push.js` → 200 `application/javascript` (so the dev-server MIME trap does not bite here)
  - `typeof globalThis.__odinPush === "object"`, `capability() === "default"`, `PushManager in self`, `isSecureContext`
  - service worker **registered and `activated`** at scope `http://localhost:8080/`, `scriptURL` `.../sw.js`
  - `currentSubscription()` → `""` (none)
  - `subscribe('not-a-real-vapid-key')` → **`"err;InvalidAccessError"`** — a real `DOMException.name` survived the boundary as a resolved value, which is the whole point of the callback bridge
  - app boots to `#login`, no console errors
- JVM test `WebPushDecisionsTest` (4 cases) covers the health matrix and the payload shape — including that `p256DH` survives `CamelCaseNamingStrategy` and a null `expirationTime` is dropped by `explicitNulls = false`. `NotificationSettingsUiTest` adds two cases pinning that the status panel stays behind the debug gesture off web.

**UNVERIFIED** — verification was cut short at the author's request, and some of it needs a live tenant:

- **A real message push is received and displayed — NOT TESTED.** Neither foreground (page open) nor background (tab closed, service worker). This is the headline acceptance criterion and it is untested.
- The whole authenticated path is untested end to end: `getVapidPublicKey()` against a live identity, `subscribeWebPush()` being accepted, and the resulting subscription actually being delivered to.
- `notificationclick` → `clients.focus()` + `postMessage` → `NotificationClickRouter` → in-app navigation. The SW side parses and activates; the round trip is untested.
- The web-push affordances in Notification Settings have not been exercised against a live identity — enable, repair and the status row all need an authenticated session.
- The production sub-path mount (`-PpublicPath=/apps/chat-wasm/`) was not built. The scope is computed from `document.baseURI` and the substitution machinery is unchanged and pre-existing, but that specific build was not run.
- iOS Safari Home-Screen-install path: not tested on a device.
- Android/iOS/Desktop were compile-checked and unit-tested, but not run. Behaviour there should be unchanged — `webPushBridge()` returns null and every entry point short-circuits.

## Known rough edge, deliberately left

`PermissionsManager.web.kt` still answers `true` for `GALLERY` / `RECORD_AUDIO` / camera — see the deviation note under phase 1. Flipping those would change the composer's visible gates, which is unrelated to push.


2002Bishwajeet:
Banner is gone; Notification Settings is now the single place web push is enabled, inspected and repaired. `70555fc` — 12 files, +114 / −199.

## Deleted

- `WebPushOfferBanner.kt` (the whole file, 88 lines)
- Its `AppNavHost` overlay and import
- `AppViewModel`: `showWebPushOffer`, `refreshWebPushOffer()`, `enableWebPush()`, `dismissWebPushOffer()`, the `refreshWebPushOffer()` call on credentials load, and the `WebPushService` constructor parameter
- `UserPreferences.webPushOfferSuppressed` (the localStorage "don't ask again" flag)
- `WebPushService.shouldOffer()` / `suppressOffer()` and, with them, its `UserPreferences` dependency — `WebPushService(api)` now takes one argument
- `shouldOfferWebPush()` in `WebPush.kt`, and the `WebPushDecisionsTest` case that covered it
- The four `web_push_offer_*` strings

Nothing references any of it — `grep -rn "showWebPushOffer\|enableWebPush\|dismissWebPushOffer\|shouldOfferWebPush\|webPushOfferSuppressed\|WebPushOfferBanner\|web_push_offer"` over the repo returns nothing.

`WebPushService` is **still needed** in Koin — `SettingsViewModel` uses `disableLocally()` on logout and `NotificationSettingsViewModel` uses `isSupported` / `evaluate()` / `enable()`. Only `AppViewModel`'s edge went away; `viewModelOf(::AppViewModel)` is constructor-injected so the DI line was untouched.

## What a web user gets in Notification Settings

Everything below is behind `isWeb()` in the screen or `webPushService.isSupported` in the ViewModel — matching how the file already gated the "Open Settings" button — so Android / iOS / Desktop render and behave exactly as before.

**Enable.** The permission card's *Enable Notifications* button, which already existed. That click is the user gesture the browser gates `Notification.requestPermission()` on — the banner used to be that gesture, and this is now the only one, so I checked the path end to end: button → `RequestPermission` → `PermissionsManager.web.kt#askPermission` → `bridge.requestPermission()` → `updatePermissionStatus` → `loadWebPushStatus()` → `evaluate()` sees granted-but-unsubscribed as `NEEDS_REPAIR` and subscribes + registers. Intact. The `!isWeb()` guard on `AppNavHost`'s blind `askPermission` is untouched.

**Repair.** The existing *Re-register Push Notifications* card now routes to `webPushService.enable()` when `isSupported`. This one was not optional: on web the FCM path it used to run calls `api.unsubscribe()` first and then finds no device token, so tapping it **destroyed a working web subscription** and reported failure. Same card, same `isReRegistering` spinner and `reRegisterResult` banner, no parallel affordance. Only success gets the green banner — every failure is already spelled out by the permission card or the status row, so a red "Registration failed: BLOCKED" would just be noise (and would have needed six new strings).

**See the true state.** The *Push notification status* row is no longer behind the five-tap debug gesture on web, where it is a user's only view of whether push actually works. The device-token and FCM server-verification rows are hidden there instead, and `verifyServerSubscription()` returns early — which retires the `TOKEN_MISMATCH` rough edge this PR previously shipped with. `WebPushHealth` is still the source of truth; `getSubscription()`'s redacted `firebaseDeviceToken` never enters the picture.

**Told plainly when there's nothing to do.** `applyWebPushHealth()` makes health, not the permission callback, the authority for the card: `NEEDS_INSTALL` → the Home Screen message, no button; `BLOCKED` → a denial message, no button. Two corrections fell out of that:

- A denial that predated the visit used to read as a plain "not granted", so the card offered an Enable button the browser silently ignores. `loadWebPushStatus()` now also runs on a refused permission result, so `BLOCKED` is recognised on arrival.
- The denial copy said "enable them in your device settings", which is wrong in a browser and points at a `launchSettings()` that is a no-op. One new string, `settings_notifications_denied_browser_body`, says site settings + reload.

## Tests

`WebPushDecisionsTest` lost the offer-gate case (it asserted deleted code); the health matrix and payload cases stay. `NotificationSettingsUiTest` gains two cases pinning that the status panel stays behind the debug gesture off web — `isWeb()` is an `expect fun`, so a common test can only exercise the non-web side, which is exactly the "don't change what phone users see" half. Both were mutation-checked: forcing the panel visible fails `pushStatusPanelStaysBehindTheDebugGestureOffWeb` only; forcing `fcmDebugRows = false` fails `pushStatusPanelShowsOnceTheDebugGestureUnlocksIt` only. Restored and re-run green.

## Verification (all actually run)

- `./gradlew homebase-core:jvmTest homebase-common:jvmTest homebase-chat:jvmTest homebase-api:jvmTest --rerun-tasks` — **BUILD SUCCESSFUL**, **3923 tests, 0 failures / 0 errors** (core 675, common 617, chat 1381, api 1250). Includes `homebase-common`'s Konsist `ArchitectureTest`, which is what would catch a hardcoded `Text("…")`.
- `./gradlew :homebase-core:compileKotlinJvm :homebase-core:compileAndroidMain :homebase-core:compileKotlinWasmJs :homebase-core:compileKotlinIosSimulatorArm64` — **BUILD SUCCESSFUL**. One run of this failed partway; rather than accept the retry I re-ran each target with `--rerun-tasks` from cold (wasmJs alone, then iOS + Android together) and all four compiled clean, so the failure was host contention, not the change.
- `./gradlew :webApp:compileDevelopmentExecutableKotlinWasmJs` — **BUILD SUCCESSFUL**.

## Left alone deliberately

- The re-register card's label still reads "Re-register Push Notifications" on web. It is accurate for repair and re-wording it would cost a string for no clarity.
- `rememberOpenSystemNotificationSettings()` behind the *Message Sound* row is presumably inert on web. Unrelated to push, unchanged.
- `PermissionsManager.web.kt` still answers `true` for gallery/microphone/camera — the deviation already documented in the PR body.
- Everything in the PR's **UNVERIFIED** list stands: no push has been received end to end, and the new settings affordances have not been exercised against a live identity.

PR #1528, opened 2026-09-15
Consolidate repeated predicates and helper blocks

Seifert69:
## Summary

A cleanup sweep for the most-repeated compound conditions and copy-pasted blocks. Each gets one home; call sites route through it. **31 files, +183 / −292.**

### Repeated predicates → named helpers
- `CharSequence.isSurrogatePairAt / charCountAt / decodeCodePointAt` (`api/util/StringExtensions.kt`) replace 8 hand-rolled surrogate checks (api `truncateToCodePoints`/`codePointCount`, `GroodleDescriptor`'s private duplicate of `codePointCount`, `EmojiText`, `RichTextSplice`, `RichTextExtensions` ×2, homebase-common's private `codePointAt`).
- `isQuarterTurn(degrees)` (`api/util/Rotation.kt`) replaces 7 copies of `((r % 360) + 360) % 360 → 90 || 270` across the four `FFmpegUtils` actuals, `FfmpegCompressPlanner` and `ImageUtil.skia.kt`.
- `KeyEvent.isEnter` (homebase-common `core/util/KeyEventExtensions.kt`) replaces 5 `Key.Enter || Key.NumPadEnter` checks in `MessageInputBar` and `MomentDetailScreen`.
- `showHoverActions` in `MessageBubble` replaces `isDesktopOrWeb() && !message.isDeleted` at 6 sites.

### Repeated blocks → helpers
- `HoverActionIcon` composable: 6 near-identical `IconButton` blocks in `MessageBubble.kt`.
- `formatLatLon` in api `LocationPreview.kt`: 3 private copies deleted; `EventComposerSheet`'s full-precision `"$lat, $lon"` fallback uses it too.
- `PayloadDescriptor.largestThumbnail` / `displayAspectRatio` + `ThumbnailDescriptor.aspectRatio`: 5 copies of the thumbnail aspect math plus `ImageInfoOverlay`'s largest-thumb filter. `remember {}` around the trivial division dropped (`PayloadDescriptor` is `@Immutable`).
- `ImageSize.positiveOrNull`: 7 `w > 0 && h > 0` guards in `ImageHeaderParser` / `ThumbnailGenerator`.
- FFmpeg trim: the per-platform `effectiveTrimStart/End` pairs are gone; `FfmpegCompressPlan` carries `trimDurationMs`, so the planner alone decides "both bounds or nothing".

## Behaviour changes (small, all in the direction of a fix)
1. `MessageInputBar` caption composer checked `Key.Enter` only — NumPadEnter now sends there too (the macOS Return-as-NumPadEnter case that the main composer already handled).
2. `RichTextExtensions.getCodePointAt` only combines a real surrogate pair; a lone high surrogate followed by a normal char decodes as itself.
3. `ThumbnailGenerator` SVG parsing returns `null` for `width="0"` instead of `ImageSize(0, 0)`.
4. `largestThumbnail` picks by area, not `lastOrNull()` — the server doesn't sort thumbnails uploaded by other clients. `displayAspectRatio` falls back to the preview thumb when the largest thumb has no dims (previously `null`).

## Deliberately not touched
- `LocationContent.format5()` rounds rather than truncates — left alone.
- HLS copy-vs-re-encode command duplication across the four actuals: real, but native builds a quoted shell string, so it needs a quoting decision first. Follow-up.
- homebase-common's own `ImageSize` duplicating api's — pre-existing, out of scope.
- `MomentMediaGallery.aspectRatioFor` keeps its preview-first precedence (the TEMPORARY `MomentAspect` log is chasing that).

## Verification
- `compileKotlinJvm` for api/common/chat/core, `:homebase-api:compileAndroidMain`, `:homebase-api:compileKotlinWasmJs` green.
- `homebase-api:jvmTest homebase-common:jvmTest homebase-chat:jvmTest` green.
- iOS not compiled locally (Linux host); native changes are the same mechanical swaps as the other actuals.

https://claude.ai/code/session_01RW5j5PY4QwfrELkFvSGVqf


Seifert69:
Super interesting but looks a little risky so don't want to do it right now

Issue #1223, opened 2026-08-05
Windowed chat sync (Part 2): load older messages from the server on local EOS, then enable the chat DriveSyncPolicy


Seifert69:
## Summary

Complete the windowed-chat-sync story ("Part 2"). Today, first login **fully syncs every drive to 100%** — a chat drive with 50K+ messages makes the first login slow, which is especially painful in the **WASM/web client** for casual use. We already built the ability to windowed-sync only the last N days on a fresh login, but it's **disabled** because the missing half is: on a partially-synced conversation, when you scroll to the top of local history (**local EOS**), the app should **query the server for older messages** (a `queryBatch` of ~1,000 more for that conversation, spinner shown), upsert them locally, and keep paging. The cursor is straightforward to build from the oldest message available locally.

## What already exists (grounded)

- **Windowed fresh-sync engine — built + tested, currently OFF.**
  `DriveSyncPolicy(fullSyncWindow, initialQueries)` (`homebase-api/.../sync/DriveSyncPolicy.kt`) is applied in `DriveSync.kt` (seed the fresh-sync cursor to `now - window`, plus `runInitialQueries` for the full conversation-file list). Covered by `DriveSyncTest.kt` (`fullSyncWindow = 60.days` + `FileQueryParams(fileType = ConversationFileType)`), and diagnostics confirmed the window cursor is correct (zero dupes / no floor breaches at 7/30/60 days).
- **It is not enabled in production.** `DriveSyncManager(... driveSyncPolicies = emptyMap())`; the chat policy is commented out in `homebase-core/.../di/AppModule.kt:405-431` with a note that it's **Part 2**, gated on *this* feature — because enabling the window without scroll-to-load "would only see the last N days, which is confusing without that scroll-to-load path."
- **Message pagination is local-DB only.** `ChatMessageStream.loadOlderMessages` (`ChatMessageStream.kt:295`) → `fetchMessages` (`:746`) reads the **local** index via `QueryBatch.queryBatchAsync`. `hasOlderMessages = result.hasMoreRows` reflects **local** exhaustion. When the local DB runs out it simply **stops** — even if older messages exist on the server outside the sync window.
- **Proven "fetch a header from the server and map it locally" pattern** already exists: `ChatMessageStream.resolveForNotification` (`:651`) falls back to `driveFileProvider.getFileHeaderByUid(...)`, which decrypts with the shared secret and maps through the same `mapToMessageData` pipeline. The bulk version of this is what we need.

## The work

[...]

PR #1545, opened 2026-09-16
Keep add-on drives mounted when the registry file is outside the chat sync window

toddmitchell:
## Problem

Moments "Set it up" did nothing, and every add-on drive (Moments, Vault, Location, …) was unmounted right after each launch.

The log showed the sequence on every cold start:

```
DriveRegistry bootstrap local-DB returned 0 drive(s)
bootstrap() end (reconciled local=0 + server=7 -> 7 drive(s))
mountDrive(a85f8562…, Moments) registered
AuthCC: onConnected post-sync done
unmountDrive(…) × 7
```

No grant-revoked or 403 warning — the unmounts came from `DriveRegistry.reconcile`.

**Root cause:** since the windowed chat sync (#1223), a fresh sync only pulls chat files changed in the last 14 days plus the conversation list. The registry file (fileType 4242) is older than that, so it never reaches the local index. The observer's first Chat `Stopped` read the missing row as an empty list, diffed it against the 7-drive baseline from bootstrap, and unmounted all of them. An incremental sync never brings the file back, so it repeated on every launch. Reconnects also read the local row, so they subscribed the WebSocket to the mandatory drives only.

**Why the button was dead:** `MomentsViewModel` activates on a `permissionsGranted` false→true edge. The launch-time check already set it true, and a StateFlow doesn't re-emit the same value, so the recheck on tap produced nothing. Direct activation existed only for web. WebDrop had the same bug; Vault, Location and Email already handled it.

## Changes

- **`DriveRegistry`**: a missing local registry row is "not synced yet", not "no drives" — `reconcile` skips removals. When bootstrap fetches the server copy and there's no local row, it upserts it into the local index (same path as `ChatServerHistory`), which heals existing installs on their next launch.
- **`AppModule`**: the chat drive's fresh-sync `initialQueries` also pull `RegistryDriveFileType`.
- **`MomentsViewModel` / `WebDropViewModel`**: "Set it up" activates directly when permissions are already granted.

## Testing

- New `DriveRegistryTest` cases, both verified failing without the fix:
  - `observerKeepsBaselineWhenChatSyncLandsWithoutTheRegistryFile`
  - `bootstrapStoresTheServerRegistryLocallyWhenTheLocalRowIsMissing`
- `DriveRegistryTest` 44/44 pass; `homebase-core` compiles for JVM and Android.
- Not yet verified on device. Expect `registry file missing locally — stored the server copy` on the next launch and no `unmountDrive` after `onConnected post-sync done`.


Seifert69:
### Startup cost: this removes a per-launch round-trip rather than adding one

Worth stating explicitly, since `storeLocally()` looks like it might add startup work.

**No new request is made.** `runBootstrap` on main already calls `fetchServerRegistry()` unconditionally — the comment above it explains why (a drive activated on another device exists only in the server file, so serving local-when-non-empty was the original reconciliation bug). This PR reuses that same call as `fetchServerRegistryFile()` and keeps the header it already fetched. The added work is one local index upsert, only when the row is missing.

**Startup already skips the fetch in the healthy case.** `AuthConnectionCoordinator.kt:286` calls `bootstrap(deferServerReconcile = true)`, which returns the local list with no network when a local registry row exists (`DriveRegistry.kt:153-161`).

**So the per-device cost changes like this:**

| | today (main) | with this PR |
|---|---|---|
| device missing the registry row | `loadDrives()` is empty → deferred path falls through → **server fetch on every launch, forever** (nothing ever writes the row back) | fetch once on the next launch, `storeLocally()` writes the row → **zero registry requests on later launches** |
| device with the row | no registry request at startup | unchanged |

The devices hit by this bug are exactly the ones currently paying a fetch per launch, so healing the row is also what stops that.

Remaining exception: an identity whose registry file does not exist server-side yet (`serverFile == null` — nothing to store) keeps taking the deferred-path fetch each launch. That is a fresh identity with no add-on activated, and it resolves the moment one is.

Unchanged by this PR: the post-connect `scheduleRegistryReconcile` (`AuthConnectionCoordinator.kt:919-928`), which runs off the connect path and doesn't block startup.

